PR #1, opened 2017-12-05
Fix nested routes children empty problem

MengWeiChen:
Hi paulgv, 

This lib is so great, it saves me a lot of time to add i18n feature into my project.

I found a bug which causes Nested Routes children return empty, then I do some fix in this PR.


paulgv:
Thanks a lot for your help @MengWeiChen !

PR #3, opened 2017-12-06
feat: Add ignorePaths option

paulgv:
ignoreRoutes accepts an array of String representing routes that should not be localized

https://github.com/paulgv/nuxt-i18n/issues/2


Evild67:
I have an issue with this pull request. 

This won't work with my folders :
```
admin
- editor
-- index.vue
-- _slug.vue
```
I want to ignore this folder, my config :
```
ignoreRoutes: [
          'admin-editor'
]
```

But routes was not ignored. 
* /en/admin/editor
* /admin/editor
No 404 error

To fix this, I changed `route.name` check with `route.path` and also my ignoreRoutes.

```javascript
  baseRoutes.forEach(route => {

    if (ignoreRoutes.indexOf(route.path) !== -1) {
      newRoutes.push(route)
    } else {
      localizedRoutes.push(route)
    }
  })
```

```javascript
ignoreRoutes: [
          '/admin/editor'
]
```

PR #4, opened 2017-12-06
Fix nested routes children wrong path problem

MengWeiChen:
Hi paulgv,

Sorry, I haven’t done enough test for the previous PR. 

The new PR have fixed the Nested Routes problem completely.


paulgv:
Thanks @MengWeiChen !
I never used nested routes in Nuxt and don't have any test case but I'll try and write some tests specifically for this. Merging this in the meantime!

PR #6, opened 2018-01-10
change ignoreRoutes from name to path

Evild67:


paulgv:
Thanks! I’ll take a closer look ASAP, probably in ~2 weeks and will release the change along with some tests.
I wonder if it would be interesting to be able to specify available languages in the pages themselves, rather than relying a the configuration.

paulgv:
Thanks @Evild67 !

PR #13, opened 2018-02-07
Fix for URLs with hashes. 

### LINKED ISSUE #12: Currently breaks if hashes are used in the URL.
E.g.`switchLocalePath` turns `/#intro` into `/fr#intro/`, which results in the following error: `DOMException: Failed to execute 'querySelector' on 'Document': '#intro/' is not a valid selector.`.

jwahdatehagh:
Fixes #12


paulgv:
Awesome! Thank you @jwahdatehagh :)

PR #25, opened 2018-02-20
Add localePath and getRouteBaseName to Nuxt's app

### LINKED ISSUE #24: Cannot access localePath in middleware
I'm making a middleware which redirects users to some pages depending on the store informations. However, I'm lacking `localePath` to localise the path for the redirect. I expected it to be available as `context.app.localePath`, but it's not.

More interestingly, it's available at `context.app.i18n.vm.localePath` but the `this` context is not correct. `localePath` is somehow bound to a component which has neither `$router` or `$i18n` resulting in an error in i18n.routing.plugin.js

For now I'm running out of idea. I suspect that it might be a problem with nuxt lifecycle between plugins / middleware but it's still obscure.

Versions:
nuxt: 1.3
nuxt-i18n: 2.2.4

RomainLK:
Fix #24 . Vue Mixin through plugins are already too far in the lifecycle to be registered on Nuxt's app. The app is in fact, already created when the plugins are registered. 

This PR extracts `localePath` and `getRouteBaseName` into factories so as to configure where to get `$router, $route, $i18n` and registers these 2 methods both in the mixin and on app.

( I don't believe anyone will use `switchLocalePath` from app... )

### DIFF
--- a/lib/templates/i18n.routing.plugin.js
+++ b/lib/templates/i18n.routing.plugin.js
@@ -1,46 +1,68 @@
 import './i18n.routing.middleware';
 import Vue from 'vue'
 
+
+function localePathFactory(i18nPath, routerPath) {
+  return function localePath(route, locale) {
+    if (!route) return
+    locale = locale || this[i18nPath].locale
+    if (!locale) return
+    // If route parameters is a string, consider it as the route's name
+    if (typeof route === 'string') {
+      route = { name: route }
+    }
+    // Build localized route options
+    const name = route.name + '-' + locale
+    const baseRoute = Object.assign({}, route, { name })
+    // Resolve localized router
+    const resolved = this[routerPath].resolve(baseRoute)
+    let { href } = resolved
+    // Cleanup href
+    href = (href.match(/^\/\/+$/)) ? '/' : href
+    return href
+  }
+}
+
+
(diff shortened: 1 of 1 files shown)


paulgv:
Awesome thank you @RomainLK !

felixdenoix:
Well in fact the switchLocalePath could be quite useful when trying to detect the user locale in the store `nuxtServerInit` hook in order to redirect him directly to the page with the correct user locale !
That could work like so
```
user goes to ==> foo.com
website default locale set to 'en'
user locale set to 'it'
user redirected to ==> foo.com/fr   (if 'fr' is configured)
(user can switch to an other language afterwards if he wishes)
```
This way you could avoid a redirection on app loading
```js
  async nuxtServerInit({ commit }, { req, app, redirect }) {
    let reqLocale = req.headers['accept-language']
      ? req.headers['accept-language'].substring(0, 2)
      : 'fr'

    let currentLocale = app.i18n.locale

    console.log('currentLocale', currentLocale)
    console.log('reqLocale', reqLocale)

    if (currentLocale !== reqLocale) {
      app.i18n.locale = reqLocale
      return redirect('/' + reqLocale)  <=== [avoid this redirection]
    }

    (........)
  }
```
Regarding of this use case don't you think it could be interesting to add `switchLocalePath`to the app ?

PR #41, opened 2018-03-11
Update vue-i18n version

manniL:
For changes see the [vue-i18n changelog](https://github.com/kazupon/vue-i18n/blob/dev/CHANGELOG.md)


paulgv:
Thanks!

PR #46, opened 2018-03-19
Avoid server error without accept-language header

manniL:
When no Accept-Language header is provided, a 500 server error is currently thrown. This should be avoided

Related: https://github.com/nuxt/nuxt.js/issues/2034#issuecomment-342602183


paulgv:
Thank you @manniL !

manniL:
Sure thing :relaxed:

PR #74, opened 2018-04-18
Fix problem for route names containing locale code

mohanklein:
If you have nested routes such as `services-development-de` (german) and  `services-development-en` (english) `services-development-de` becomes `servicesvelopment-de`.

This commit fixes that problem!


paulgv:
Thanks @mohanklein !

PR #75, opened 2018-04-29
feat(seo): Add og:locale tags

manniL:
Related proposal: https://nuxtjs.cmty.io/nuxt-community/nuxt-i18n/issues/c70

Other changes: 

* Moved return (if no i18n variable was found or route is ignored) to top


paulgv:
I'll add this to v3.x as well sometime today!

PR #91, opened 2018-05-28
feat(loadLanguageAsync): Support promises return in lang files

ezypeeze:
Based on this feature request that I opened myself: 
https://github.com/nuxt-community/nuxt-i18n/issues/60

### DIFF
--- a/docs/lazy-load-translations.md
+++ b/docs/lazy-load-translations.md
@@ -8,7 +8,7 @@ To enable translations lazy-loading, follow these 4 steps when configuring **nux
 * Set `langDir` option to the directory that contains your translation files (this can NOT be empty)
 * Configure `locales` option as an array of object, where each object has a `file` key which value is the translation file corresponding to the locale
 * Optionnaly, remove all messages that you might have passed to vue-i18n via `vueI18n` option
-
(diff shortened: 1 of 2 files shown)


paulgv:
Nice! thanks @ezypeeze !

liuhuanhui:
Hello, you can provide an example

git-antonyuk:
**Hi all! First of all, thank you for all your hard work!**

In example above we fetching all data of one language for all app, but, if we have a lot of data? 
Is there any proper way to load data, lets say some local GlobalData (Header and Footer) and data of only one page (local) wich is loading? Or maybe, better solution is to use middleware and mutations build in nuxt-i18n config (I18N_SET_MESSAGES).   

In my new project, I geting data from firebase and it works great, but project is growing and I looking for better solution for fetching data, so to say, by small piece. Code example:

```javascript 
import axios from 'axios';
import Vue from 'vue';

Vue.use(axios);

export default () => {
  return axios.get('https://bla-bla-bla.firebaseio.com/en.json')
  .then(function (response) {
    return response.data;
  })
  .catch(function (error) {
    error;
  });
}
```

desislavsd:
Hey, is there a way to provide a single function to nuxt-i18n that accepts `lang` parameter and avoid creating all the lang files?

PR #92, opened 2018-05-29
Use x-forwarded-host

darcoder:
For use x-forwarded-host

setting
differentDomains:true,
forwardedHost: true,

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+<a name="3.2.3"></a>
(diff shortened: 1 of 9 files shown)


paulgv:
Cool thanks @darcoder !
Could you add a default value (`false`) for the new option in `src/helpers/constants.js`? 
Also anyway you could add some doc about it? Or maybe give me some details on what it does and I'll try to document it myself.

Thanks!

darcoder:
I use apache and the host should be defined by the x-forwarded-host

<VirtualHost {server}>
ServerName {server}
DocumentRoot "{path}"
ProxyPass /api {server}/api 
  <Location />
    ProxyPass {nuxt_server}
    ProxyPassReverse {nuxt_server}
  </Location>
</VirtualHost>

paulgv:
Thanks @darcoder, I'll add some doc when I get the chance!

PR #101, opened 2018-06-20
Routes with children cannot have a name property.

elrolito:
This should resolve nuxt-community/nuxt-i18n#98

Sets route name *before* checking for children, and if children exist, delete the name property from the route.

cc @htchtc052


paulgv:
Thanks @elrolito !

elrolito:
Glad I could contribute.

PR #102, opened 2018-06-26
fix: support locales with names that match other locales (`en` and `en-us`)

tuqire:
Currently when an app supports multiple locales, which might match each others regex, the `getRouteBaseName` method returns the wrong result.

Example:

```
// locales = ['en', 'en-us']
// route.name = 'index___en-us'
// route.name.replace(regexp, '') will return 'index-us' and not 'index'
````


paulgv:
Thanks @tuqire !

PR #119, opened 2018-08-21
Update README.md

gangsthub:
* updated username variable to match the current one
* added `https` option by default

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,9 @@
 ```sh
 cd my-nuxt-project/
 mkdir modules/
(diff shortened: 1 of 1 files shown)


paulgv:
Hi @gangsthub 
Thanks for the PR! I like the addition of the HTTPS option.
Concerning the username variable, since the first step is to fork the project, that part shouldn't be `nuxt-community` in most cases, so maybe we should just leave `<username>` there, what do you think?

gangsthub:
Sorry about that. Totally missed that point.

paulgv:
No problem! Thanks!

PR #143, opened 2018-10-17
Set i18n_redirect cookie path to '/'

MisterEffix:
Set i18n_redirect cookie path to '/'.

Without that the cookie is created for every page...

### DIFF
--- a/src/templates/middleware.js
+++ b/src/templates/middleware.js
@@ -69,11 +69,13 @@ middleware['i18n'] = async ({ app, req, res, route, store, redirect, isHMR }) =>
         const date = new Date()
         if (isSpa) {
           Cookies.set(cookieKey, 1, {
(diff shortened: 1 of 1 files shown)


paulgv:
Hi @MisterEffix, thanks for the PR!
I wonder why we should avoid setting the cookie on all pages, is it an issue?
In theory, we would like to be able to redirect to the browser's language from any page right? Maybe we should have a way of customizing the cookie options from Nuxt's config instead, what do you think? The defaults could be let as is and you would be able to set the cookie's path and duration from the config.

appinteractive:
I think this is needed as you do not want to be sammed by cookies for every route you set the locale. You want to set the locale site-wide and not only for the current route. 

grafik

MisterEffix:
@appinteractive explained exactly what I meant.
Sorry, I should have taken the time to do it myself.

Thanks @appinteractive

Issue #114, opened 2018-08-11
Default locale & browser language detection confusion


paulgv:
### Version

[v5.0.3](https://github.com/nuxt-i18n/releases/tag/v5.0.3)

### Reproduction link

[https://github.com/nuxt-community/nuxt-i18n](https://github.com/nuxt-community/nuxt-i18n)

### Steps to reproduce

Quoting @mapeal2 
https://cmty.app/nuxt/nuxt-i18n/issues/c96

```
Given 2 locales ('en', 'fr'), default locale is 'en' and the preferred browser language is 'fr', with the 'prefixexceptdefault' option, the steps to reproduce it would be:
Visit home page '/' --> browser language is detected and redirected to '/fr', which is correct.
Close the browser.
Open a new browser and enter the home page ('/') again 
```

### What is expected ?

```
I would expect that, when opening the home page again, the application would be redirected to '/fr' instead of staying in '/'.

```

### What is actually happening?

```
 'en' locale is chosen, ignoring cookie.
```

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c116">#c116</a>)</em></sub></div>

Aperrix:
Same bug with v5.4.4 of nuxt-i18n (using the [nuxt-i18n rxample](https://github.com/paulgv/nuxt-i18n-example) by @paulgv 

```javascript
modules: ['nuxt-i18n' {
  locales: [
    { code: 'fr', iso: 'fr-FR', name: 'Français' },
    { code: 'en', iso: 'en-GB', name: 'English' },
    { code: 'de', iso: 'de-DE', name: 'Deutsch' },
    { code: 'es', iso: 'es-ES', name: 'Español' },
    { code: 'it', iso: 'it-IT', name: 'Italiano' },
    { code: 'pt', iso: 'pt-PT', name: 'Português' }
  ],
  defaultLocale: 'en',
  detectBrowserLanguage: {
    useCookie: true,
    alwaysRedirect: true
  },
  vueI18n: {
    messages: { fr, en, de, es, it, pt }
  }
}],
```
![image](https://user-images.githubusercontent.com/26674548/48317506-a56d0880-e5f3-11e8-873b-6ae896dd882f.png)

pschaub:
[...]

Issue #106, opened 2018-07-12
detectBrowserLanguage and cookies


MisterEffix:
### What problem does this feature solve?

Root path redirection

### What does the proposed changes look like?

The common way to handle the redirection from root path to the language path is :

1) check if the user already visits the website and redirect him to his latest used language by using a cookie commonly called "lang".
2) if no "lang" cookie is available, use the browser language.
3) if the browser language is unknown or unavailable, use a default language.

I have updated your "middelware.js" following this way.
Maybe it's not 100% clean, but it works.
Could you integrate this behaviour in your nuxt-i18n ?

FX

"middelware.js" modifactions :

  if (detectBrowserLanguage) {
    // Get browser language either from navigator if running in mode SPA, or from the headers
    let browserLocale = null
    if (isSpa && typeof navigator !== 'undefined' && navigator.language) {
      browserLocale = navigator.language.toLocaleLowerCase().substring(0, 2)
    } else if (req && typeof req.headers['accept-language'] !== 'undefined') {
      browserLocale = req.headers['accept-language'].split(',')[0].toLocaleLowerCase().substring(0, 2)
    }

    const localeFromRoute=getLocaleFromRoute(route, routesNameSeparator, locales);

    if (browserLocale) {
      const { useCookie, cookieKey } = detectBrowserLanguage

      const redirectToBrowserLocale = () => {
        const routeName = route && route.name ? app.getRouteBaseName(route) : 'index'
        if (browserLocale && browserLocale !== app.i18n.locale && locales.indexOf(browserLocale) !== -1) {
          redirect(app.localePath(Object.assign({}, route , {
            name: routeName
          }), browserLocale))
        }
      }

      const redirectToCookieLocale = () => {
        const routeName = route && route.name ? app.getRouteBaseName(route) : 'index'
        if (getCookie()) {
          redirect(app.localePath(Object.assign({}, route , {
            name: routeName
          }), getCookie()))
        }
      }

      const getCookie = () => {
        if (isSpa) {
          return Cookies.get(cookieKey);
        } else if (req && typeof req.headers.cookie !== 'undefined') {
          const cookies = req.headers && req.headers.cookie ? cookie.parse(req.headers.cookie) : {}
          return cookies[cookieKey]
        }
        return null
      }

      const setCookie = () => {
        const date = new Date()
        if (isSpa) {
          Cookies.set(cookieKey, localeFromRoute||browserLocale, {
[...]

Issue #86, opened 2018-05-10
Set cookie to locale that it redirected to, and always redirect (unless manually switched to different locale)


FransTwisk:
### What problem does this feature solve?

I implemented my own version of detectBrowserLanguage when I was still using an older version of nuxt-i18n and I was happy to see this feature was added in a later version! I just tried it out, and I think the current implementation can be improved, because now it will only redirect the first time, not when a user returns to the main url. 

### What does the proposed changes look like?

The cookie should be set to the initially detected language, en then always check what the (last selected) language is set to in the cookie. Then we can always redirect to the language stored in the cookie. If the user manually switched language through the `switchLocalePath` function, the cookie can be updated with the manually selected language, and also remember this manual overwrite on the next visit.

<div align="right"><sub><em>This feature request is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt-community/nuxt-i18n/issues/c81">#c81</a>)</em></sub></div>

borie88:
Doesn't seem to work though

lukaVarga:
Not working for me either. If I go to `main_url/locale`, it will use that locale and store it in the cookie. The next time I visit `main_url` (so without the `locale` suffix), it will load the default locale instead of using the one that's stored in the cookie

borie88:
Thanks. I ended up doing something similar, making a plugin:
`
import VueCookie from 'vue-cookie'

export default function ({ app }) {
  app.i18n.onLanguageSwitched = (oldLocale, newLocale) => {
    if (process.browser) {
      if (VueCookie.get('i18n_redirected') !== newLocale) {
        VueCookie.set('i18n_redirected', newLocale)
      }
    }
  }
}`

rchl:
This is now implemented. Redirection always happens with `alwaysRedirect` enabled and saved locale can be updated with `setLocaleCookie()` method.

Can be closed.

MasterHiei:
@rchl Hi, when `alwaysRedirect` is enable, can I switch the locale manually and immediately just like `switchLocalepath()` method? I used `setLocaleCookie()` but it only works on the next navigation.

rchl:
@MasterHiei You can set cookie and then call `this.$router.replace(this.switchLocalePath(locale));`

MasterHiei:
@rchl That's it! I'm foolish! Thank you for your help!

stact:
[...]

PR #148, opened 2018-10-25
Reworked the browser detection

Magnum5234:
Reworked the browser detection, so that the cookie stores the language. This cookie value is now the preferred value when selecting the browsers locale.
Added the option to redirect every time the user visits the app
Added the options for a fallback locale if no language configured matches the browsers locale
Fixed a bug which lead to redirect loops when redirecting without a cookie

There is currently the issue of how to allow for language switches by the users choice. Accidentally this works when switching language via switchLocalePath because the redirect is not applied (as it is the same). If you have an idea how to "know" in the middleware that the user changed the language explicitly, please let me know (or if you have any other solutions).

Unfortunately I was not able to run the tests locally.

This Pull-Request should (at least partially) cover #114 #106 #86

### DIFF
--- a/docs/browser-language-detection.md
+++ b/docs/browser-language-detection.md
@@ -48,3 +48,14 @@ To completely disable the browser's language detection feature, set `detectBrows
   detectBrowserLanguage: false
 }]
 ```
+
+To redirect the user every time they visit the app and keep their selected choice, enable alwaysRedirect:
+
+```js
+// nuxt.config.js
+
+['nuxt-i18n', {
+  useCookie: true,
+  alwaysRedirect: true
+}]
+```
--- a/docs/options-reference.md
+++ b/docs/options-reference.md
@@ -55,7 +55,11 @@ Here are all the options available when configuring the module and their default
     // Set to false to redirect every time
     useCookie: true,
     // Cookie name
-    cookieKey: 'i18n_redirected'
+    cookieKey: 'i18n_redirected',
+    // Set to always redirect to value stored in the cookie, not just once
+    alwaysRedirect: false,
+    // If no locale for the browsers locale is a match, use this one as a fallback 
+    fallbackLocale: null
   },
 
   // Set this to true if you're using different domains for each language
--- a/src/helpers/constants.js
+++ b/src/helpers/constants.js
@@ -33,7 +33,9 @@ exports.DEFAULT_OPTIONS = {
   rootRedirect: null,
   detectBrowserLanguage: {
     useCookie: true,
-    cookieKey: 'i18n_redirected'
+    cookieKey: 'i18n_redirected',
+    alwaysRedirect: '',
+    fallbackLocale: null
   },
   differentDomains: false,
   forwardedHost: false,
--- a/src/templates/middleware.js
+++ b/src/templates/middleware.js
@@ -33,91 +33,107 @@ middleware['i18n'] = async ({ app, req, res, route, store, redirect, isHMR }) =>
 
   // Handle browser language detection
   const detectBrowserLanguage = <%= JSON.stringify(options.detectBrowserLanguage) %>
+  const routeLocale = getLocaleFromRoute(route, routesNameSeparator, locales)
+
+  const getCookie = () => {
+    if (isSpa) {
+      return Cookies.get(cookieKey);
+    } else if (req && typeof req.headers.cookie !== 'undefined') {
+      const cookies = req.headers && req.headers.cookie ? cookie.parse(req.headers.cookie) : {}
+      return cookies[cookieKey]
+    }
+    return null
+  }
+
+  const setCookie = (locale) => {
+    const date = new Date()
+    if (isSpa) {
+      Cookies.set(cookieKey, locale, {
+        expires: new Date(date.setDate(date.getDate() + 365))
+      })
+    } else if (res) {
+      const redirectCookie = cookie.serialize(cookieKey, locale, {
+        expires: new Date(date.setDate(date.getDate() + 365))
+      })
+      res.setHeader('Set-Cookie', redirectCookie)
+    }
+  }
+
+  const { useCookie, cookieKey, alwaysRedirect, fallbackLocale } = detectBrowserLanguage
+
+  const switchLocale = async (newLocale) => {
+    // Abort if different domains option enabled
+    if (app.i18n.differentDomains) {
+      return
+    }
+
+    // Abort if newLocale did not change
+    if (newLocale === app.i18n.locale) {
+      return
+    }
+
+    const oldLocale = app.i18n.locale
+    app.i18n.beforeLanguageSwitch(oldLocale, newLocale)
+    if(useCookie) {
+      setCookie(newLocale)
+    }
+    // Lazy-loading enabled
+    if (lazy) {
+      const { loadLanguageAsync } = require('./utils')
+      const messages = await loadLanguageAsync(app.i18n, newLocale)
+      app.i18n.locale = newLocale
+      app.i18n.onLanguageSwitched(oldLocale, newLocale)
+      syncVuex(locale, messages)
+    } else {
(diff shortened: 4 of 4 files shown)


manniL:
@Magnum5234 Does that PR introduce any breaking changes?

Magnum5234:
It should work the same as before if you don't use the new configuration values.

manniL:
@Magnum5234 Many thanks for your contribution! 🙏

Magnum5234:
Your are welcome.
If you have an idea about to tackle the language switch issue I would try to fix that, too :)

manniL:
@Magnum5234 Could you provide more details regarding the problem?

Magnum5234:
> There is currently the issue of how to allow for language switches by the users choice. Accidentally this works when switching language via switchLocalePath because the redirect is not applied (as it is the same). If you have an idea how to "know" in the middleware that the user changed the language explicitly, please let me know (or if you have any other solutions).

This is what I wrote initially. 

If you set ```alwaysRedirect``` to _true_ every time the middleware is executed the language will be confirmed that it is up to date with whats stored in the cookie - otherwise redirect. Meaning that the user cannot change the language changing the url. 
If the user triggers the language switch via ```switchLocalePath``` nuxt link the middleware is called with the route still having the old language and the updated locale object. This does trigger the redirect but since the route is not updated yet it does not do anything and in the end it seems to work out. At least that is my understanding. (writing from memory)
There are two ways how to handle the problem
1) ```switchLocalePath``` should also change the cookie - How do you get the context like spa in there?
2) in the middleware check if it's a user triggered language switch and don't try to redirect - how would you do this?

Hope this clarifies things :)

manniL:
That clarifies a lot :relaxed: 

1. The function already accesses the ctx (https://github.com/nuxt-community/nuxt-i18n/blob/513cbf92f9aca614b62c37a96c7476941d0ccf4d/src/plugins/routing.js#L54) once :relaxed: 

2. Probably through the Vuex store. Setting a flag or something there when the language is switched and let the middleware work and unset the flag again

Aperrix:
Hi @Magnum5234 and @manniL, thank you for taking your time to make sure everything works properly !

Being webdesigner i do not quite have the necessary skill to propose a bugfix myself but i take this opportunity to try to improve myself because I need it to work properly as soon as possible to push a new release in production 😓

I hope you find a solution before me because I may take a long time to get there by myself 😄 
so .. thanks again 😃

Love you

manniL:
@Magnum5234 First one sounds suitable, yup!

Magnum5234:
The mentioned redirect bug does not exist, here's why:

Upon clicking on the nuxt-link the middleware is called

The current locale is still the same as before, so is the browser detection value (usually the stored cookie). The route locale on the other hand has the new language.
The alwayRedirect will therefore not trigger because current locale and the browser detection value are still the same
If those are the same and the route locale is set, the locale will be switched to the route locale - and changing the cookie value

examples:

```
GET http://sandbox:3000/
locale: de routeLocale: de cookieValue: de
cookieValue === locale de === de
-> switchLocale(routeLocale) (no change as locale is the same as routeLocale)
```
```
GET http://sandbox:3000/en
locale: en routeLocale: en cookieValue: de
cookieValue !== locale de !== en
-> switchLocale(cookieValue) + redirect

(2nd middleware call)
locale: de routeLocale: de cookieValue: de
cookieValue === locale de === de
-> switchLocale(routeLocale) (no change as locale is the same as routeLocale)
```
```
GET http://sandbox:3000/
(see above)

(switchLocale to en)
locale: de routeLocale: en cookieValue: de
cookieValue === locale de === de
-> switchLocale(routeLocale)
```
```
GET http://sandbox:3000/
(see above)

(switchLocale to /en/secondPage)
locale: de routeLocale: en cookieValue: de
cookieValue === locale de === de
-> switchLocale(routeLocale)
```

(Thanks @manniL for helping me find this not issue)

In the meantime I found some bogus lines (will make a pull request for it tomorrow)

https://github.com/nuxt-community/nuxt-i18n/blob/master/src/templates/middleware.js#L138
(copy paste error, should be `routeLocale ? routeLocale : locale`)

https://github.com/nuxt-community/nuxt-i18n/blob/master/src/templates/middleware.js#L119
(this setCookie is not necessary at it will be set in switchLocale which will be called every time the locale changes)

Magnum5234:
The mentioned redirect bug does not exist, here's why:

Upon clicking on the nuxt-link the middleware is called

The current locale is still the same as before, so is the browser detection value (usually the stored cookie). The route locale on the other hand has the new language.
The alwayRedirect will therefore not trigger because current locale and the browser detection value are still the same
If those are the same and the route locale is set, the locale will be switched to the route locale - and changing the cookie value

examples:

```
GET http://sandbox:3000/
locale: de routeLocale: de cookieValue: de
cookieValue === locale de === de
-> switchLocale(routeLocale) (no change as locale is the same as routeLocale)
```
```
GET http://sandbox:3000/en
locale: en routeLocale: en cookieValue: de
cookieValue !== locale de !== en
-> switchLocale(cookieValue) + redirect

(2nd middleware call)
locale: de routeLocale: de cookieValue: de
cookieValue === locale de === de
-> switchLocale(routeLocale) (no change as locale is the same as routeLocale)
```
```
GET http://sandbox:3000/
(see above)

(switchLocale to en)
locale: de routeLocale: en cookieValue: de
cookieValue === locale de === de
-> switchLocale(routeLocale)
```
```
GET http://sandbox:3000/
(see above)

(switchLocale to /en/secondPage)
locale: de routeLocale: en cookieValue: de
cookieValue === locale de === de
-> switchLocale(routeLocale)
```

(Thanks @manniL for helping me find this not issue)

In the meantime I found some bogus lines (will make a pull request for it tomorrow)

https://github.com/nuxt-community/nuxt-i18n/blob/master/src/templates/middleware.js#L138
(copy paste error, should be `routeLocale ? routeLocale : locale`)

https://github.com/nuxt-community/nuxt-i18n/blob/master/src/templates/middleware.js#L119
(this setCookie is not necessary at it will be set in switchLocale which will be called every time the locale changes)

HapLifeMan:
Hi @Magnum5234, I though your PR fixed this issue because it's a big part of the problem I have, but that's not the case, do you know what happened? https://github.com/nuxt-community/nuxt-i18n/issues/166

manniL:
@HapLifeMan Be aware that this PR is not included it any release yet

HapLifeMan:
@manniL Yes I used the commit reference 😊

BroTrue:
@manniL is this PR on release ? Because I still go same issue.  **"nuxt-i18n": "5.8.0",**

My `nuxt.config.js`
```
...
  modules: [

    ['nuxt-i18n', {
      locales: [
          { code: 'ru', iso: 'ru-RU'},
        { code: 'en', iso: 'en-US' },
        { code: 'ua', iso: 'ua-UA' }
        ],
      defaultLocale: 'ru',
      detectBrowserLanguage: {
        useCookie: true,
        cookieKey: 'i18n_redirected',
      alwaysRedirect: true
      },
      vueI18n: {
        fallbackLocale: 'ru',
        messages: {
          'ru': require('./locales/ru.json'),
          'en': require('./locales/en.json'),
          'ua': require('./locales/ua.json')
        }
      },
    }],
...
```

manniL:
@BroTrue should be, yup. 
image

BroTrue:
Maybe @Magnum5234 can help me with my problem ? Also i noticed that `switchLocalePath()` is changing local at Vuex, but NOT changing cookies value.

Magnum5234:
@BroTrue Do you mean the same issue as @HapLifeMan is having? I will have a look at this.

@HapLifeMan Sorry I did not see the post in December, missed the email somehow. I will have a look at this asap.

manniL:
Better talk about that in a new issue guys

BroTrue:
Ok, I will create it soon.

Issue #144, opened 2018-10-20
Nuxt i18n triggers vue-meta getMetaInfo for every component created/destroyed


adrienbaron:
### Version

[v5.4.3](https://github.com/nuxt-i18n/releases/tag/v5.4.3)

### Reproduction link

[https://codesandbox.io/s/github/adrienbaron/nuxt-i18n-performance-issue](https://codesandbox.io/s/github/adrienbaron/nuxt-i18n-performance-issue)

### Steps to reproduce

- Open Chrome dev tools
- Go to the performance tab.
- Click on record in the performance tab
- Toggle the list to create items
- Stop recording

You'll notice that getMetaInfo from `vue-meta` was triggered by the creation of BasicListItems. They do not have meta info so this should not happen.

### What is expected?

`getMetaInfo` from vue-meta should not be triggered when components without meta info are created.

### What is actually happening?

`getMetaInfo` is triggered by the creation of any component.

### Additional comments?

The reason this seems to be happening is that Nuxt i18n registers this mixin globally: https://github.com/nuxt-community/nuxt-i18n/blob/master/src/plugins/seo.js#L3.

The mixin registers the `head()` hook for every single component. This is the hook used by Nuxt for page meta info in vue-meta.
The hook method checks whether `this._hasMetaInfo` is true for the component, but it is always true as `this._hasMetaInfo` is set to true if the `head` hook is defined. (see: https://github.com/declandewet/vue-meta/blob/master/src/shared/plugin.js#L47 and https://github.com/nuxt/nuxt.js/blob/dev/lib/core/meta.js#L17).

This has a big impact on performance on big pages as any components created triggers a recursive search for meta from the root component. And for each component the alternates meta links are computed and merged with the ones from the parent.

A small mitigation is setting `seo: false`, but you still get the recursion, and you loose the alternate links.

I first thought this was an issue in `vue-meta` and filed an issue there: https://github.com/declandewet/vue-meta/issues/278.

I'm open to work on a fix and do a pull request, something like managing to only register this hook on nuxt pages component for example could be a solution.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c156">#c156</a>)</em></sub></div>

adrienbaron:
Ok here is what I've looked into so far:
[...]

PR #153, opened 2018-11-06
Fix/perf issue hook only nuxt

adrienbaron:
This is a patch for issue #144. I'm not sure this is the best solution to the problem, but I tested on my (rather big) project and did see a big performance improvement.
I'm opening this PR as suggested by @manniL to get the conversation going 👍


adrienbaron:
Seen with @manniL for another solution (see #154)

PR #154, opened 2018-11-07
Expose the head SEO function to be used in layouts

adrienbaron:
This PR exposes the SEO head hook function as `this.$nuxtI18nSeo` on components.

This allows to disable seo head mixin on all components (by setting `seo: false` in nuxt i18n options).
Then it's possible to do this in your app layout (or wherever):
```js
    head () {
      return this.$nuxtI18nSeo()
    },
```
This greatly improves performance as SEO links are not recomputed for every components but only for the layout itself.

This changes is not breaking as the default global registration with the same behaviour remains if `seo: true` in the configuration.

Resolves: #144

### DIFF
--- a/src/plugins/main.js
+++ b/src/plugins/main.js
@@ -1,9 +1,10 @@
 import Vue from 'vue'
 import VueI18n from 'vue-i18n'
+import { nuxtI18nSeo } from './seo-head'
 
 Vue.use(VueI18n)
 
-export default async ({ app, route, store, req }) => {
+export default async ({ app, route, store, req }, inject) => {
   // Options
   const lazy = <%= options.lazy %>
   const vuex = <%= JSON.stringify(options.vuex) %>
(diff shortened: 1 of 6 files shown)


paulgv:
Thanks @adrienbaron !
Could you also update the documentation to explain how to use the `nuxtI18nSeo()` method?

adrienbaron:
@paulgv Done!
Btw thanks a lot to @manniL for his help on this, forgot to mention, and couldn't have done it without his help 👍

manniL:
Happy we found a solution 🙊  (Also one that can be default at the next major version)

src/plugins/main.js
```diff
@@ -1,9 +1,10 @@
 import Vue from 'vue'
 import VueI18n from 'vue-i18n'
+import { nuxtI18nSeo } from './seo-head'
 
 Vue.use(VueI18n)
 
-export default async ({ app, route, store, req }) => {
+export default async ({ app, route, store, req }, inject) => {
```
manniL: we can omit that `inject` param now ☺️ @adrienbaron

kevinmarrec:
@adrienbaron @manniL Last commit on `master` seems to be related to this PR, and tests are not passing anymore, any ideas ? (CircleCI build fails)

manniL:
@kevinmarrec Should be just the vue-ssr id(`4eb03412:0`). Will fix

manniL:
@kevinmarrec fixed

manniL:
@harmnot There hasn't been a release yet

harmnot:
I was asking this https://cmty.app/nuxt/nuxt-i18n/issues/c176

harmnot:
then, I try to find out, and follow the instruction https://github.com/nuxt-community/nuxt-i18n/blob/master/docs/seo.md here and after that , show the $nuxtI18nSeo is not a function

paulgv:
Sorry for the delay, this wasn't released until recently in version `5.5.0`: https://github.com/nuxt-community/nuxt-i18n/blob/master/CHANGELOG.md#550-2019-01-14

PR #158, opened 2018-11-19
Check if there is a locale available

### LINKED ISSUE #131: Cannot read property 'iso' of undefined
### Version

### Reproduction link

[https://github.com/Livog/nuxt-i18n-bug](https://github.com/Livog/nuxt-i18n-bug)

### Steps to reproduce

Node Version: 9.7.1
NPM Version: 6.4.1
Nuxt Version: 2.1.0
Nuxt i18n Version: 5.3.0

I m getting this error whenever I install the repo that I linked.
TypeError: Cannot read property 'iso' of undefined

### What is expected ?

No bugs.

This does not seem to cause any problem, just worried it will cause a problem later.

### Additional comments?

I have noticed that I did set parsePages to true I think and then I stopped happening but not sure if that's the case. I have tested it on my mate's computer and he got the same error. This is a very scaled down version of what I m currently working on but it got the same dependencies. If there are parts missing going to API that doesn't work I m sorry.

I renamed the package-lock since my git did not want to push it, don't know why. Just rename it back to lock dependencies. 

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c142">#c142</a>)</em></sub></div>

dipsaus9:
Fixes #131. Double check if value is not null to prevent error
`Cannot read property 'iso' of undefined` on line 54 of the seo-head file.

### DIFF
--- a/src/templates/seo-head.js
+++ b/src/templates/seo-head.js
@@ -5,65 +5,76 @@ export const nuxtI18nSeo = function () {
     !this.$i18n ||
     !this.$i18n.locales ||
     this.$options[COMPONENT_OPTIONS_KEY] === false ||
-    (this.$options[COMPONENT_OPTIONS_KEY] && this.$options[COMPONENT_OPTIONS_KEY].seo === false)
+    (this.$options[COMPONENT_OPTIONS_KEY] &&
+      this.$options[COMPONENT_OPTIONS_KEY].seo === false) ||
+    !this.$i18n.locale
   ) {
-    return {};
-  }
-  const LOCALE_CODE_KEY = '<%= options.LOCALE_CODE_KEY %>'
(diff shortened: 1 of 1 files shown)


paulgv:
Hi @dipsaus9 
Really sorry for checking your PR so late.
It looks like a big change and I don't have enough time to check them thoroughly, any help in reviewing this would be much appreciated!

dipsaus9:
First error was created after turning on the `SEO` option. Looks like the variable `this.$i18n.locale` wasn't being stored. 

Couldn't reproduce the same error again after updating to 5.5. 
Closing this issue for now

PR #159, opened 2018-11-26
Update routes.js

Seranth:
This solves the problem where the console receives one warning per route for duplicate routes added, and changes the behavior of nuxt-link from going to the "clean" default path of e.g. /contact to always resolving to e.g. /en/contact

### DIFF
--- a/src/helpers/routes.js
+++ b/src/helpers/routes.js
@@ -91,7 +91,7 @@ exports.makeRoutes = (baseRoutes, {
       )
 
       if (locale === defaultLocale && strategy === STRATEGIES.PREFIX_AND_DEFAULT) {
-        routes.push({ ...localizedRoute, path })
+        routes.push({ ...route, path })
       }
 
       if (shouldAddPrefix) {


paulgv:
Hi @Seranth 
Sorry for the long delay before checking your PR.
I'm concerned that this change might break some of the routing features since `localizedRoute` contains, at the very least, the generated localized name for the base route. If I understand the issue correctly, we might want to delete the base routes all together once we're done generating the localized routes, any thoughts?

Seranth:
I've been pretty busy so haven't worked with this for a while, but I've picked it up again now that I saw your message.

My current ideas honestly is to possibly make my own version of this project (open source of course) that does exactly what I want it to, the deeper I delve into the code the more lost I feel. Essentially the functionality I'm looking for is that "/" is the path you'd get for your detected preferred language as well as having the possibility of changing the language in i18n, triggering loadLanguage to add that language to the "/" version of the site as your new preferred language. Then having e.g. "/en" and "/sv" etc to override the preferred language. That way a user visits the site, it identifies your preferred language, then when you select another language you can remain on the clean path but with any of the languages, and if you want a specific language for a link you can visit that locales path. The fallback locale would always be loaded since it's the fallback, but the default locale would be the identified one.

As far as I can tell the exact functionality I want (apart from being difficult to put into words) is something that's proven pretty tricky to properly implement into this project without some fairly big changes.

I'm gonna try for a bit longer before I decide that it's just easier to make my own project instead.

paulgv:
This PR has been inactive for a while, closing it for now.

Issue #161, opened 2018-11-28
docs: don't suggest to provide a function for lazy loading


manniL:
### What problem does this feature solve?

related: https://github.com/nuxt/nuxt.js/issues/4424
https://nuxt-community.github.io/nuxt-i18n/lazy-load-translations.html

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c172">#c172</a>)</em></sub></div>

felixdenoix:
In my case an external API is providing me all of the website texts and placeholders and thus, I found that fetching this data asynchronously can be useful. 
The issue I'm facing right now is that I cannot use the installed nuxt-Axios module to fetch these datas and that I had to install the basic Axios module to be able to fetch these.

manniL:
@michaelize Awesome!

@felixdenoix can you second this? ☺️

felixdenoix:
@manniL Seems ok  for me !

PR #163, opened 2018-12-05
Wait for lazy loading promises

Alex--C:
According to the docs, lazy loaded locales should be able to return a promise. Doing this creates a warning message ever since nuxt switched to https://github.com/Rich-Harris/devalue .

This pull request prevents this warning. If lazy loading returns a promise, it waits for it before returning from `loadLanguageAsync` and returns the resolved value instead.
Resolves https://github.com/nuxt/nuxt.js/issues/4424 . Related: #161 .

### DIFF
--- a/src/templates/utils.js
+++ b/src/templates/utils.js
@@ -20,9 +20,10 @@ export async function loadLanguageAsync (i18n, locale) {
         try {
           const module = await import(/* webpackChunkName: "lang-[request]" */ '~/<%= options.langDir %>' + file)
           const messages = module.default ? module.default : module
-          i18n.setLocaleMessage(locale, typeof messages === 'function' ? await Promise.resolve(messages()) : messages)
+          let result = typeof messages === 'function' ? await Promise.resolve(messages()) : messages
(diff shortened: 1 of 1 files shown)


src/templates/utils.js
```diff
@@ -20,9 +20,10 @@ export async function loadLanguageAsync (i18n, locale) {
         try {
           const module = await import(/* webpackChunkName: "lang-[request]" */ '~/<%= options.langDir %>' + file)
           const messages = module.default ? module.default : module
-          i18n.setLocaleMessage(locale, typeof messages === 'function' ? await Promise.resolve(messages()) : messages)
+          let result = typeof messages === 'function' ? await Promise.resolve(messages()) : messages
```
falkodev: `const result` is better I guess

Alex--C: Thanks for the feedback. I changed that.

paulgv:
Thanks @Alex--C and sorry for the delay!

Alex--C:
> Thanks @Alex--C and sorry for the delay!

No problem - thank you for the merge! :D

PR #164, opened 2018-12-07
Fixed minor issues in the browser detection

Magnum5234:
For reference please see https://github.com/nuxt-community/nuxt-i18n/pull/148


paulgv:
@Magnum5234 woops just noticed your PR! Good catch! Publishing your fixes now!

PR #165, opened 2018-12-07
[WIP] prefix_none strategy

appinteractive:
I need to be able to switch the language without a prefix as the locale is only used for the interface, not for the content (which can be multilingual).

Therefore I startet to add a `PREFIX_NONE` strategy.

> I would be happy for some feedback

## Problems to solve:
- [x] do not prefix links
- [ ] fix locale switch (currenty does not work on the root route `/`)
- [x] write tests for the `PREFIX_NONE` strategy (found in `test/fixture/prefix_none/`)
- [ ] more link tests
- [ ] clean up code

## How to test
- **in browser:** `node test/fixtures/prefix_none/start.js` or `node test/fixtures/basic/start.js`
- **run tests:** `yarn jest test/fixtures/prefix_none` or `yarn jest test/fixtures/basic`

### DIFF
--- a/src/templates/middleware.js
+++ b/src/templates/middleware.js
@@ -10,6 +10,7 @@ middleware['i18n'] = async ({ app, req, res, route, store, redirect, isHMR }) =>
   // Options
   const lazy = <%= options.lazy %>
   const vuex = <%= JSON.stringify(options.vuex) %>
+  const strategy = '<%= options.strategy %>'
   const differentDomains = <%= options.differentDomains %>
   const isSpa = <%= options.isSpa %>
 
@@ -21,8 +22,7 @@ middleware['i18n'] = async ({ app, req, res, route, store, redirect, isHMR }) =>
   const locales = getLocaleCodes(<%= JSON.stringify(options.locales) %>)
   const syncVuex = <%= options.syncVuex %>
 
-
-  let locale = app.i18n.locale || app.i18n.defaultLocale || null
+  let locale = app.i18n.locale || app.i18n.defaultLocale
 
   // Handle root path redirect
   const rootRedirect = '<%= options.rootRedirect %>'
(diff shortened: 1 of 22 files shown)


src/templates/middleware.js
```diff
@@ -135,5 +137,21 @@ middleware['i18n'] = async ({ app, req, res, route, store, redirect, isHMR }) =>
...
+  console.log('route.path', route.path)
+  console.log('app.i18n.locales', app.i18n.locales)
+
+  const routePathLang = route.path.slice(1).split('/').shift()
+  console.log('routePathLang', routePathLang)
+
+  if (routePathLang && routePathLang.length === 2) {
+    console.log('REDIRECTING TO ' + (route.path.slice(3) || '/'))
+    locale = routePathLang
+    app.i18n.switchLocale(locale)
```
Magnum5234: "app.i18n.switchLocale is not a function"

appinteractive: 🤔 hmm that could be explain why it does not switch on the root route

Magnum5234: 2018-12-07 16_30_22-start
I recommend using an IDE that is highlighting such errors in advance :)

appinteractive:
So I guess we need more tests.

src/templates/middleware.js
appinteractive: @Magnum5234 which editor are you using?

Magnum5234: I use fall all web related stuff PHP Storm
https://www.jetbrains.com/phpstorm/

JakubKoralewski:
What seems to currently be the problem? This PR kind of died. I'd love to help.

appinteractive:
The problem is that the current strategies do not fit for web apps with multi lang content, only websites. As for us the Lang setting only effects the ui language, not the content we don’t want to have any reference to the current Lang in the url and therefore need to switch the Lang also in place without switching the url. 

All of that is currently not possible and turned out to be very error prone in the current implementation as the tests also are more of an decoration then really useful atm. 

We gave up and implemented it on our own which turned out to be very simple. So at the Moment that PR is indeed dead but feel free to take it further if it is useful also for your case. Would be nice if the defacto standard community plugin would also work for multi Lang content sites.

paulgv:
This PR has been inactive for a while, closing it for now.

Hopefully, someone can pick this up, it would certainly be a nice feature to have.

PR #167, opened 2018-12-09
Update types

iliyaZelenko:
That's what RawLocation is:

```ts
export type RawLocation = string | Location;

export interface Location {
  name?: string;
  path?: string;
  hash?: string;
  query?: Dictionary<string | string[]>;
  params?: Dictionary<string>;
  append?: boolean;
  replace?: boolean;
}
```
This type, for example, is used in the "push" method:

```ts
push (location: RawLocation, onComplete?: Function, onAbort?: ErrorHandler): void;
```

Without this fix, I had a error:

image

### DIFF
--- a/types/vue.d.ts
+++ b/types/vue.d.ts
@@ -3,12 +3,12 @@
  */
 
 import Vue from "vue";
-import { Route } from "vue-router";
+import { RawLocation } from "vue-router";
 
 declare module "vue/types/vue" {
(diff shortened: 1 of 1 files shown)


iliyaZelenko:
These changes will not spoil anything, but only allow us to use the necessary route object.

paulgv:
Thanks @iliyaZelenko !

PR #168, opened 2018-12-09
Update main.js

iliyaZelenko:
This will take [this method](https://kazupon.github.io/vue-i18n/api/#methods) to the app object, it will be convenient to use on the server.

For, example, instead of this:
```js
app.i18n.t('posts_page_error_404')
```
can be used:
```js
app.$t('posts_page_error_404')
```


paulgv:
Thanks @iliyaZelenko ! Sorry for taking so long to check the PRs.

PR #169, opened 2018-12-10
rootRedirect's redirect does not preserve query

SnirShechter:
**Case:** user enters the website on root path ( '/' ) with some query params. a common case of this is tracking query params.

**Current Behavior:** query params are lost when redirected to rootRedirect

**Desired Behavior:** user will redirect to the rootRedirect URL with the query params he had

This fixes the problem and preserves the query.


paulgv:
Thanks @SnirShechter !

PR #172, opened 2018-12-24
 fix duplicate route names with prefix_and_default strategy

souvenir:
fix duplicate route names with prefix_and_default strategy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+<a name="5.5.1"></a>
+## [5.5.1](https://github.com/nuxt-community/nuxt-i18n/compare/v5.5.0...v5.5.1) (2018-12-24)
+
+
+
 <a name="5.5.0"></a>
 # [5.5.0](https://github.com/nuxt-community/nuxt-i18n/compare/v5.4.4...v5.5.0) (2018-12-24)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+<a name="5.5.0"></a>
+# [5.5.0](https://github.com/nuxt-community/nuxt-i18n/compare/v5.4.4...v5.5.0) (2018-12-24)
+
+
+### Features
+
+* expose head SEO function to use in layout ([#154](https://github.com/nuxt-community/nuxt-i18n/issues/154)) ([ce373c4](https://github.com/nuxt-community/nuxt-i18n/commit/ce373c4))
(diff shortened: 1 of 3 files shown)


src/helpers/routes.js
```diff
@@ -91,7 +91,7 @@ exports.makeRoutes = (baseRoutes, {
       )
 
       if (locale === defaultLocale && strategy === STRATEGIES.PREFIX_AND_DEFAULT) {
-        routes.push({ ...localizedRoute, path })
+        routes.push({ ...route, path })
```
rchl: This will change behavior of localePath/switchLocalePath. For example, with 'en' being default language ,`this.localePath('index', 'en')` will always resolve to URL with language code prefix added.

rchl: Which would also break this recent addition: dcd1f7957039ad34f57f822be6bec1f7d4742bd2

CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
...
+# [5.5.0](https://github.com/nuxt-community/nuxt-i18n/compare/v5.4.4...v5.5.0) (2018-12-24)
+
+
+### Features
+
+* expose head SEO function to use in layout ([#154](https://github.com/nuxt-community/nuxt-i18n/issues/154)) ([ce373c4](https://github.com/nuxt-community/nuxt-i18n/commit/ce373c4))
+* rework browser detection and save lang to cookie ([#148](https://github.com/nuxt-community/nuxt-i18n/issues/148)) ([d1bbc84](https://github.com/nuxt-community/nuxt-i18n/commit/d1bbc84))
+
+
+
```
rchl: I don't think you meant to include these changes.

src/helpers/routes.js
paulgv: @rchl agreed, this change would probably break some things in the module since `localizedRoute` contains some important data for the custom routing to work properly.
If I understand correctly, the problem here is that original routes are preserved while and duplicated for the default language, so maybe we should consider deleting all original routes once the custom routes are generated?

rchl: > maybe we should consider deleting all original routes once the custom routes are generated

Would that make routes with prefix not work anymore?

I mean, default language being `en`, both `site.com/` and `site.com/en` should work.

paulgv: Oh yeah you're right, I had forgotten what `prefix_and_default` is supposed to do. Well maybe we should simply change the way route names are generated by adding an additional suffix on the duplicated default routes names.

rchl: I don't know code well enough to tell if that's the best approach but I was actually thinking of same solution. Basically name them `index__en__2` rather than `index__en`. Of course some logic would need to change to ignore that extra part, as the code is parsing that name to figure out target name and language.

paulgv: I don't remember exactly how this whole thing works but maybe we could prepend an additional fragment to the suffix, rather than appending it, it might result in the duplication issue being resolved without requiring any change in the logic. ie: `index__2__en` and `index__en`

rchl: That last suggestion would also break in a same way as this change.

`localePath` computes path based on passed-in route name so it wouldn't find that alternate index___2___en that has path without locale and would instead find the one with locale so it would return `en/` rather than just `/`.

rchl: Created PR #230 that should fix this problem without creating any extra issues.

paulgv:
Closing in favor of https://github.com/nuxt-community/nuxt-i18n/pull/230

PR #174, opened 2018-12-28
Allow i18n component to load json

robyedlin:
Adds support for loading json in i18n component with `type: 'javascript/auto'` in webpack config.

http://kazupon.github.io/vue-i18n/guide/sfc.html#webpack


paulgv:
Thanks @robyedlin !

PR #175, opened 2018-12-30
New strategy NO_PREFIX

JakubKoralewski:
### What is the problem:

I want to have a custom URL path for every link, so I want to have something like this:
* /kontakt
* /contact

But instead it seems like the only option nuxt-i18n gives me is:
* /kontakt
* /en/contact

### How I fixed it:

I added a new strategy called **NO_PREFIX**. It is chosen by the user in the nuxt config file with **'no_prefix'**. I've also updated some docs to reflect the changes.

Suggested here:
* https://cmty.app/nuxt/nuxt-i18n/issues/c166
* https://cmty.app/nuxt/nuxt-i18n/issues/c184 (by me, excerpt used to explain)


JakubKoralewski:
Duplicate of #165

Issue #173, opened 2018-12-27
Cannot set property 'locale' of undefined


willin:
### Version

[v5.3.0](https://github.com/nuxt-i18n/releases/tag/v5.3.0)

### Reproduction link

- demo: [https://willin.guru](https://willin.guru)
- source: https://github.com/willin/willin.github.io

### Steps to reproduce

when development mode it's ok, but when built and deployed it shown error.

### What is expected ?

works properly

### What is actually happening?

```
c6b25ed7d1e571a9b85f.js:1 [nuxt] Error while initializing app TypeError: Cannot set property 'locale' of undefined
    at f.<anonymous> (c6b25ed7d1e571a9b85f.js:1)
    at 33b42310abd05a803713.js:2
    at 33b42310abd05a803713.js:2
    at Array.forEach (<anonymous>)
    at 33b42310abd05a803713.js:2
    at f._withCommit (33b42310abd05a803713.js:2)
    at f.commit (33b42310abd05a803713.js:2)
    at f.commit (33b42310abd05a803713.js:2)
    at o.commit (33b42310abd05a803713.js:2)
    at f.setLocale (c6b25ed7d1e571a9b85f.js:1)
```

source code: https://github.com/nuxt-community/nuxt-i18n/blob/master/src/plugins/main.js#L41

<img width="1078" alt="wx20181230-005037 2x" src="https://user-images.githubusercontent.com/1890238/50540455-f6998180-0bcc-11e9-9012-5aa5901880fd.png">

state got `undefined`

### Additional comments?

Config:

```js
  modules: [
    ['nuxt-i18n', {
      locales: [
        { name: '简体中文', code: 'zh', iso: 'zh-CN', file: 'zh.js' },
        { name: 'English', code: 'en', iso: 'en-US', file: 'en.js' }
      ],
      lazy: true,
      langDir: 'locales/',
      strategy: 'prefix',
      defaultLocale: 'zh',
      detectBrowserLanguage: {
        useCookie: true,
        cookieKey: 'i18n_redirected'
      },
      rootRedirect: 'en',
      vueI18n: {
        fallbackLocale: 'en'
      }
    }]
  ],
```

deployed files: https://dev.tencent.com/u/willin/p/willin.wang/git/tree/master

- ROOT
 - 404.html
 - zh/ folder
 - en/ folder 

[...]

PR #178, opened 2019-01-07
Fix state not defined issue in #173

michaelize:
This PR fixes https://github.com/nuxt-community/nuxt-i18n/pull/173

The problem occurs when deploying with SSR to a static host, such as AWS S3, then access `https://example.com/en`. Since there is no matching file,  Nuxt loads the fallback file `https://example.com/200.html`.

Now due to there is no server rendered state value defined(https://nuxtjs.org/guide/vuex-store#modules-mode), client preserves the state without the i18n module state, which raised an exception later when trying to set the locale.

The fix is to set `{ preserveState: false }` by default. If need to pass value from the server, then set`preserveState` to true in the options, and create a module vuex file as described here: https://nuxtjs.org/guide/vuex-store#modules-mode

### DIFF
--- a/docs/options-reference.md
+++ b/docs/options-reference.md
@@ -58,7 +58,7 @@ Here are all the options available when configuring the module and their default
     cookieKey: 'i18n_redirected',
     // Set to always redirect to value stored in the cookie, not just once
     alwaysRedirect: false,
-    // If no locale for the browsers locale is a match, use this one as a fallback 
+    // If no locale for the browsers locale is a match, use this one as a fallback
     fallbackLocale: null
   },
 
@@ -84,7 +84,7 @@ Here are all the options available when configuring the module and their default
   vuex: {
     // Module namespace
     moduleName: 'i18n',
-    
+
     // Mutations config
     mutations: {
       // Mutation to commit to store current locale, set to false to disable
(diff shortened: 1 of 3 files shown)


priezz:
I would like to note, that #173 occures even when the SSR is not meant at all, i.e. in SPA nuxt mode.

michaelize:
@priezz You are correct. This is not unique to SSR. It's how Vuex implements preserveState option. When preserveState is set to true, it skips adding state for the module since it assumes the state value already exists, as it can be passed from SSR.

Here's the explanation about Vuex preserveState; https://github.com/vuejs/vuex/issues/1130#issuecomment-356612119

paulgv:
Hi and thanks for the PR!
Should I assume that merging this would fix the issue for both SSR and SPA modes?
Also, wouldn't it make more sense to set `preserveState` default value to `true` instead of `false`? Let me know.

Sorry for the delay in checking the PR.

michaelize:
@paulgv Sorry for the late response.

TLDR: it works for both cases.

The plugin became broken due to a Nuxt change mentioned in one of the issues you posted here: https://github.com/nuxt/nuxt.js/issues/3757#issuecomment-422437601. Nuxt now overrides the store.registerModule method for universal apps after this PR https://github.com/nuxt/nuxt.js/pull/3909/files#diff-c1b82c7b8931d52d83e399952282cbc2R52

When preserveState is set to false, it simply initiates with local state. Since this module doesn't have any default server-rendered state, it should work fine in either case.

The reason @priezz is having this issue even when SSR is not being used is very likely due to the mode is set to universal. Pls let me know if this is not correct.

This problem was also identified on Nuxt repo: https://github.com/nuxt/nuxt.js/issues/4370

priezz:
@michaelize Mode is set to 'spa' in my case.

michaelize:
@priezz I couldn't reproduce this issue under spa mode. Can you create a sample repo?

PR #192, opened 2019-01-18
Fixed link to CMTY.

timosperisen:
Link was pointing to non-existing page. Added query parameter to make it point to questions.

If it should point to all kinds of issues then perhaps even https://cmty.app/nuxt/nuxt-i18n/ would be even better as link location.


paulgv:
Thanks!

PR #193, opened 2019-01-18
Creates correct baseroute for switchlocalepath

dipsaus9:
Old version didn't create correct object for the `router.resolve` by Vue. 
Params should start with a 0 according to the Vue specs. Solved this issue by copying the route object and changing the params. Resulting in not creating warnings for each file and component being checked by Webpack. 

Solution:

Create a copy of the route by substracting the params element. Params element has a pathMatch value. If not this will result in undefined and Vue will handle this the correct way. Otherwise baseRoute will result in a valid url and won't start a warning. Other checks will remain the same.

### DIFF
--- a/src/plugins/routing.js
+++ b/src/plugins/routing.js
@@ -43,9 +43,9 @@ function switchLocalePathFactory (i18nPath) {
     if (!name) {
       return ''
     }
-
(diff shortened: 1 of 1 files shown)


paulgv:
Thanks @dipsaus9 !

PR #194, opened 2019-01-18
Add canonical link to PREFIX_AND_DEFAULT duplicated pages

Nabellaleen:
When using `prefix_and_default` strategy, the default language pages have "duplicated routes". For example, if I have a route `A/B` with `en` as default language, I will get `/en/A/B` and `/A/B` routes.

For SEO, we needs, then, a `rel="canonical"` link.
This MR add on the `/en/A/B` route a canonical link to the `/A/B` route

More info: https://support.google.com/webmasters/answer/182192#dup-content

PS: I tried to add tests, but I've failed to setup tests on multiple nuxt.config.js files :/


paulgv:
Thanks you @Nabellaleen

PR #197, opened 2019-01-24
Fixes #179: Fix nested route locale prefix.

### LINKED ISSUE #179: Wrong routes for nested pages
### Version

### Reproduction link

[https://www.vysokyujezduprahy.cz/en](https://www.vysokyujezduprahy.cz/en)

### Steps to reproduce

All non-nested pages have their routes generated correctly - try visiting all pages except Pricemap.
Pricemap is a nested page and its routes do not include /en as they should.

### What is expected ?

URL: /en/pricemap

URL: /pricemap

### Additional comments?

```
module.exports = {
    seo: false,
    locales: [
        {
          code: 'en',
          iso: 'en-US',
          name: 'English',
          file: 'en.json'
        },
        {
          code: 'cs',
          iso: 'cs-CZ',
          name: 'Čeština',
          file: 'cs.json'
        },
        {
          code: 'ru',
          iso: 'ru-RU',
          name: 'Russian',
          file: 'ru.json'
        }
    ],
    lazy: true,
    langDir: 'locales/',
    defaultLocale: 'cs',
    parsePages: false,
    detectBrowserLanguage: false,
    pages: {
        index: {
            cs: '/',
            en: '/',
            ru: ''
        },
        'about': {
            cs: '/o-projektu',
            en: '/about-project',
            ru: ''
        },
        'properties': {
            cs: '/nemovitosti',
            en: '/properties',
            ru: ''
        },
        'location': {
            cs: '/lokalita',
            en: '/location',
            ru: ''
        },
        'pricemap/map': {
            cs: '/mapovy-cenik',
            en: '/pricemap',
            ru: ''
        },
        'pricemap/starred': {
            cs: '/mapovy-cenik/vybrane',
            en: '/pricemap/starred',
            ru: ''
        },
        'pricemap/starred-ids': {
            cs: '/mapovy-cenik/vybrane/:ids',
            en: '/pricemap/starred/:ids',
            ru: ''
        },
        'pricemap/detail': {
            cs: '/mapovy-cenik/detail/:id',
            en: '/pricemap/detail/:id',
            ru: ''
        },
        'pricemap/area': {
          cs: '/mapovy-cenik/lokalita/:id',
          en: '/pricemap/area/',
          ru: ''
        },
        'financing': {
            cs: '/koupe-financovani',
            en: '/financing',
            ru: ''
        },
        'contact': {
            cs: '/kontakt',
            en: '/contact',
            ru: ''
        },
        'u-kulen': {
            cs: '/u-kulen',
            en: '/u-kulen',
            ru: '/u-kulen'
        },
        'zamecke-zahrady': {
            cs: '/zamecke-zahrady',
            en: '/zamecke-zahrady',
            ru: '/zamecke-zahrady'
        },

    }
}
```

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c186">#c186</a>)</em></sub></div>

michaelize:
When parsePage is disabled, we extract paths from options. Those paths are treated as full URI, including those nested child pages. Therefore, we should treat all pages as top level pages when parsePage is disabled.


docs/routing.md
```diff
@@ -91,6 +91,22 @@ export default {
...
+If it's a nested route, paths that start with `/` will be treated as a root path:
+
+```js
+// pages/information/about.vue
+
+export default {
+  nuxtI18n: {
+    paths: {
+      en: '/about-us', // -> accessible at /about-us (no prefix since it's the default locale)
+      fr: 'a-propos', // -> accessible at /fr/information/a-propos
```
paulgv: I've been testing this change and the page is actually accessible at `/fra-propos`.
In this case, the expected config would be to specify the full path for each locale, ie:

```js
paths: {
  en: '/about-us',             // -> accessible at /about-us (no prefix since it's the default locale)
  fr: '/information/a-propos', // -> accessible at /fr/information/a-propos
  es: '/sobre'                 // -> accessible at /es/sobre
}
```

If we wanted to handle the case without the leading `/`, it would require more checks in the routes generation, which would lead to additional complexity in setting up a project with nuxt-i18n. I'm not sure we need to handle this for now.

src/helpers/routes.js
```diff
@@ -85,8 +85,10 @@ exports.makeRoutes = (baseRoutes, {
       const shouldAddPrefix = (
         // No prefix if app uses different locale domains
         !differentDomains &&
-        // Only add prefix on top level routes
-        !isChild &&
+        // Only add prefix when
+        // 1. on top level routes
+        // 2. parsePages is false, since we are not going from parent path to child path
+        !(isChild && parsePages) &&
```
paulgv: Do we really want to skip adding a prefix on children routes in this case? I'm not sure I understand this additional check.

michaelize: This is when it's top level or when we are not parsing pages, we add the prefix.

`!(isChild && parsePages)` is the same as `!child || !parsePages`.

docs/routing.md
michaelize: Sorry, I did the condition below incorrectly earlier. With the latest commit, we will add prefix when
1. it is top level path
2. the path starts with `/`

No complexity is added since vue-router already supports this path pattern. https://router.vuejs.org/guide/essentials/nested-routes.html here is says `Note that nested paths that start with / will be treated as a root path. This allows you to leverage the component nesting without having to use a nested URL.`

`path.startsWith('/')` is added to match with vue-router's design. This example is for nested routes only. Non-nested routes should always start with `/`.

src/helpers/routes.js
michaelize: This condition is incorrect, pls see the latest commit.

michaelize:
@paulgv With the latest commit, I've changed the condition, so it is now correctly generating the path at `/fr/a-propos`. Since the path starts with `/`, so it will be treated as a root path, and will not be prepended with `information/` after `/fr/`.

In `.nuxt/router.js` it has
```
{
      path: "/fr/a-propos",
      component: _535908ce,
      name: "about___fr"
}
```

PR #199, opened 2019-01-24
Fixes #191: Add encodeURI option to allow skipping encodeURI for custom paths

### LINKED ISSUE #191: Encoding custom paths breaks RegExps (introduced in #145)
### Version

### Reproduction link

[https://github.com/tronstoner/nuxt-i18n-bugreport](https://github.com/tronstoner/nuxt-i18n-bugreport)

### Steps to reproduce

1- git clone https://github.com/tronstoner/nuxt-i18n-bugreport.git
2- npm install
3- npm run dev
4- open console in browser
4- go to http://127.0.0.1:3000
5- Error: `[vue-router] missing param for named route "foo-id___de": Expected "id" to match "%5Cd+", but received "666"`

### What is expected ?

RegExps in custom paths in translations should be left untouched.

encodeURI() breaks custom paths. e.g:
```
"/foo/:slug-:id(\\d+)"
becomes:
"/en/foo/:slug-:id(%5Cd+)"
```
A change introduced in:
https://github.com/nuxt-community/nuxt-i18n/commit/98c99459c5b7db2521c333e65a8ddf88b6287229

### Additional comments?

I would be happy if for now encodeURI could be an just option (default: true), so we could simply take care of URI encoding ourselves before pushing paths to the router.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c191">#c191</a>)</em></sub></div>

michaelize:
When a custom path has regular expression inside, encodeURI will escape special characters. `/foo/:slug-:id(\\d+)` will becomes `/foo/:slug-:id(%5Cd+)`.

This PR adds `encodeURI` option per the suggestion in #191. If it's set to false, then custom paths will not be encoded through `encodeURI` method. Since no encoding will happen, the user will need to ensure the correctly encoded paths are specified.

### DIFF
--- a/docs/options-reference.md
+++ b/docs/options-reference.md
@@ -106,6 +106,12 @@ Here are all the options available when configuring the module and their default
   // the pages option, refer to the "Routing" section for usage
   pages: {},
 
+  // By default, custom paths will be encoded using encodeURI method.
+  // This does not work with regexp: "/foo/:slug-:id(\\d+)". If you want to use
+  // regexp in the path, then set this option to false, and make sure you process
+  // path encoding yourself.
+  encodeURI: true,
+
   // Called right before app's locale changes
   beforeLanguageSwitch: () => null,
 
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -175,31 +175,47 @@ You would need to set up your `pages` property as follows:
   pages: {
     about: {
       en: '/about',
-      fr: '/a-propos', 
+      fr: '/a-propos',
     },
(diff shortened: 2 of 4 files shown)


paulgv:
Awesome! Thanks a lot @michaelize !

paulgv:
Renaming the option to `encodePaths` to prevent conflicts with `encodeURI` method.

PR #210, opened 2019-02-03
Docs for YAML in i18n blocks

ljoonal:
I was trying to figure out how to use YAML in my <i18n> blocks, and noticed that there was no guide to do it. 
If there's some better way, please do let me know, since I want to use this myself :)


paulgv:
Thank @ljoonal !

PR #212, opened 2019-02-05
Update types

iliyaZelenko:
This code throws a error: 

> Property 'locales' does not exist on type 'VueI18n & IVueI18n'. Did you mean 'locale'?

```
get availableLocales () {
  return this.$i18n.locales.filter(i => i.code !== this.$i18n.locale)
}
```

It helps me.

Maybe this is not the best solution, but it should fix the error.

I think it's worth adding this for the rest of the variables:

https://github.com/nuxt-community/nuxt-i18n/blob/90bcd80e7f0cbf4f00d15353c4eadea31f6ff892/src/plugins/main.js#L53-L59

### DIFF
--- a/types/vue.d.ts
+++ b/types/vue.d.ts
@@ -1,6 +1,6 @@
 import Vue from "vue";
 import { RawLocation } from "vue-router";
-import VueI18n from "vue-i18n"
+import VueI18n from "vue-i18n";
 
 /**
  * The nuxt-i18n types namespace
--- a/types/vue.d.ts
+++ b/types/vue.d.ts
@@ -12,3 +12,9 @@ declare module "vue/types/vue" {
     getRouteBaseName(route: RawLocation): string;
   }
 }
+
+declare module 'vue-i18n' {
+  interface IVueI18n {
+    locales: any
+  }
+}
--- a/types/vue.d.ts
+++ b/types/vue.d.ts
@@ -9,9 +9,6 @@ declare namespace NuxtVueI18n {
   type Locale = VueI18n.Locale
 
   namespace Options {
-    // e.g.: ['en', 'fr', 'es']
-    type LocalesCodesArr = Locale[]
-
     // e.g.:
     // [
     //   { code: 'en', iso: 'en-US', file: 'en.js' },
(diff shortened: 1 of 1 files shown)


kevinmarrec:
@iliyaZelenko `$i18n` is directly related to `vue-i18n` and there is no `locales` object.

Assuming `nuxt-i18n` code seems to set `app.i18n = new VueI18n(...)` and then set things on `app.i18n` that is not defined in `VueI18n` itself, we might need to extend VueI18n interfaces to achieve what you're trying to do.

paulgv:
@iliyaZelenko do you see what's missing according to kevin's comment? Do you think you'd be able to complete the definitions before we merge?

iliyaZelenko:
@paulgv well, today I will make types for the remaining properties.

kevinmarrec:
@iliyaZelenko Cool, for the `locales` property, I think it can be better than `any` type, would need to check source code to know what's the correct type (It's probably an array of object having `code` and other properties defined in documentation)

iliyaZelenko:
@kevinmarrec It looks like I made types for all options, but there wasn’t much point, as these types will not be used in the code of this package and it is unlikely that someone will use them in their project.

The main thing is that I have extended the `IVueI18n` interface so that the `$i18n` object has additional properties, I have separated options that fall into `$i18n` in a separate interface `NuxtVueI18n.Options.VueI18nInterface`.

I tested these types on my small project, there were no errors.

kevinmarrec:
@iliyaZelenko Alright, great job !

I can't really take the time to check if types are all good, but it seems good and anyway it's not something that can break the project, it's "just" types.

So if it turns out that a type is not properly defined, we'll change it.

I trust you about having tested these types so let's stick with that !

paulgv:
Many thanks to both of you!
We now have some errors in the tests: [https://circleci.com/gh/nuxt-community/nuxt-i18n/355](https://circleci.com/gh/nuxt-community/nuxt-i18n/355)
I'm not sure what's missing, can you have a look @iliyaZelenko ?
I'll publish these changes as soon as the tests are fixed!

iliyaZelenko:
@paulgv this should fix the error: https://github.com/nuxt-community/nuxt-i18n/pull/216

PR #213, opened 2019-02-05
Update lang-switcher.md

iliyaZelenko:
The '$i18n.locales' variable inside 'v-for' directive should be replaced with a computed property that returns filtered array instead. You should not mix 'v-for' with 'v-if'. (vue/no-use-v-if-with-v-for)


paulgv:
Thanks!

PR #222, opened 2019-02-15
Fixes #221 store unsync when changing language in lazy 

### LINKED ISSUE #221: Vuex store current locale is out of sync when swithing between locales if lazy loading is enabled
### Version

### Reproduction link

[https://github.com/nuxt-community/nuxt-i18n](https://github.com/nuxt-community/nuxt-i18n)

### Steps to reproduce

Have lazy loading enabled and then switch between locales on frontend using the switchLocalePath helper.

### What is expected ?

The vuex store locale value should update

It uses the previous locale value instead which means it lags by one locale as you switch between multiple locales

### Additional comments?

I have tracked in down to the generated code in .nuxt/nuxt-i18n/middleware.js in the if statement checking if lazy loading is enabled the generated code has syncVuex(locale, messages) but it should be syncVuex(newLocale, messages)

Ace plugin btw :)

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c205">#c205</a>)</em></sub></div>

xibman:
Fixes vues store unsync state when lazy i18n loading and you try to change language

### DIFF
--- a/src/templates/middleware.js
+++ b/src/templates/middleware.js
@@ -85,7 +85,7 @@ middleware['i18n'] = async ({ app, req, res, route, store, redirect, isHMR }) =>
       const messages = await loadLanguageAsync(app.i18n, newLocale)
       app.i18n.locale = newLocale
       app.i18n.onLanguageSwitched(oldLocale, newLocale)
-      syncVuex(locale, messages)
+      syncVuex(newLocale, messages)
     } else {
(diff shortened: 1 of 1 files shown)


paulgv:
Thanks @xibman !

PR #230, opened 2019-02-21
Fix duplicated route names with prefix_and_default strategy (fixes #140)

### LINKED ISSUE #140: Duplicate route names with prefix_and_default strategy
### Version

### Reproduction link

[https://codesandbox.io/s/ql7wx2mvk6](https://codesandbox.io/s/ql7wx2mvk6)

### Steps to reproduce

1. Configure nuxt-i18n module with these settings:
```js
    [
      "nuxt-i18n",
      {
        locales: [
          { code: "pt", iso: "pt-PT", file: "pt.json" },
          { code: "us", iso: "en-US", file: "en.json" }
        ],
        langDir: "locales/",
        lazy: true,
        strategy: "prefix_and_default",
        defaultLocale: "us"
      }
    ]
```

2. Load the root path.
3. Look at the console (the server one, at the bottom of the codesandbox UI)

### What is expected ?

No warnings about duplicated routes.

Getting console warnings:

```
[vue-router] Duplicate named routes definition: { name: "about___us", path: "/us/about" } 
[vue-router] Duplicate named routes definition: { name: "index___us", path: "/us" }
```

### Additional comments?

Appears to be due to using `strategy: "prefix_and_default",` setting which creates routes with duplicate names for default locale.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c153">#c153</a>)</em></sub></div>

rchl:
Fixed by adding a `___default` suffix (customizable) to prefix-less (default locale),
extra routes generated when using `prefix_and_default` strategy. So for example,
with 'en' being default locale, two generated index 'en' routes will have names:
 - index___en
 - index___en___default

Additional changes done to handle extra suffix properly, without breaking existing
behavior. That is, when being on a route without prefix, `localePath` and
`switchLocalePath` will still return paths without prefix.

Sort of unrelated to this change but made `routesNameSeparator` not be exposed
on `app.i18n` as it's neither documented to be exposed there nor it shouldn't be
needed for anything.

### DIFF
--- a/docs/options-reference.md
+++ b/docs/options-reference.md
@@ -28,6 +28,10 @@ Here are all the options available when configuring the module and their default
   // need to change this
   routesNameSeparator: '___',
 
+  // Suffix added to generated routes name for default locale if strategy is prefix_and_default,
+  // you shouldn't need to change this
+  defaultLocaleRouteNameSuffix: 'default',
+
   // Routes generation strategy, can be set to one of the following:
   // - 'prefix_except_default': add locale prefix for every locale except default
   // - 'prefix': add locale prefix for every locale
--- a/src/helpers/constants.js
+++ b/src/helpers/constants.js
@@ -27,6 +27,7 @@ exports.DEFAULT_OPTIONS = {
   locales: [],
   defaultLocale: null,
   routesNameSeparator: '___',
+  defaultLocaleRouteNameSuffix: 'default',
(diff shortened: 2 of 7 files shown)


paulgv:
Very nice! Thanks @rchl !

PR #231, opened 2019-02-22
Update browser-language-detection.md

Igloczek:
Missing wrapper around config options added.


paulgv:
Good catch, thanks @Igloczek !

Issue #80, opened 2018-05-02
Cannot read property 'iso' of undefined


maxmckenzie:
^ I am not handing this site a token for my github account.

i'm getting the error 

`Cannot read property 'iso' of undefined`

i'd like to know which iso you are using there is `es` v1 of the iso and `es-ES`

soooooo which is it.

maxmckenzie:
ah i see your bot closes the ticket if i dont open it somewhere else.

maxmckenzie:
BOOOOOOOOOOOOO!!!!! this sucks. why remove github functionality and make me use this cmty thing.

BOOOOOOOOOOOOOOOO

maxmckenzie:
even when you use the correct ISO code there is still the error.

paulgv:
Hi @maxmckenzie 
What does your config look lilke?

maxmckenzie:
Hey @paulgv thanks so much for the reply, unfortunately, I'm working on a time-sensitive project. So I've now gone with just using vue and rolling my own boilerplate. As such i've removed the config and i didnt commit a version of it so its not in my git history.

If i get some time spare i shall try to replicate the issue.

mahnuh:
@paulgv here you can find a project with this error: https://github.com/mahnuh/nuxt-test

paulgv:
Thanks @mahnuh !
There were some checks missing in the SEO plugin, the issue should be fixed in v3.2.2 :smiley:

mahnuh:
Thanks @paulgv , you just made my day - again! 😃

shrpne:
@paulgv 
I have the same issue, but only on 404 page, 
Nuxt-i18n version is 4.0.0, my config:
```
    modules: [
        ['nuxt-i18n', {
            locales: [
                {
                    code: 'ru',
                    iso: 'ru',
                },
                {
                    code: 'en',
                    iso: 'en',
                },
            ],
            strategy: 'prefix',
            rootRedirect: 'ru/1',
            vueI18n: {
                fallbackLocale: 'ru',
                messages: {
                    ru: {
                        welcome: 'Welcome'
                    },
                    en: {
                        welcome: 'Bienvenue'
                    },
                }
            },
        }]
    ]
```

begueradj:
I am facing the same issue with the current Nuxt version (2.1.0)

acidjazz:
[...]

PR #233, opened 2019-02-26
Fix 'cannot read property "iso"' crash (issue #80)

rchl:
On visiting route that is not defined (404), and module's `defaultLocale`
is not defined (default), the code in main.js would set `app.i18n.locale`
to null and when the seo code would run, it would crash on trying to
access property `iso` of it. So don't allow seo code to run when locale
is not defined.


paulgv:
Thanks @rchl !

Issue #34, opened 2018-02-25
Load fallback language when needed (lazy-load)


manniL:
nuxt-i18n version: 2.3.4

Imagine you have two languages, `en` (default) and `de`.
But only parts of the `en` language is localized in the `de` file.
If you now enter a page (e.g. `/de/only-half-localized-route`), you see the correct translations (for de) where it applies, but no en translation because the fallback language isn't loaded properly because of lazy-loading. Instead, you see the translation keys.

A naive fix for it would be to load the fallback language every time, but I guess there is a way to check if a fallback is needed and load the fallback language in that case (if not already loaded).

<div align="right"><sub><em>This feature request is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt-community/nuxt-i18n/issues/c39">#c39</a>)</em></sub></div>

maximivanov:
Any update on this? Or before feature is implemented, what's the way to load fallback language every time? (and not load all of language files at the same time)

scippio:
Any news?

RecuencoJones:
As a workaround for now we might be using the following:

```javascript
// nuxt.config.js
module.exports = {
  srcDir: './src',
  ...
  modules: [
    '@nuxtjs/pwa',
    [
      'nuxt-i18n',
      {
        locales: [
          { code: 'en', iso: 'en-US', file: 'en.js' },
          { code: 'es', iso: 'es-ES', file: 'es.js' }
        ],
        defaultLocale: 'en',
        vueI18n: {
          fallbackLocale: 'en'
        },
        lazy: true,
        langDir: 'locales/'
      }
    ]
  ],
  ...
};

// src/locales/es.js
import { assignWith } from 'lodash';
import { loadTranslation } from '../services/contents';

const useDefault = (value, defaultValue) =>
  typeof value === 'object' ?
    assignWith({}, value, defaultValue, useDefault) :
    value || defaultValue;

export default () => Promise.all([
  loadTranslation('en'),
  loadTranslation('es')
])
.then(([ defaultMessages, localizedMessages ]) => 
  assignWith({}, localizedMessages, defaultMessages, useDefault));
```

JkmAS:
@RecuencoJones how does the `loadTranslation` function look like? Are you using axios or nuxt-axios in `loadTranslation`? Thx.

rchl:
[...]

PR #260, opened 2019-03-26
Fix load fallback locale when lazy-load

orblazer:
Hello,

I have fix the fallback locale on lazy-load.
Before this PR when using local fallback in lazy-load it had no effect, the language was not loading.
This is fix the issue #34.

Thanks.


src/plugins/main.js
```diff
@@ -88,6 +88,12 @@ export default async ({ app, route, store, req }) => {
   // Lazy-load translations
   if (lazy) {
     const { loadLanguageAsync } = require('./utils')
+
+    // Load fallback lang
+    if(app.i18n.fallbackLocale) {
+      loadLanguageAsync(app.i18n, app.i18n.fallbackLocale)
```
rchl: - Need to await
- No need to call it if fallback is the same as current
- I think it's not the best place to do it as it will trigger for every request. It should be done at init stage instead.

orblazer: I have fix that but i don't know where i can move that. I think in `index.js` but this need change `loadLanguageAsync`

rchl: I think it actually has to be done on each request so disregard my last point.

This fix looks good to me (apart from missing space after `if` ;)).

orblazer: Its done

tvld:
Strange... the bug still exists today and fix was not accepted? Or am I missing something?

rchl:
It was integrated as #291 so your issue might be something else.

tvld:
No, it's still not fixed properly as far as I can tell. For me, it only works if I configure in ``nuxt.config.js`` like so:
```js
   vueI18n: {
          fallbackLocale: 'en'
        },
        detectBrowserLanguage: {
          alwaysRedirect: true,
          fallbackLocale: 'en'
        },
```
The trick is to configure **both** ``vue-i18n.fallbackLocale`` and  ``detectBrowserLanguage. fallbackLocale: 'en'``.  Which does not make sense for a new-bie, reading all the docs ;-)

rchl:
It's not the best place to discuss this problem so please create a new issue/question if you want to discuss but:
1) Those `fallbackLocale` options are sort of unrelated.
 - `vueI18n.fallbackLocale` tells VueI18n to use strings from English locale if currently loaded locale doesn't have a given string ID.
 - `detectBrowserLanguage.fallbackLocale` decided the locale to use if the current browser locale is not supported.
2) The fix was related to `vueI18n.fallbackLocale` setting. It didn't work without the fix because the fallback locale was not loaded.

tvld:
Thank you for your thorough answer )):
1 - it is clear now, but was not after reading the docs., Thanks for clear explanation
2/3 - "it" was: fallback language was not loaded
About the issue reporting: I tried three times (two times other issue than this one), but Nuxt made it so complicated to report issues, that I failed and gave up...  Maybe it works for big bugs but for small or trivial issues I recommend everybody to not report as it takes a lot of time...

rchl:
Sorry but if it's not worth the time the report it then it's probably not worth fixing either.

I claim that this feature works since I use it in my project and also tests back it up, so it must be either a case special to your code/strings or wrong use.

In that case, I wouldn't be able to fix it without having a reproducible case which is where you would have to make the effort if you want it fixed. And possibly you would even find the issue yourself when trying to create minimal testcase.

tvld:
Well.. especially the little things that _new users_, _not_ beginners in vue, stumble upon.. are they worth fixing or not? It is because a new users came from different angle that made sense and gives new insights? Or just an own mistake...oversight? It's really not trivial to decide. And then being forced to set up code pens and so on...well.. feels like over-killing the issue pipeline. And I would not like to waste time of the guru's )) 

Al in all, I concluded first project with Nuxt now... and love it, grateful to the quality, thoughts and process. Will be happy to stick around Nuxt and nuxt-vuei18n for sure ))  so thank you ! )))

PR #266, opened 2019-04-08
Fix localePath call on initial redirect if dynamic route

rchl:
Code has failed to redirect to route with locale prefix if that route
happened to be dynamic (_.vue).

As an example:

 - User goes to /foo handled by _.vue.
 - Nuxt-i18n (with `detectBrowserLanguage` enabled) tries to redirect
   to /fr/foo
 - VueRouter.resolve fails to match route.

 Problem is with VueRoute creating route with `params: {pathMatch: foo}`
 that nuxt-i18n is passing to `localePath()` which uses
 `VueRouter.resolve` to resolve that route with locale. Problem is,
 `VueRouter.resolve` API expects the param for dynamic route be in a '0'
 property and not 'pathMatch' (This is an inconsistency in VueRouter but
 that's not the point). To handle that case properly, we need to remap
 'pathMatch' to '0', if 'pathMatch' is provided.

 As a bonus, it also allows one to use localePath with pathMatch argument
 instead of 0 which is nicer and matches VueRouter documentation better.
 For example: we can do `localePath({name: 'all', params: {pathMatch: 'foo'}})`

 Notice that `switchLocalePath` function has equivalent fix applied already.


paulgv:
@rchl Sorry for delay, I missed this PR! Thanks for working on this!

PR #272, opened 2019-04-16
just pulls out the file COMPONENT_OPTIONS_KEY

darcoder:
https://nuxt-community.github.io/nuxt-i18n/routing.html#custom-paths


paulgv:
Hi @darcoder ! Thanks for the PR!
I like the fact that we could remove `acorn` dependency by implementing our own parser.
The only issue I'm seeing for now is that this change is not covering the case where a boolean is passed as `nuxtI18n` value instead of an object, eg: https://nuxt-community.github.io/nuxt-i18n/routing.html#ignore-routes
Any chance you could update your PR to take this case into account?

Thanks!

PR #274, opened 2019-04-18
remove acorn

darcoder:
only one COMPONENT_OPTIONS_KEY in file

### DIFF
--- a/src/helpers/components.js
+++ b/src/helpers/components.js
@@ -2,8 +2,8 @@ const { readFileSync } = require('fs')
 const { COMPONENT_OPTIONS_KEY } = require('./constants')
 
 const acorn = require('acorn')
-const walker = require('acorn-walk')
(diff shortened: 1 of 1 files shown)


paulgv:
Hi @darcoder ! Thanks again!
I have the same comment than on your initial PR: https://github.com/nuxt-community/nuxt-i18n/pull/272#issuecomment-483802345

> The only issue I'm seeing for now is that this change is not covering the case where a boolean is passed as `nuxtI18n` value instead of an object, eg: https://nuxt-community.github.io/nuxt-i18n/routing.html#ignore-routes

I'd super happy to merge this once we've made sure we won't break this option's behavior!

darcoder:
You can connect the two options through a variable in the configuration?

(options.acorn == true){
acorn variant
}else{
this variant
}

Issue #259, opened 2019-03-26
Route is break when nuxt.config is updated


orblazer:
### Version

[v5.8.5](https://github.com/nuxt-i18n/releases/tag/v5.8.5)

### Reproduction link

Not really needed

### Steps to reproduce

1. Install `nuxt-i18n`
2. Create routes with like config :
```js
[
  'nuxt-i18n',
  {
  locales: [
    { code: 'en', name: 'English', iso: 'en-US', file: 'en.js' },
    { code: 'fr', name: 'Français', iso: 'fr-FR', file: 'fr.js' }
  ],
  defaultLocale: 'en',
  lazy: true,
  langDir: 'locale/',
  vueI18n: {
  fallbackLocale: 'en',
  dateTimeFormats: {
    fr: {
      long: {
        year: 'numeric',
        month: 'long',
        day: 'numeric',
        weekday: 'long'
      },
      timeShort: {
        year: 'numeric',
        month: 'numeric',
        day: 'numeric',
        hour: 'numeric',
        minute: 'numeric'
      },
      timeLong: {
        year: 'numeric',
        month: 'long',
        day: 'numeric',
        weekday: 'long',
        hour: 'numeric',
        minute: 'numeric'
      }
    },
    en: {
      long: {
        year: 'numeric',
        month: 'long',
        day: 'numeric',
        weekday: 'long'
      },
      timeShort: {
        year: 'numeric',
        month: 'numeric',
        day: 'numeric',
        hour: 'numeric',
        minute: 'numeric'
      },
      timeLong: {
        year: 'numeric',
        month: 'long',
        day: 'numeric',
        weekday: 'long',
        hour: 'numeric',
        minute: 'numeric'
      }
    }
  },
  detectBrowserLanguage: {
    useCookie: true,
    alwaysRedirect: true,
    fallbackLocale: 'en'
  },
  onLanguageSwitched: (previous, current) => {
    if (process.client) {
      cookie.serialize('i18n_redirected', current, {
        expires: new Date(new Date() + 365),
        path: '/'
      })
    }
  }
]
```
3. Update `nuxt.config.js`

### What is expected ?

Just update

### What is actually happening?

When `nuxt.config.js` is updated we have
```
 WARN  [vue-router] Route with name 'index___en' does not exist                                                               11:26:56
 WARN  [vue-router] Route with name 'index___fr' does not exist                                                               11:26:56
```

[...]

PR #276, opened 2019-04-19
Fix load fallback locale when lazy-load and invalid route on update

orblazer:
Hello,
I have merge my 2 PR #260 and #261.

This is fix load fallback locale when lazy-load, before this PR when using local fallback in lazy-load it had no effect, the language was not loading.
This is fix invalid route when update nuxt.config.

This is fix issues #34 and #259.

Thanks.


paulgv:
I'd need more eyes on this one.
According to [this comment](https://github.com/nuxt-community/nuxt-i18n/issues/259#issuecomment-477045528), this change might actually break the `prefix_except_default` strategy.
It would be best to fix one problem per PR. Also, you've made some formatting changes which makes the PR even harder to review because it makes the relevant changes less visible, I'd recommend you do this kind of changes in a different PR next time.

Thanks for working on this @orblazer !

orblazer:
@rchl you can puck it up because i am currently on project without nuxt and i don't have time before 2 months

rchl:
@orblazer @paulgv created #291 with this fix extracted

paulgv:
Closing in favor of #291

PR #291, opened 2019-05-06
Load fallback locale when needed if lazy-loading (fixes #34)

### LINKED ISSUE #34: Load fallback language when needed (lazy-load)

rchl:
Fallback locale is needed to properly handle strings that exist in
base/fallback locale but not in currently used one.

Fix from @orblazer

### DIFF
--- a/docs/lazy-load-translations.md
+++ b/docs/lazy-load-translations.md
@@ -7,8 +7,8 @@ To enable translations lazy-loading, follow these 4 steps when configuring **nux
 * Set `lazy` option to `true`
 * Set `langDir` option to the directory that contains your translation files (this can NOT be empty)
 * Configure `locales` option as an array of object, where each object has a `file` key which value is the translation file corresponding to the locale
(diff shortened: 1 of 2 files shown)


paulgv:
Thanks for picking this up @rchl! Thanks for submitting the initial fix @orblazer!

PR #299, opened 2019-05-09
improvement: wait for sync store to finish

rchl:
This doesn't fix any specific bug but I've noticed that code doesn't
wait for sync dispatch actions which are async so I thought that fixing
that is the right thing to do.


paulgv:
Thanks!

PR #319, opened 2019-05-30
fix: add types for nuxtI18nSeo

aldarund:


types/vue.d.ts
```diff
@@ -80,10 +80,28 @@ declare namespace NuxtVueI18n {
  * Extends types in vue
  */
 declare module "vue/types/vue" {
+  interface NuxtI18nSeo {
+    htmlAttrs: {
```
rchl: Should htmlAttrs, link and meta be marked as optional? Logic of seo-head.js doesn't guarantee them to be set.

types/vue.d.ts
```diff
@@ -80,10 +80,28 @@ declare namespace NuxtVueI18n {
...
+      name: string,
+      property: string,
+      content: string
+    }]
+  }
   interface Vue {
     localePath(route: RawLocation, locale?: string): string;
     switchLocalePath(locale: string): string;
     getRouteBaseName(route: RawLocation): string;
+    $nuxtI18nSeo: () => NuxtI18nSeo
```
rchl: Is this syntax of defining a function equivalent to `$nuxtI18nSeo(): NuxtI18nSeo`? If so then maybe for consistency it could be using other syntax?

rchl: Also missing semicolon if we are talking consistency :)

types/vue.d.ts
aldarund: @rchl i messed up syntax and declared them wrong as a array of 1 element :) Changed.
htmlAttrs, link and meta will always be returned as i see from a code. But they could be empty arrays. And lang in htmlAttrs could be optional. I changed that.

types/vue.d.ts
aldarund: yes, although at previous declaration from this file its no semicolon and that syntax, its already not consistent :) but changed here to be consistent within this interface

types/vue.d.ts
aldarund: actually i missed first return, so yes they should be optional

types/vue.d.ts
```diff
@@ -80,10 +80,28 @@ declare namespace NuxtVueI18n {
  * Extends types in vue
  */
 declare module "vue/types/vue" {
+  interface NuxtI18nSeo {
```
rchl: I don't really know much about typescript or writing declaration files but should this interface be defined outside of the `declare module` block?
@kevinmarrec

paulgv: Let's merge this as is for now.
@kevinmarrec If you get the chance to have a look at this, could you answer @rchl's question?

paulgv:
Thanks @aldarund!

types/vue.d.ts
kevinmarrec: @paulgv Sorry I was away from keyboard.

@rchl Is right, the interface definition should be before declaring module and not inside.

Issue #123, opened 2018-08-31
Server side rendered locale switching links will always start with http


Ingramz:
### Version

[v5.2.1](https://github.com/nuxt-i18n/releases/tag/v5.2.1)

### Reproduction link

[-](-)

### Steps to reproduce

-

### What is expected ?

-

### What is actually happening?

-

### Additional comments?

In https://github.com/nuxt-community/nuxt-i18n/blob/e2aa687083873b6cb20b29beb044db8572e223cf/src/plugins/routing.js#L56 `req.secure` is used, which is only defined in Express, but SSR context only provides Node's `http.IncomingMessage` which does not contain such property, resulting in any of the links generated by `switchLocalePath` server side to start with `http`.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c128">#c128</a>)</em></sub></div>

Ingramz:
It does not seem a very easy fix as the method that determines `req.secure` is quite complex. Porting all of that logic over to this plugin is not very intuitive. I would certainly contribute, if I knew what the suitable solution was.

manniL:
Would it be possible to check the requested URL through `req`? :thinking:

Ingramz:
Not if the hosted application is behind a reverse proxy.

rchl:
The logic is not that big (https://github.com/expressjs/express/blob/b8e50568af9c73ef1ade434e92c60d389868361d/lib/request.js#L306) but:
 * it requires external `proxy-addr` dependency
 * while in most cases it should do the right thing on its own, in specific setups users would need to provide configuration for that logic (same as `trust proxy` in expressjs provides here https://expressjs.com/en/4x/api.html#trust.proxy.options.table)
It feels kinda wrong to have to have all that in nuxt-i18n itself. It would be better if nuxt itself took care of those things. I wished nuxt used expressjs instead of connect...

Ingramz:
> It feels kinda wrong to have to have all that in nuxt-i18n itself. It would be better if nuxt itself took care of those things.

That's what I am thinking as well.

pi0:
Checking protocol can be easily done for this module:

https://github.com/nuxt-community/is-https

Which is already used by sitemap module

Ingramz:
@pi0 thanks. This is exactly what is needed. Including it in this package should be a no-brainer too...

PR #325, opened 2019-06-18
fix: better server side protocol detection (fixes #123)

rchl:
Use module that detects protocol used in more accurate way.


paulgv:
Thanks @rchl !

PR #327, opened 2019-06-21
fix: do not discard already present set-cookie header

rclement:
This PR aims at fixing the current behavior of `nuxt-i18n` when setting the detected language cookie.
If someone were to set a cookie within a Nuxt server-middleware (e.g. a CSRF token cookie), it will be discarded only the first time `nuxt-i18n` is setting the language redirection cookie.

The root cause comes from using `res.setHeader('Set-Cookie', redirectCookie)`: it no longer implicitly append the cookie value to the current list of `Set-Cookie` values in Express > 4, as explained here: https://github.com/expressjs/express/wiki/Migrating-from-3.x-to-4.x#ressetheaderset-cookie-val

To keep compatibility of this middleware with requests/responses from Express/Connect/etc., using the `cookies` package instead of `cookie` seems like a sensible choice to me, and fixes nicely the issue.

If I missed something while refactoring the code, please let me know so we can find the most elegant and robust solution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "acorn": "^6.1.1",
     "acorn-dynamic-import": "^4.0.0",
     "acorn-walk": "^6.1.1",
-    "cookie": "^0.4.0",
+    "cookies": "^0.7.3",
     "is-https": "^1.0.0",
     "js-cookie": "^2.2.0",
     "vue-i18n": "^8.11.2",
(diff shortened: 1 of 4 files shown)


paulgv:
LGTM! Thank you @rclement !

rchl:
Another way to fix that would be to get existing cookies and modify the array. Maybe that' better option than introducing this heavier dependency

PR #334, opened 2019-06-22
Only require is-https dependency on the server

rchl:
Makes that dependency be tree-shaken and not included at all on the
client. As a side effect, fixes IE11 that had problem with JS default
arguments used in is-https package.


paulgv:
Thanks @rchl !

PR #335, opened 2019-06-22
fix: Move SEO types out of Vue module declaration

paulgv:
This follows up on the discussion here: https://github.com/nuxt-community/nuxt-i18n/pull/319#discussion_r289303554

> rchl 22 days ago  Member
> I don't really know much about typescript or writing declaration files but should this interface be defined outside of the declare module block?
> 
> kevinmarrec 20 days ago  Member
> @rchl Is right, the interface definition should be before declaring module and not inside.

### DIFF
--- a/types/vue.d.ts
+++ b/types/vue.d.ts
@@ -76,27 +76,28 @@ declare namespace NuxtVueI18n {
   }
 }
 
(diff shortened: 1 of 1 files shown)


paulgv:
@rchl Is this the right way to fix this?

PR #345, opened 2019-07-01
feat: Dynamic route parameters translation

### LINKED ISSUE #79: Possibility to localize dynamic routes

*I'm not sure if it is possible :)*
I have a route `/article/:slug` (en) and I'd like to translate it to other languages with translated `slug`
Eg: (en) `/article/first-article` and (de) `/de/artikel/erste-artikel`

Now I can do this:
`article.vue`
```
i18n: {
  paths: {
    de: '/artikel/:slug'
  }
},
data () {
  return {
    article: { // it comes from asyncData
      // title: { en: '...', de: '...' },
      // slug: { en: '...', de: '...' }
    }
  }
}
```
but I can't change the `slug` based on the article's data

I think the `i18n.paths[locale]` should accept a function with the access of `this` (data) and should return some Object:
```
de () {
  return {
    path: '/artikel/:slug',
    params: {
      slug: this.article.slug.de
    }
  }
}
```

<div align="right"><sub><em>This feature request is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt-community/nuxt-i18n/issues/c74">#c74</a>)</em></sub></div>

paulgv:
# What's new?

Adds support for translating dynamic route parameters via the Vuex store module

BREAKING CHANGE: `preserveState` is now set automatically when registering the store module and
cannot be set via the configuration anymore

# So why do we need this?

When using `switchLocalePath` to create a global lang switcher component, there's currently no way for the method to know how to translate dynamic parameters. This is an attempt to address this issue.

# How to test this?

This has been published as a prerelease on NPM, to test this version, install `v6.0.0-0` with Yarn:

```sh
yarn add nuxt-i18n@6.0.0-0
```

Or NPM:

```sh
npm i nuxt-i18n@6.0.0-0
```

Refer to the new documentation section to see how to use this in your app: [Dynamic route parameters](https://github.com/nuxt-community/nuxt-i18n/blob/9f92005da111c4536191f4749db4fc289cd1ae2e/docs/lang-switcher.md#dynamic-route-parameters)

close #79

### DIFF
--- a/docs/lang-switcher.md
+++ b/docs/lang-switcher.md
@@ -41,3 +41,36 @@ computed: {
 ```
 
 If `detectBrowserLanguage.useCookie` and `detectBrowserLanguage.alwaysRedirect` options are enabled, you might want to persist change to locale by calling `this.$i18n.setLocaleCookie(locale)` (or `app.i18n.setLocaleCookie(locale)`) method. Otherwise locale will switch back to saved one during navigation.
+
+## Dynamic route parameters
+
+Dealing with dynamic route parameters requires a bit more work because you need to provide parameters translations to **nuxt-i18n**. For this purpose, **nuxt-i18n**'s store module exposes a `routeParams` state property that will be merged with route params when generating lang switch routes with `switchLocalePath()`.
+
+> NOTE: Make sure that Vuex [is enabled](https://nuxtjs.org/guide/vuex-store) in your app and that you did not set `vuex` option to `false` in **nuxt-i18n**'s options.
+
+To provide dynamic parameters translations, dispatch the `i18n/setRouteParams` as early as possible when loading a page, eg:
+
+```vue
+<template>
+  <!-- pages/_slug.vue -->
+</template>
+
+<script>
+export default {
+  async asyncData ({ app, store }) {
+    const params = await new Promise(resolve => {
+      resolve({
+        en: { slug: 'my-post' },
+        fr: { slug: 'mon-article' },
+      })
+    })
+    store.dispatch('i18n/setRouteParams', params)
+    return {
+      // your data
+    }
+  }
+}
+</script>
+```
+
+> NOTE: **nuxt-i18n** won't reset parameters translations for you, this means that if you use identical parameters for different routes, navigating between those routes might result in conflicting parameters. Make sure you always set params translations in such cases.
--- a/docs/lang-switcher.md
+++ b/docs/lang-switcher.md
@@ -64,7 +64,7 @@ export default {
         fr: { slug: 'mon-article' },
       })
     })
-    store.dispatch('i18n/setRouteParams', params)
+    await store.dispatch('i18n/setRouteParams', params)
     return {
       // your data
     }
--- a/docs/lang-switcher.md
+++ b/docs/lang-switcher.md
@@ -57,14 +57,11 @@ To provide dynamic parameters translations, dispatch the `i18n/setRouteParams` a
 
 <script>
 export default {
-  async asyncData ({ app, store }) {
-    const params = await new Promise(resolve => {
-      resolve({
-        en: { slug: 'my-post' },
-        fr: { slug: 'mon-article' },
-      })
+  async asyncData ({ store }) {
+    await store.dispatch('i18n/setRouteParams', {
+      en: { slug: 'my-post' },
+      fr: { slug: 'mon-article' }
     })
-    await store.dispatch('i18n/setRouteParams', params)
     return {
       // your data
     }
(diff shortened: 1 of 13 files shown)


paulgv:
@rchl If you get the chance to have a look at this, please let me know what you think.

docs/lang-switcher.md
```diff
@@ -41,3 +41,36 @@ computed: {
...
+<script>
+export default {
+  async asyncData ({ app, store }) {
+    const params = await new Promise(resolve => {
+      resolve({
+        en: { slug: 'my-post' },
+        fr: { slug: 'mon-article' },
+      })
+    })
+    store.dispatch('i18n/setRouteParams', params)
```
rchl: Should also await dispatch as it's async.

test/fixtures/basic/pages/posts/_slug.vue
```diff
@@ -13,6 +13,16 @@ import LangSwitcher from '../../components/LangSwitcher'
...
     LangSwitcher
+  },
+  async asyncData ({ app, store }) {
+    const params = await new Promise(resolve => {
+      resolve({
+        en: { slug: 'my-post' },
+        fr: { slug: 'mon-article' },
+      })
+    })
+    store.dispatch('i18n/setRouteParams', params)
```
rchl: should await dispatch

rchl:
Wouldn't it be simpler if switchLocalePath accepted an optional object with that information that is currently passed through store?

rchl:
Alternatively, I wonder if completely declarative approach is possible, extract that information from pages at built time and maybe use Route object to store that information (`Route.meta` maybe)...

paulgv:
Thanks for the feedback @rchl 

> Wouldn't it be simpler if switchLocalePath accepted an optional object

I think that users typically don't want to call `switchLocalePath` more than once in the app. It's pretty easy to create a `LangSwitcher` component and render it in the layout without having to think about special cases like the one we're trying to address here. I feel like pages need to be able provide parameters translations seamlessly and let `switchLocalePath` deal with these parameters "backstage".

> extract that information from pages at built time

Since we're dealing with dynamic paths here, we can't really get the required data at build time because it might result in huge payloads in router meta if we were to handle all possible parameters & translations for all dynamic routes. Additionally, since the data will most likely grow between each build, the router will frequently be out-of-sync with the database.

> I would prefer that this feature doesn't require one to enable store

Agreed. I ended up going for this solution even though I'm not a fan of it because I can't seem to make anything else work properly. Ideally, I'd like **nuxt-i18n** to expose a custom method trough pages components `$options`, that would return translated route parameters and the module would handle the rest. But we still need to make that data accessible *globally* for `switchLocalePath` to have access to it wherever it's used, while still being consistent on both the server and the client-side, which always a feels a bit tricky to me. I tried another solution where I injected a global `Vue.observable` that would serve as single purpose store but its state was lost on the client-side and I couldn't figure out why.

Any thoughts?

paulgv:
> Wouldn't it be simpler if switchLocalePath accepted an optional object

To clarify my thoughts here. I think that your suggestion would absolutely be valid and we probably should implement it anyway. But if I'm not mistaken, that solution isn't compatible with using a "global" lang switcher in the layout, it would require users to render the lang switcher in every page, specifying translated parameters where needed, which could get a bit tedious in large applications.

rchl:
Good point about having language switcher in the layout. Then that wouldn't work if data would be returned from page's `asyncData`. Didn't think of that.

In that case I guess we should go for current solution. I wonder though, given that it will be a major version, should we disable `setLocale` and `setMessages` mutations by default? I feel like it's something that not many people need but it can add a lot of heft and users might not even realize that.

src/plugins/main.js
```diff
@@ -30,14 +30,18 @@ export default async (context) => {
...
       }),
       actions: {
         setLocale ({ commit }, locale) {
           commit(vuex.mutations.setLocale, locale)
         },
         setMessages ({ commit }, messages) {
           commit(vuex.mutations.setMessages, messages)
+        },
+        setRouteParams ({ commit }, params) {
+          commit(vuex.mutations.setRouteParams, params)
```
rchl: I guess you have access to all locales here so it could be nice to validate keys of the object here according to available locale. That's because I feel people might commit this object with wrong structure and then will wonder why things are not working. Just a suggestion...

test/fixtures/basic/pages/posts/index.vue
```diff
@@ -17,6 +17,17 @@ import LangSwitcher from '../../components/LangSwitcher'
 export default {
   components: {
     LangSwitcher
+  },
+  async asyncData () {
+    const params = await new Promise(resolve => {
```
rchl: I'm not really sure if there is a point in returning static object through an extra promise. I guess you might be trying to simulate more real-world case but I don't think it can really change anything in practice as `asyncData` is an async function so it will run asynchronously anyway.
This applies to all modified tests.

paulgv:
Thanks for the great suggestions @rchl, I'll address them ASAP.

PR #354, opened 2019-07-10
Use default locale's custom path if not defined for a locale

jeanphilippeds:
Hello,

Here is a PR to use default locale's custom path when a custom path is not defined for a locale.
It can help if you have a lot of locales to manage and only want a custom path for a few of them.

### DIFF
--- a/src/helpers/utils.js
+++ b/src/helpers/utils.js
@@ -53,7 +53,7 @@ exports.getPageOptions = (route, pages, locales, pagesDir, defaultLocale) => {
   options.locales = options.locales.filter(locale => pageOptions[locale] !== false)
 
   // Construct paths object
-  options.paths = options.locales
+  options.locales
     .forEach(locale => {
       if (typeof pageOptions[locale] === 'string') {
         // Set custom path if any
--- a/src/helpers/utils.js
+++ b/src/helpers/utils.js
@@ -25,13 +25,14 @@ exports.getLocaleCodes = getLocaleCodes
 
 /**
  * Retrieve page's options from the module's configuration for a given route
- * @param  {Object} route    Route
- * @param  {Object} pages    Pages options from module's configuration
- * @param  {Array} locales   Locale from module's configuration
- * @param  {String} pagesDir Pages dir from Nuxt's configuration
- * @return {Object}          Page options
+ * @param  {Object} route         Route
+ * @param  {Object} pages         Pages options from module's configuration
+ * @param  {Array} locales        Locale from module's configuration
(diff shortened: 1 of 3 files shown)


src/helpers/utils.js
```diff
@@ -25,13 +25,14 @@ exports.getLocaleCodes = getLocaleCodes
...
- * @param  {Object} route    Route
- * @param  {Object} pages    Pages options from module's configuration
- * @param  {Array} locales   Locale from module's configuration
- * @param  {String} pagesDir Pages dir from Nuxt's configuration
- * @return {Object}          Page options
+ * @param  {Object} route         Route
+ * @param  {Object} pages         Pages options from module's configuration
+ * @param  {Array} locales        Locale from module's configuration
+ * @param  {String} pagesDir      Pages dir from Nuxt's configuration
+ * @param  {String} defaultLocale Pages dir from Nuxt's configuration
```
lmichelin: the description seems to have been copy-pasted from the previous line 😕

jeanphilippeds: fixed :stuck_out_tongue_winking_eye:

src/helpers/utils.js
```diff
@@ -47,16 +48,22 @@ exports.getPageOptions = (route, pages, locales, pagesDir) => {
...
-      // Set custom path if any
-      options.paths[locale] = pageOptions[locale]
-    }
-  })
+
+  // Remove disabled locales from page options
+  options.locales = options.locales.filter(locale => pageOptions[locale] !== false)
+
+  // Construct paths object
+  options.paths = options.locales
```
paulgv: We probably don't want to reassign `options.path` here.

```suggestion
  options.locales
```

jeanphilippeds: That was so wrong :/
I was using reduce before, forgot to change that line. Thanks!

rchl:
@jeanphilippeds I know this is an old PR but I've just stumbled upon the behavior introduced by this change and I do wonder how it does make sense...

When we have
 - two locales: `en` and `fr`
 - the default locale is `fr`
 - we have an `about.vue` page for which we set custom path:
```js
{
  'fr': '/about-fr'
}
```

With this change here the English locale will also use the `about-fr` name for the about page. This doesn't seem right to me.

Without this change, the `en` locale would use the `about` name.

Maybe the intention here was to handle a case like for example:
 - three locales: `en`, `fr`, `es`
 - the default locale is `en`
 - only `fr` locale has overridden path:
```js
{
  'fr': '/about-fr'
}
```

And then both the `en` and `es` locales default to `about`. That makes sense but that would also happen without this change. So I don't really get it.

PR #357, opened 2019-07-11
(fix): localePath in hashmode

vanhoofmaarten:

When using Nuxt's router in [hash mode](https://nuxtjs.org/api/configuration-router#mode) `localePath` returns a string containing a hash which breaks the router functionality.

#### Replication:
https://codesandbox.io/s/nuxt-i18n-localepath-hash-mode-5xfz4

Both the _About_ button on the [homepage](https://5xfz4.sse.codesandbox.io/#/) and the _Back home_ button on the [about page](https://5xfz4.sse.codesandbox.io/#/about) do not resolve.

Vue-Router seems to work when you provide a string of a route without a hash in front.

Make `localePath` return the resolved routes `route.fullPath` instead of the `href`. `route.fullPath` does not include the starting hash character.

I added a few extra tests to specifically test for these conditions. Because `jsdom`, used by Nuxt's `renderAndGetWindow` does not allow navigation, I could not provide any validation after a programmatic router.push. I could only provide testing for the returned localized route.

At this time I cannot find breaking changes caused by this PR. 

I happily await your feedback.

### DIFF
--- a/src/plugins/routing.js
+++ b/src/plugins/routing.js
@@ -37,16 +37,8 @@ function localePathFactory (i18nPath, routerPath) {
 
     // Resolve localized route
     const router = this[routerPath]
-    const resolved = router.resolve(localizedRoute)
-    let { href } = resolved
-
-    // Remove baseUrl from href (will be added back by nuxt-link)
-    if (router.options.base) {
-      const regexp = new RegExp(router.options.base)
-      href = href.replace(regexp, '/')
-    }
-
-    return href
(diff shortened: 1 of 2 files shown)


paulgv:
Thank you @vanhoofmaarten! The changes LGTM. I'll release this in the next major version.

paulgv:
Using the `fullPath` feels way better, although I wonder why we were not already using it. Hopefully this doesn't break some edge-case I might have forgotten about. The module badly lacks tests in that regard... So thank you even more for adding some test cases!

PR #363, opened 2019-07-14
refactor: Rename and flatten vuex options

paulgv:
- `vuex.mutations.setLocale` has been renamed to `vuex.syncLocale`
- `vuex.mutations.setMessages` has been renamed to `vuex.syncMessages`

BREAKING CHANGE: Store module's options have been flattened and renamed


rchl:
Maybe renaming dynamicRouteParams to syncRouteParams (it syncDynamicRouteParams) would also make sense?

paulgv:
@rchl Sounds good, I'll change it!

PR #367, opened 2019-07-16
Add description of iso code to documents

munierujp:
I think that documents should show clearly what kind of value to use for `iso` of options.
So, I've add description of iso code to documents as comment of source code.

### DIFF
--- a/docs/options-reference.md
+++ b/docs/options-reference.md
@@ -18,9 +18,9 @@ Here are all the options available when configuring the module and their default
   //   { code: 'fr', iso: 'fr-FR', file: 'fr.js' },
   //   { code: 'es', iso: 'es-ES', file: 'es.js' }
   // ]
-  //   `iso` value should have:
-  //   - code of ISO 639-1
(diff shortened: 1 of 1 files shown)


docs/options-reference.md
```diff
@@ -18,6 +18,9 @@ Here are all the options available when configuring the module and their default
   //   { code: 'fr', iso: 'fr-FR', file: 'fr.js' },
   //   { code: 'es', iso: 'es-ES', file: 'es.js' }
   // ]
+  //   `iso` value should have:
+  //   - code of ISO 639-1
+  //   - code of ISO 639-1 and code of ISO 3166-1 alpha-2, with a hyphen
```
rchl: Should it rather be `should have either:`?
Also maybe add an example for each?

munierujp: Thanks, I think `should have either` is better.
And I'll add example.

munierujp: I've done in b3585f1 💪

paulgv:
Thank you @munierujp!

PR #395, opened 2019-08-13
chore: collect coverage from src

aaharu:


paulgv:
Makes much more sense, thanks!

PR #409, opened 2019-08-22
feat: Add NO_PREFIX strategy + setLocale function

richie3366:
For those who only want to use SEO, lazy locale load, cookie store and/or browser-language detection without having to deal with sub-domains nor route prefixes nor route aliases, the `no_prefix` Strategy will allow nuxt-i18n to fit their needs.

I've also added a `setLazyLocale` function in `app.i18n` to let developers call it programatically.


src/plugins/routing.js
```diff
@@ -27,6 +27,8 @@ function localePathFactory (i18nPath, routerPath) {
     // Match route without prefix for default locale
     if (locale === defaultLocale && STRATEGY === STRATEGIES.PREFIX_AND_DEFAULT) {
       name += routesNameSeparator + defaultLocaleRouteNameSuffix
+    } else if (STRATEGY === STRATEGIES.NO_PREFIX) {
```
rchl: Nit: Sorry but can you instead use ternary expression on the line that declares `name`. I find the flow a bit weird when variable is assigned something just to be overridden later like that.

Maybe like this:
```suggestion
    let name = route.name + (STRATEGY === STRATEGIES.NO_PREFIX ? '' : routesNameSeparator + locale)
```

rchl:
I think we should have a warning (console.warn) in `localePath` function that triggers when `locale` argument is provided. Basically there is no way to handle that with `no-prefix` strategy so we should warn that passing locale is unsupported in that case.

rchl:
Also this code should be disabled with no-prefix strategy:
https://github.com/nuxt-community/nuxt-i18n/blob/ff6937bf8bebd3d98d51abf2098046b5323b13f7/src/templates/middleware.js#L102-L104
It would also be problematic if you would follow my advice and trigger warning when calling `localePath` with `locale` argument.

rchl:
Please also add two new APIs here:
https://github.com/nuxt-community/nuxt-i18n/blob/ff6937bf8bebd3d98d51abf2098046b5323b13f7/types/vue.d.ts#L45-L54

richie3366:
Thanks for your appreciation & for all your comments @rchl!

Some of the things you noticed were caused by the fact I had implemented them 6 months ago (without making a PR because I thought it may be undesired since it's kinda the core/main functionnality of this module) in a different context, and I applied the same changes without checking if it could have a better fit.

I'll look further into your comments during the day (CEST timezone) and I'll do my best to answer &/or comply to them.

src/helpers/routes.js
```diff
@@ -67,8 +67,11 @@ exports.makeRoutes = (baseRoutes, {
         continue
       }
 
-      // Make localized route name
-      localizedRoute.name = name + routesNameSeparator + locale
+      // Strategy NO_PREFIX implies that route are named without any suffix
+      if (strategy !== STRATEGIES.NO_PREFIX) {
+        // Make localized route name
+        localizedRoute.name = name + routesNameSeparator + locale
+      }
```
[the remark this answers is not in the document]
richie3366: You are definitely right!
Not calling `extendRoutes` in module's index does the job without any error/warn. Plus it avoids iterating purposelessly on the routes array.
I remember the struggle I had on that part, I'm quite disappointed that I hadn't figured out the simplest way to do it.

src/helpers/routes.js
```diff
@@ -58,7 +58,7 @@ exports.makeRoutes = (baseRoutes, {
       const locale = componentOptions.locales[i]
       const { name } = route
       let { path } = route
-      const localizedRoute = { ...route }
+      const localizedRoute = { ...route, name }
```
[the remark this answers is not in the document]
richie3366: Agreed. I really don't know why I did it in the past.

src/plugins/main.js
```diff
@@ -78,6 +79,19 @@ export default async (context) => {
 
   const detectBrowserLanguage = <%= JSON.stringify(options.detectBrowserLanguage) %>
   const { useCookie, cookieKey } = detectBrowserLanguage
+  const { loadLanguageAsync } = require('./utils')
```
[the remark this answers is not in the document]
richie3366: I wanted to declare it near after the other declarations, but I've made the requested change.

src/plugins/main.js
```diff
@@ -133,19 +157,25 @@ export default async (context) => {
...
     locale = domainLocale ? domainLocale : locale
-  } else {
+  } else if (!noPrefix) {
     const routesNameSeparator = '<%= options.routesNameSeparator %>'
     const defaultLocaleRouteNameSuffix = '<%= options.defaultLocaleRouteNameSuffix %>'
 
     const routeLocale = getLocaleFromRoute(route, routesNameSeparator, defaultLocaleRouteNameSuffix, app.i18n.locales)
     locale = routeLocale ? routeLocale : locale
+  } else {
+
```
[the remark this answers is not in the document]
richie3366: Yes, it is an habit from another project where we decided the opposite. Maybe it should be declared into the linter config file(?)

src/plugins/main.js
```diff
@@ -114,6 +128,16 @@ export default async (context) => {
   app.i18n.beforeLanguageSwitch = <%= options.beforeLanguageSwitch %>
   app.i18n.onLanguageSwitched = <%= options.onLanguageSwitched %>
   app.i18n.setLocaleCookie = setLocaleCookie
+  app.i18n.getLocaleCookie = getLocaleCookie
+
+  app.i18n.setLazyLocale = async (locale) => {
+    await loadLanguageAsync(context, locale)
+    app.i18n.locale = locale
+
+    if (useCookie)
```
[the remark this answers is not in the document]
richie3366: Also an habit from another project. Fixed.

src/plugins/main.js
```diff
@@ -114,6 +128,16 @@ export default async (context) => {
   app.i18n.beforeLanguageSwitch = <%= options.beforeLanguageSwitch %>
   app.i18n.onLanguageSwitched = <%= options.onLanguageSwitched %>
   app.i18n.setLocaleCookie = setLocaleCookie
+  app.i18n.getLocaleCookie = getLocaleCookie
+
+  app.i18n.setLazyLocale = async (locale) => {
```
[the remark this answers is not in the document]
richie3366: Indeed, done.

src/plugins/main.js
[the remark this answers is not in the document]
rchl: It's hard if not impossible to make eslint cooperate with lodash templates. I haven't found any plugin that would work properly.

richie3366:
@rchl: I think I've made all the requested changes. Do you advise to rewrite my commit (& force push) or to create an additional one?

rchl:
Feel free to add new commits. It's easier to review that way. I can retire stuff when merging. Thanks

rchl:
Can't comment without seeing the code but my idea for warnings was to trigger them only when locale is passed AND it doesn't match current locale. Is that how you did it?

richie3366:
> Can't comment without seeing the code but my idea for warnings was to trigger them only when locale is passed AND it doesn't match current locale. Is that how you did it?

I compared the values like you said:
* `localePath`: when `locale` differs from current locale:
```javascript
    if (!locale) return // <- was already there

    if(STRATEGY === STRATEGIES.NO_PREFIX && locale !== this.$i18n.locale) {
      console.warn('[<%= options.MODULE_NAME %>] localePath is unsupported when using no_prefix strategy', locale, this.$i18n.locale)
    }
```


* `switchLocalePath`: when `locale` differs from default locale:
```javascript
    if(STRATEGY === STRATEGIES.NO_PREFIX && locale && locale !== defaultLocale) {
      console.warn('[<%= options.MODULE_NAME %>] switchLocalePath is unsupported when using no_prefix strategy', locale, defaultLocale)
    }
```

This is what I have in my console:
Screenshot at 15-19-14

rchl:
Ok. Feel free to either investigate more or leave it out. I can look later.

richie3366:
> Ok. Feel free to either investigate more or leave it out. I can look later.

You seem to have a good knowledge about the mechs/code of this lib, I think you'll find the fix quicker 
(and maybe cleaner) than me

richie3366:
Whoops, forgot to run the tests before committing, something broke in between, sorry for that.

src/plugins/routing.js
```diff
@@ -16,13 +16,17 @@ function localePathFactory (i18nPath, routerPath) {
     locale = locale || this[i18nPath].locale
     if (!locale) return
 
+    if(STRATEGY === STRATEGIES.NO_PREFIX && locale !== this.$i18n.locale) {
+      console.warn('[<%= options.MODULE_NAME %>] localePath is unsupported when using no_prefix strategy', locale, this.$i18n.locale)
```
rchl: It's not that `localePath` is unsupported but passing non-current locale is. So the warning should be earlier, before `locale` is overwritten so that you can check what was passed and compare to current.

Also, just printing `locale` and `this.$i18n.locale` as part of the warning, without explaining what those are is not ideal for production code. But I guess you did that just for debugging.

src/plugins/routing.js
```diff
@@ -49,6 +53,10 @@ function switchLocalePathFactory (i18nPath) {
   const LOCALE_CODE_KEY = '<%= options.LOCALE_CODE_KEY %>'
 
   return function switchLocalePath (locale) {
+    if(STRATEGY === STRATEGIES.NO_PREFIX && locale !== defaultLocale) {
```
rchl: Nit: space after `if` missing.

src/plugins/routing.js
```diff
@@ -16,13 +16,17 @@ function localePathFactory (i18nPath, routerPath) {
     locale = locale || this[i18nPath].locale
     if (!locale) return
 
+    if(STRATEGY === STRATEGIES.NO_PREFIX && locale !== this.$i18n.locale) {
```
rchl: Nit: space after `if` missing.

src/plugins/routing.js
```diff
@@ -49,6 +53,10 @@ function switchLocalePathFactory (i18nPath) {
   const LOCALE_CODE_KEY = '<%= options.LOCALE_CODE_KEY %>'
 
   return function switchLocalePath (locale) {
+    if(STRATEGY === STRATEGIES.NO_PREFIX && locale !== defaultLocale) {
+      console.warn('[<%= options.MODULE_NAME %>] switchLocalePath is unsupported when using no_prefix strategy', locale, defaultLocale)
```
rchl: Same as with other warning. `switchLocalePath` is supported but only when passing current locale or not passing anything.

types/vue.d.ts
```diff
@@ -50,7 +50,9 @@ declare namespace NuxtVueI18n {
       routesNameSeparator: string
       beforeLanguageSwitch: () => any
       onLanguageSwitched: () => any
+      getLocaleCookie: () => null | string
```
rchl: Should return `string | undefined | null` since js-coookie can return `undefined` and our method here can return `null`.

js-cookie: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/47f0eec500962b5beec44f18b7ae33b843c5d699/types/js-cookie/index.d.ts#L51
cookie: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/47f0eec500962b5beec44f18b7ae33b843c5d699/types/cookie/index.d.ts#L113

types/vue.d.ts
```diff
@@ -50,7 +50,9 @@ declare namespace NuxtVueI18n {
...
       setLocaleCookie: (locale: string) => undefined
+      setLocale: (locale: string) => undefined
```
rchl: Should return `Promise<undefined>`.

HapLifeMan:
The majority of your request changes are problems that would not occur with a stricter ESLint config, what do you think about complete the `.eslintrc.js` file @rchl? That would improve **a lot** every contribution to this repo and save a lot of time for each contributor and reviewer. 🙂

rchl:
@HapLifeMan If you can figure out a config that would work in files that contain lodash templates then I'm all ears. I've failed before. :)

rchl:
As for the warnings, it's due to seo-head.js adding `alternate` links for all locales. That code needs to be disabled for `NO_PREFIX` locale as we have no alternate links then. I think it's OK to keep the code that adds `og:locale:alternate` enabled.

src/plugins/main.js
```diff
@@ -114,6 +127,18 @@ export default async (context) => {
...
+
+  const { loadLanguageAsync } = require('./utils')
+  app.i18n.setLocale = async (locale) => {
+    await loadLanguageAsync(context, locale)
+    app.i18n.locale = locale
+
+    if (useCookie) {
+      setLocaleCookie(locale)
+    }
+  }
```
rchl: We should also call `syncVuex`.

There is also an issue with this API when using other than `no-prefix` strategies. It will change locale but route URL will stay unchanged (with previous locale prefix). I'm not sure what's the best way to handle that - two options I see are: call nuxt's `redirect` or replace route usign `VueRouter.replace()`. Both can be destructive but I guess that should be OK... I can handle that myself later.

HapLifeMan:
I've updated files according to your last request changes!

About the ESLint, we have a very strict ESLint config where we are working and this is not my repo so I wouldn't want to oblige contributors to follow our personal ESLint rules without your permission.
But I'm pretty sure you can work with [this plugin](https://github.com/ota-meshi/eslint-plugin-lodash-template). If you're open to ESLint config suggestions that may involve a lot of syntax changes, let me know and I'll be happy to share it here 🙂

rchl:
This repo is already using pretty strict rules that would catch all the issues I've pointed out. Problem is that all files with templates are ignored. I've looked at that plugin before and had some problems but maybe those problems don't apply here. I'll check again later.

Will look at your changes a bit later when I'm home. Thanks :)

HapLifeMan:
You're welcome, I said that because we have a 60 rules ESLint config 😉

src/plugins/routing.js
```diff
@@ -1,19 +1,25 @@
...
-  const STRATEGIES = <%= JSON.stringify(options.STRATEGIES) %>
-  const STRATEGY = '<%= options.strategy %>'
-  const defaultLocale = '<%= options.defaultLocale %>'
   const defaultLocaleRouteNameSuffix = '<%= options.defaultLocaleRouteNameSuffix %>'
 
   return function localePath (route, locale) {
     // Abort if no route or no locale
     if (!route) return
+
+    if (STRATEGY === STRATEGIES.NO_PREFIX && locale !== this.$i18n.locale) {
```
rchl: Should also have `&& locale` condition so that it doesn't warn when not passing locale.

src/plugins/routing.js
```diff
@@ -49,6 +55,10 @@ function switchLocalePathFactory (i18nPath) {
   const LOCALE_CODE_KEY = '<%= options.LOCALE_CODE_KEY %>'
 
   return function switchLocalePath (locale) {
+    if (STRATEGY === STRATEGIES.NO_PREFIX && locale !== defaultLocale) {
```
rchl: Should be checking `locale !== this.$i18n.locale` instead. And missing `&& locale` condition.

rchl:
@HapLifeMan this project is extending this config: https://github.com/standard/eslint-config-standard/blob/master/eslintrc.json
So it has quite a bit of rules. :)

src/plugins/routing.js
rchl: Missed this before: should also use `this[i18nPath].locale`.

src/plugins/routing.js
rchl: Also use `this[i18nPath].locale` instead.

HapLifeMan:
Updated! Oh I didn't saw the `extends`, sorry about that 🙈

rchl:
I'm now happy with those changes.

Some remaining things that can be done separately:
 - problem I've mentioned in one of the comments:
> There is also an issue with this API when using other than no-prefix strategies. It will change locale but route URL will stay unchanged (with previous locale prefix). I'm not sure what's the best way to handle that - two options I see are: call nuxt's redirect or replace route using VueRouter.replace(). Both can be destructive but I guess that should be OK... I can handle that myself later.
 - improving test coverage. Testing needs some cleanup...

rchl:
@HapLifeMan 
 - created #415 for improving linting but as you can see here, I'm not having too much luck.
 - I'm waiting for https://github.com/nuxt-community/module-test-utils/pull/2 before I'll progress further here

richie3366:
HapLifeMan's commit : `await` was needed to fix a bug in our project, reason: vue-i18n was loading before nuxt-i18n had finished to init & load the default locale

My recent commits: cf. commit descriptions

src/plugins/main.js
```diff
@@ -148,13 +169,7 @@ export default async (context) => {
...
   app.i18n.setLocaleCookie = setLocaleCookie
   app.i18n.getLocaleCookie = getLocaleCookie
-  app.i18n.setLocale = async locale => {
-    await loadAndSetLocale(locale)
-
-    if (useCookie) {
-      setLocaleCookie(locale)
-    }
-  }
+  app.i18n.setLocale = loadAndSetLocale
```
rchl: Should public API have access to `init` parameter? I suppose not in which case it would be better to do:
```js
app.i18n.setLocale = locale => loadAndSetLocale(locale)

File: src/plugins/main.js [hunk 18]
```diff
@@ -118,25 +120,44 @@ export default async (context) => {
...
+    }
 
-    let messages
+    // Abort if newLocale did not change
+    if (newLocale === app.i18n.locale) {
+      return
+    }
 
+    const oldLocale = app.i18n.locale
+    app.i18n.beforeLanguageSwitch(oldLocale, newLocale)
```
rchl: Actually both `beforeLanguageSwitch` and `onLanguageSwitched` should not trigger on initial setting of the locale also as it's not really a "switch" but more a "initialization".

src/plugins/main.js
```diff
@@ -105,6 +120,45 @@ export default async (context) => {
...
+      app.i18n.onLanguageSwitched(oldLocale, newLocale)
+      await syncVuex(newLocale, messages)
+    } else {
+      // Lazy-loading disabled
+      app.i18n.locale = newLocale
+      app.i18n.onLanguageSwitched(oldLocale, newLocale)
+      await syncVuex(newLocale, app.i18n.getLocaleMessage(newLocale))
+    }
+
+    const routeName = route && route.name ? app.getRouteBaseName(route) : 'index'
```
rchl: Should be moved inside `if`.

src/plugins/main.js
```diff
@@ -105,6 +120,45 @@ export default async (context) => {
     }
   }
 
+  const loadAndSetLocale = async (newLocale, init = false) => {
```
rchl: I really don't like this `init` argument. It's a very generic name that changes code flow in unclear (unless you dig into the code) ways.

I don't really have much better suggestion but at least we could make it a bit clearer at call site by using options object:
```suggestion
  const loadAndSetLocale = async (newLocale, { initialSetup = false } = {}) => {
```

src/plugins/main.js
```diff
@@ -118,25 +120,44 @@ export default async (context) => {
...
+      app.i18n.onLanguageSwitched(oldLocale, newLocale)
+      await syncVuex(newLocale, app.i18n.getLocaleMessage(newLocale))
     }
 
-    app.i18n.locale = locale
-    await syncVuex(locale, messages)
+    const routeName = route && route.name ? app.getRouteBaseName(route) : 'index'
+
+    if (!init && STRATEGY !== STRATEGIES.NO_PREFIX) {
+      redirect(app.localePath(Object.assign({}, route , {
```
rchl: (Just some thoughts that don't have to be handled now.)

On the client side we have an option of using VueRouter to push/replace the route. That would make more sense if that API is called manually I guess. But maybe not when middleware calls it (even on the client)... Something to investigate.

src/plugins/main.js
```diff
@@ -118,25 +120,44 @@ export default async (context) => {
...
-      if (app.i18n.fallbackLocale && locale !== app.i18n.fallbackLocale) {
-        await loadLanguageAsync(context, app.i18n.fallbackLocale)
-      }
-
-      messages = await loadLanguageAsync(context, locale)
+      const { loadLanguageAsync } = require('./utils')
+      const messages = await loadLanguageAsync(context, newLocale)
+      app.i18n.locale = newLocale
+      app.i18n.onLanguageSwitched(oldLocale, newLocale)
+      await syncVuex(newLocale, messages)
```
rchl: Those three lines above can be shared, just need to load messages in `if` branches.

rchl:
> HapLifeMan's commit : await was needed to fix a bug in our project, reason: vue-i18n was loading before nuxt-i18n had finished to init & load the default locale

Clearly a right thing to do. My bad for forgetting it.

src/plugins/main.js
rchl: Actually arrow function might not work so:
```js
app.i18n.setLocale = function(locale) { return loadAndSetLocale(locale); }
```

src/plugins/main.js
richie3366: If I'm not wrong, it was already behaving with redirect prior to this PR. But I agree with you, it may be a good idea to investigate further after merging.

rchl: "Behaving" yes but I'm thinking of behaving better since that API is now public so it will be called from user action and it would be better to use `router.replace` rather than `location.href` which that API does on the client side.

richie3366:
> I'll keep it unapproved for now as I'm still blocked with tests.

Okay, if you need anything else from me, feel free to ask!

HapLifeMan:
> Will rebase and merge this tomorrow after you agree that it's ready. :)
> Thank you both for your work.

Thank you for your suggestions and improvements! This is really a good thing to have an active community open to new features like that. You can rebase & merge 😄

rchl:
@paulgv have you tried fix in #418?

rchl:
Actually I can still reproduce with the fix. Will look more into it.

PR #413, opened 2019-08-24
test: use @nuxtjs/module-test-utils testing helper

rchl:
This cleans up tests a bit and also uses getPort package which makes
it possible to use random port for testing so that it doesn't conflict
with anything or tests itself.

Due to how utils work, moved test file to the root of the 'test'
directory instead of being inside fixtures directory.

Renamed 'fixtures' directory to 'fixture' as that is what utils expect.


rchl:
I'll take liberty of approving it myself as it's just a testing change that doesn't break testing. :)

Issue #411, opened 2019-08-23
differentDomains totally crashes browser


seltsam23:
### Version

[6.0.2](https://github.com/nuxt-i18n/releases/tag/6.0.2)

### Reproduction link

[https://github.com/paulgv/nuxt-i18n-example](https://github.com/paulgv/nuxt-i18n-example)

### Steps to reproduce

1.) Clone example Repo
2.) Update nuxt-i18n & nuxt to newest version
3.) Change config/index.js as shown below
4.) Make hosts-file entries for domains
5.) Make sure you don't have  the i18n_redirected cookie set in your browser / clear cookies 

``` /etc/hosts
127.0.0.1 en.nuxt.breaks
127.0.0.1 fr.nuxt.breaks
127.0.0.1 es.nuxt.breaks
```

``` /config/index.js
const I18N = {
  differentDomains:true,
  locales: [
    {
      code: 'en',
      iso: 'en-US',
      name: 'English',
      domain: 'en.nuxt.breaks:3000'
    },
    {
      code: 'fr',
      iso: 'fr-FR',
      name: 'Français',
      domain: 'fr.nuxt.breaks:3000'
    },
    {
      code: 'es',
      iso: 'es-ES',
      name: 'Español',
      domain: 'es.nuxt.breaks:3000'
    }
  ],
  defaultLocale: 'en',
  parsePages: false,
  pages: {
    'about': {
      fr: '/a-propos',
      en: '/about-us'
    },
    'posts': {
      fr: '/articles'
    },
    'post/_id': {
      fr: '/article/:id?',
      es: '/articulo/:id?'
    },
    'dynamicNested/_category': {
      fr: 'imbrication-dynamique/:category'
    }
  },
  vueI18n: {
    fallbackLocale: 'en',
    messages: { en, fr, es }
  }
}
```

### What is expected ?

Navigation works as expected, i18n_redirected cookie is set.

### What is actually happening?

Browser is freezing as soon as you try to navigate the page via the nav-menu.  i18n_redirected cookie is not set. 

### Additional comments?

If you set the cookie  i18n_redirected  manually it works as expected, but until then the browser just crashes.
It feels like there is some redirect-looping going on in the background. No Errors are thrown anywhere.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c288">#c288</a>)</em></sub></div>

[...]

PR #418, opened 2019-08-27
fix: missing redirect to domain with differentDomains enabled

rchl:
Changes in middleware.js are just indentation fixes.

### DIFF
--- a/src/plugins/main.js
+++ b/src/plugins/main.js
@@ -161,6 +161,7 @@ export default async (context) => {
     await syncVuex(newLocale, app.i18n.getLocaleMessage(newLocale))
 
     if (!initialSetup && STRATEGY !== STRATEGIES.NO_PREFIX) {
+      const route = app.i18n.__route
       const routeName = route && route.name ? app.getRouteBaseName(route) : 'index'
 
       redirect(app.localePath(Object.assign({}, route , {
@@ -180,6 +181,9 @@ export default async (context) => {
   app.i18n.setLocaleCookie = setLocaleCookie
   app.i18n.getLocaleCookie = getLocaleCookie
   app.i18n.setLocale = (locale) => loadAndSetLocale(locale)
+
+  // Current route. Updated from middleware also.
+  app.i18n.__route = route
 
   // Extension of Vue
   if (!app.$t) {
--- a/src/plugins/main.js
+++ b/src/plugins/main.js
@@ -143,6 +143,12 @@ export default async (context) => {
     // Lazy-loading enabled
     if (lazy) {
       const { loadLanguageAsync } = require('./utils')
+
+      // Load fallback locale.
+      if (app.i18n.fallbackLocale && newLocale !== app.i18n.fallbackLocale) {
+        await loadLanguageAsync(context, app.i18n.fallbackLocale)
+      }
+
       await loadLanguageAsync(context, newLocale)
     }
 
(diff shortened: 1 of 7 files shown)


rchl:
@richie3366 @HapLifeMan Issue caused by refactoring (duplication removal).

rchl:
Actually this is not the correct fix. The problem is that `app.i18n.locale` is not set and it's still skipped with my fix. I'll think more about it. It's quite fragile stuff.

I think automated test could be tricky as it requires specific domain setup (using hosts file for example). That's probably not an option for test...

rchl:
> Indeed, it's a bit touchy to write tests
Actually found nicer way. Since code gets locale's domain from `Host` header (on the server), I just pass whatever I please through headers.

Fixed problem with locale not being properly initialized with `differentDomains` enabled and also found another regression with lazy loading. The code that loaded fallback locale didn't survive refactorings. Now restored.

Also added tests for both issues.

src/plugins/main.js
```diff
@@ -119,9 +119,17 @@ export default async (context) => {
     }
   }
 
+  const redirectToNewLocaleRoute = newLocale => {
```
rchl: Hmm...
I've refactored this into separate function to call it also from `!initialSetup && app.i18n.differentDomains` to match old code. Currently the redirect will not be called from middleware for `differentDomains`.

But I have suspiction that current code is actually correct because it's not possible to redirect to different domain using that code (it uses `localePath` which doesn't consider different domains -- for some reason...). So I think that code has actually caused infinite redirect bug in old code.

rchl: But then... It won't be able to redirect to different domain but I think it can work when route has overriden path through component/page options...

I guess I need to test this.
So many variables...

rchl: Actually checked the concerning part and I think it's OK right now.

Beheavior differs from old code in this situation:
 - differentDomains = true
 - detectBrowserLanguage != false

Previously redirect was called, now it isn't. I think that's OK because that redirect can't redirect to different domain anyway and also it's based on browser language so that's not really compatible with having different domains which should itself dictate which language to choose.

Possibilly `detectBrowserLanguage` should even be force-disabled when `differentDomains` is set. Or alternatively, those could work together but code would need to be fixed to be able to redirect to different domain.

rchl: And I've verified now that it actually fixes #411 without having to explicitly disable `detectBrowserLanguage`.

rchl:
Issue reported by Paul:

> It's a bit late for me to notice this, but it looks like this PR might have broken something in the module, my first guess would be that it's related to the code that was removed from the middleware.
> 
> To reproduce the bug I'm seeing:
> 
> Run yarn dev:basic in the module's directory
> Visit http://localhost:3000/
> Click on About us
> Click on Français to switch locales
> I would expect to end up on http://localhost:3000/fr/a-propos but I'm actually redirected to http://localhost:3000/fr. The lang switcher's link itself is correct, so I assume that something in the middleware overrides the route and redirects to the homepage for whatever reason.
> 
> @richie3366 @rchl WDYT? Is there any way we can fix this quickly? Should we revert these changes for the time being?

rchl:
Right, I see the problem now. The `route` object in plugin and middleware is not the same object. Previously `redirect` call was based on respective `route`, now it's always called on `plugin`'s `route`.

 - `plugin` code runs on initial request to the page
 - `middleware` code runs on initial request to the page (after `plugin`) and then also on each SPA navigation

I'm just not quite sure how to test SPA navigation... Would require some sort of real browser environment. I'll investigate that after mergin the fix as this is critical to fix quickly.

rchl:
Rewrote history into 3 separate fixes.

PR #419, opened 2019-08-27
docs: Document 4.x to 5.x and 5.x to 6.x migration path

paulgv:


rchl:
Will have a deeper look in hour or two but after quick glance it looks good but would also mention seo option default value change. Also maybe modify change log to reference migration guide in notes for version 6?

PR #429, opened 2019-08-30
docs: Add link to discord and add info about types configuration

rchl:
Removing link to CMTY as nobody replies there (I just did a bit but I don't think I'll able to keep monitoring it).


paulgv:
Thanks!

PR #431, opened 2019-08-31
fix(types): various fixes to type definitions

rchl:
Changes:
 - exported `NuxtVueI18n` namespace to allow to annotate configuration
   object in `nuxt.config.ts`.
 - fixed various types in Options, made all optional to make them
   actually usable, added missing ones.
 - moved getLocaleCookie, setLocaleCookie and SetLocale to proper
   interface.
 - removed `return null` from `getLocaleCookie` to simplify types.
 - Updated NuxtI18nSeo interface to use VueMeta types.
 - Also removed `app.$t` API as it never worked - it needs to be called
   on the VueI18n instance instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,8 @@
   },
   "eslintIgnore": [
     "src/templates/*.*",
-    "src/plugins/*.*"
+    "src/plugins/*.*",
+    "**/*.d.ts"
   ],
   "files": [
     "src",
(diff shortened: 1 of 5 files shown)


rchl:
I've divided all changes into separate commits for book keeping and to make it easier to review. If you go commit-by-commit it should be quite clear what the changes are.

paulgv:
LGTM @rchl. I can't really speak for the TS part, hopefully @kevinmarrec can have give this a look.

rchl:
@kevinmarrec gave me an informal review on discord. :)

PR #434, opened 2019-09-02
fix: make `parsePages` compatible with typescript decorators

### LINKED ISSUE #76: Allow decorators (and TypeScript) in component

Allow using decorators inside component (and TypeScript).

I'm using TypeScript in `.vue` files with `nuxt-property-decorator`.
Since nuxt-i18n 3.x is using acron to parse each component files to check `i18n` property, it will throw exception on decorations (e.g. `@Component` from `nuxt-property-decorator`)

As I know, acron has a plugin [acron-es7](https://github.com/angelozerr/acorn-es7) can parse.
But the better solution is to load component files via webpack processor that webpack will solve the future syntax.

While using nuxt-i18n 3.x, I can use these syntax below in component file:

```
<script lang="ts">
import {
  Component,
  Vue,
} from 'nuxt-property-decorator'
import AppLogo from '~/components/AppLogo.vue'

@Component({ // <- acron (or nuxt-i18n) should not blame here
  components: {
    AppLogo
  }
})
export default class extends Vue {
}
</script>
```

<div align="right"><sub><em>This feature request is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt-community/nuxt-i18n/issues/c71">#c71</a>)</em></sub></div>

### LINKED ISSUE #408: Can't load the nuxt-i18n module in Nuxt with Typescript : SyntaxError: Unexpected character '@'
### Version

[v6.0.2](https://github.com/nuxt-i18n/releases/tag/v6.0.2)

### Reproduction link

[https://github.com/nuxt-community/nuxt-i18n](https://github.com/nuxt-community/nuxt-i18n)

### Steps to reproduce

1. Create fresh Nuxt 2.9.1 project
2. Make TypeScript works with the guide on https://typescript.nuxtjs.org/guide/setup.html
3. Install and add nuxt-i18n module
4. Create a class with @Component in a layout/page

### What is expected ?

* No SyntaxError: Unexpected character '@'
* Compatible with parsePages: true;

Using `@Component` in a page/layout vue single file throws a `SyntaxError: Unexpected character '@'`.
**=> works if parsePages is set to false**

### Additional comments?

New issue based on https://github.com/nuxt-community/nuxt-i18n/issues/128

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c287">#c287</a>)</em></sub></div>

rchl:
Resolves #408
Resolves #76

Thanks to @Seb-L for the code.


package.json
```diff
@@ -57,10 +58,9 @@
     ]
   },
   "dependencies": {
+    "@babel/parser": "^7.5.5",
+    "@babel/traverse": "^7.5.5",
     "@kazupon/vue-i18n-loader": "^0.4.0",
-    "acorn": "^7.0.0",
-    "acorn-dynamic-import": "^4.0.0",
-    "acorn-walk": "^7.0.0",
```
paulgv: It makes me happy to see us drop `acorn`! :)

docs/routing.md
```diff
@@ -99,13 +97,13 @@ export default {
...
 
-Make sure you set the `parsePages` option to `false` to disable acorn parsing and add your custom paths in the `pages` option:
+Make sure you set the `parsePages` option to `false` to disable babel parsing and add your custom paths in the `pages` option:
 
 ```js
 // nuxt.config.js
 
 ['nuxt-i18n', {
-  parsePages: false,   // Disable acorn parsing
+  parsePages: false,   // Disable babel parsing
```
paulgv: Just a question here: should we plan on dropping this option at some point? I don't feel too good about having 2 different ways for configuring page-specific stuff with `nuxt-i18n` (in-component options and in Nuxt's config). IMO, having the ability to change `nuxt-i18n`'s behaviour right in the Vue files feels much simpler and I would rather drop the other method if we can guarantee there won't be any compatibility issues like we had with `acorn` parsing. WDYT?

src/helpers/components.js
```diff
@@ -17,18 +15,32 @@ exports.extractComponentOptions = (path) => {
...
+
+  traverse(parsed, {
+    enter (path) {
+      if (path.node.type === 'Property') {
+        if (path.node.key.name === COMPONENT_OPTIONS_KEY) {
+          const data = script.substring(path.node.start, path.node.end)
+          try {
+            componentOptions = eval(`({${data}})`)[COMPONENT_OPTIONS_KEY] // eslint-disable-line
+          } catch (error) {
+            // Ignore
```
paulgv: Should we log the error here? If anything breaks here, it would be useful for user to have a message they can paste when reporting the error.

rchl: I'll add a warning.

But note that this is only catching `eval` of data, not parsing of the vue file itself.

Data can be wrong in two ways:
 - component option has JS syntax error - it will be caught by Nuxt later, when parsing same page. It's better to let Nuxt fail as it gives much better error and stacktrace than `eval` failing.
 - Code is formed correctly but not in the form that nuxt-i18n expects and thus won't eval successfully. For that case Nuxt will not fail at parsing the the vue file.

So yes, even though for the first case we wouldn't need a warning, I'll add it for both anyways (it's not that I can distinguish them easily anyway).

docs/routing.md
rchl: I agree that defining those in the vue file itself is much more fitting than doing that in module's options. That said, people hate breaking changes so I would feel a bit bad about dropping the other method... And it's fairly trivial to support module option alternative. So not sure really...

src/helpers/components.js
rchl: I've actually expanded try/catch to parsing also as that is the one that gives bad errors and Nuxt should report better one later anyway.

Errors when `eval`'ing or using Function constructor are not really catchable but try/catch wraps that part too just in case.

docs/routing.md
paulgv: Actually, my main concern is the documentation: it's a bit cumbersome to have to maintain documentation for these 2 different approaches, and I think that it makes onboarding harder for people who start using **nuxt-i18n**. I think we're fine for now but we might want to discuss this further for some next major release.

rchl: Right, then the best way to do it is to deprecate `pages` option (add console warning when used) and remove in next major release.

Issue #215, opened 2019-02-08
Support passing path to localePath for easier resolving of dynamic routes


rchl:
### What problem does this feature solve?

`localePath` takes either string or object.
String is expected to be route's internal name.
Object, with current implementation of `localePath`, will be passed to VueRouter's resolve function but the `name` property will get `___[locale]` appended to it.

The problem is that the object variant expects `name` to be set, otherwise it will create `name` value like `undefined___[locale]` and route will not match.

Given that it won't work to pass an object with just `{path: '/about'}`, for dynamic routes like `pages/_.vue`, we must instead do an awkward localePath call like `localePath({name: 'all', params: {0: 'about'}})` (haven't really tested but I think it should work). This is IMO too cumbersome and what's more, it requires knowledge of internal route's name. It's basically not user friendly.

### What does the proposed changes look like?

It should be possible to call code below to get localized path to dynamic route
```js
localePath({path: '/about'});
```
and let router figure out the path.

Note that with PREFIX_AND_DEFAULT mode, it should ideally only return path with locale prefix added, if resolving to non-default language OR when current route includes locale prefix.

So (with `en` being default locale):
 - `/about` >> localePath({path: '/about'}) >> `/about`
 - `/en/about` >> localePath({path: '/about'}) >> `/en/about`
 - `/fr/about` >> localePath({path: '/about'}) >> `/fr/about`

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c201">#c201</a>)</em></sub></div>

sambrezo:
I found this information to be very useful for decoupling my CMS and serving content with Nuxt. I'm using the slightly less awkward syntax `{name: 'all', params: ['about']}`, but support for the `path` property is definitely preferable.

I'd like to add that you must add your default locale last to the locales definition, eg:

```
{
  modules: [
    ['nuxt-i18n', {
      locales: [
        { code: 'en', name: 'English', iso: 'en-US' },
        { code: 'de', name: 'Deutsch', iso: 'de-DE' } // <-- Add this last
      ],
      defaultLocale: 'de'
    }]
}
```
So you get the following routes generated for your `_.vue` page in this particular order:
```
[{
  path: '/en/*',
  name: 'all___en'
},
{
  path: '/*',
  name: 'all___de'
}]
```
Else your localized dynamic route never matches.

[...]

PR #441, opened 2019-09-05
Add support for using 'path' parameter in localePath(...)

jengel3:
Hey! Trying to determine whether or not support for this is worth pursuing, or if this is already an issue that can't be supported for whatever reason. This also builds on #215.

Here's the issue:

If you need to dynamically generate a localized route, there is no way to do so because you need to pass the route's name to `localePath` for it to generate the correct path. 

In my case, I need to be able to generate a dynamic path based on a response from a remote database, which only knows of the route's path, not the name. 

My proposal is to add support for using the `path` key in `localePath({ ... })`. 

Example:

Given languages: `en` *(default)* and `ru`

```js
localePath({ path: '/dashboard' })
``` 

Will generate `/dashboard` and `/ru/dashboard` respectively.

This is comparable to using `localePath({ name: 'dashboard' })` or `localePath('dashboard')`, but in this instance we don't know the name of the route, so it can't be used.

In order to check whether or not the plugin should use the `path` parameter, I ensure it's presence and then double check that the `name` parameter is not also supplied, in which case we'll skip this because we have a more accurate way of resolving the route. This prevents conflict with the `switchLocalePath()` functionality, which passes both `name` and `route` to `localePath()` when generating the path to switch languages. 

Then I determine whether or not the locale needs to be prepended to the path based on a) not using the default locale and we're not using a strategy that prefixes the default locale, and b) not using the differentDomains functionality.

I'll of course add tests and so-on, again, assuming that this can be pursued.

### DIFF
--- a/src/plugins/routing.js
+++ b/src/plugins/routing.js
@@ -27,6 +27,14 @@ function localePathFactory (i18nPath, routerPath) {
       route = { name: route }
     }
 
+    // if route has a path defined but no route name, generate path without route
+    if (route.path && !route.name) {
+      // prepend locale if current is not default and not using different domains
+      const appendPath = locale !== defaultLocale && !this[i18nPath].differentDomains
+      const fullPath = (appendPath ? `/${locale}${route.path}` : route.path)
+      return fullPath
+    }
+
     // Build localized route options
     let name = route.name + (STRATEGY === STRATEGIES.NO_PREFIX ? '' : routesNameSeparator + locale)
 
--- a/src/plugins/routing.js
+++ b/src/plugins/routing.js
@@ -27,31 +27,42 @@ function localePathFactory (i18nPath, routerPath) {
       route = { name: route }
     }
 
-    // if route has a path defined but no route name, generate path without route
+    let localizedRoute
+    const router = this[routerPath]
+
     if (route.path && !route.name) {
-      // prepend locale if current is not default and not using different domains
-      const appendPath = !(locale === defaultLocale && strategy === STRATEGIES.PREFIX_EXCEPT_DEFAULT) && !this[i18nPath].differentDomains
-      const fullPath = (appendPath ? `/${locale}${route.path}` : route.path)
-      return fullPath
-    }
+      // if route has a path defined but no name, resolve full route using the path
+      const isPrefixed = (
+        // don't prefix default locale
+        !(locale === defaultLocale && STRATEGY === STRATEGIES.PREFIX_EXCEPT_DEFAULT) &&
+        // no prefix for any language
+        !(STRATEGY === STRATEGIES.NO_REFIX) &&
+        // no prefix for different domains
+        !this[i18nPath].differentDomains
+      )
 
-    // Build localized route options
-    let name = route.name + (STRATEGY === STRATEGIES.NO_PREFIX ? '' : routesNameSeparator + locale)
+      const path = (isPrefixed ? `/${locale}${route.path}` : route.path)
 
-    // Match route without prefix for default locale
-    if (locale === defaultLocale && STRATEGY === STRATEGIES.PREFIX_AND_DEFAULT) {
-      name += routesNameSeparator + defaultLocaleRouteNameSuffix
-    }
+      localizedRoute = Object.assign({}, route, { path })
+    } else {
+      // otherwise resolve route via the route name
+      // Build localized route options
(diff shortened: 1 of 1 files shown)


src/plugins/routing.js
```diff
@@ -27,6 +27,14 @@ function localePathFactory (i18nPath, routerPath) {
       route = { name: route }
     }
 
+    // if route has a path defined but no route name, generate path without route
+    if (route.path && !route.name) {
+      // prepend locale if current is not default and not using different domains
+      const appendPath = !(locale === defaultLocale && strategy === STRATEGIES.PREFIX_EXCEPT_DEFAULT) && !this[i18nPath].differentDomains
```
rchl: - Also NO_PREFIX strategy needs no prefix
 - variable `strategy` doesn't exist, `STRATEGY` does

src/plugins/routing.js
```diff
@@ -27,6 +27,14 @@ function localePathFactory (i18nPath, routerPath) {
...
+      const fullPath = (appendPath ? `/${locale}${route.path}` : route.path)
+      return fullPath
```
rchl: Not going through `router.resolve` means that this code won't handle `params` or `query`.

(I don't know if router will be able to handle route without name though...)

jengel3: I think the router should be able to resolve the it without the name. [Example here](https://router.vuejs.org/guide/essentials/navigation.html#programmatic-navigation) uses path + query in one of samples, and I'm assuming that's resolved through the router the same way the others samples are. Will fix.

jengel3:
Added support for resolving the `path` routes through vue-router. This adds support for the `query` option, but not `params`. However, vue-router doesn't support resolving path-based routes with parameters, per the [docs](https://router.vuejs.org/guide/essentials/navigation.html#programmatic-navigation).

> Note: params are ignored if a path is provided, which is not the case for query, as shown in the example above. Instead, you need to provide the name of the route or manually specify the whole path with any parameter:

But they do suggest interpolating the route path with the parameters yourself if it's necessary.

```js
const userId = '123'
router.push({ name: 'user', params: { userId } }) // -> /user/123
router.push({ path: `/user/${userId}` }) // -> /user/123
// This will NOT work
router.push({ path: '/user', params: { userId } }) // -> /user
```

It probably makes sense to use that same behavior here.

Will add some tests as well.

src/plugins/routing.js
```diff
@@ -27,23 +27,42 @@ function localePathFactory (i18nPath, routerPath) {
       route = { name: route }
     }
 
-    // Build localized route options
-    let name = route.name + (STRATEGY === STRATEGIES.NO_PREFIX ? '' : routesNameSeparator + locale)
+    let localizedRoute
+    const router = this[routerPath]
```
rchl: nit: This line moved up here but we don't use it until later so I would prefer its original location. :)

src/plugins/routing.js
```diff
@@ -27,23 +27,42 @@ function localePathFactory (i18nPath, routerPath) {
       route = { name: route }
     }
 
-    // Build localized route options
-    let name = route.name + (STRATEGY === STRATEGIES.NO_PREFIX ? '' : routesNameSeparator + locale)
+    let localizedRoute
```
rchl: nit: maybe `Object.assign` route right away and then just update needed property in individual branches.

src/plugins/routing.js
```diff
@@ -27,23 +27,42 @@ function localePathFactory (i18nPath, routerPath) {
...
+      // Build localized route options
+      let name = route.name + (STRATEGY === STRATEGIES.NO_PREFIX ? '' : routesNameSeparator + locale)
+
+      // Match route without prefix for default locale
+      if (locale === defaultLocale && STRATEGY === STRATEGIES.PREFIX_AND_DEFAULT) {
+        name += routesNameSeparator + defaultLocaleRouteNameSuffix
+      }
 
-    const localizedRoute = Object.assign({}, route, { name })
+      localizedRoute = Object.assign({}, route, { name })
```
rchl: nit: Can just be a `name` property update if you assign beforehand.

rchl:
I've done remaining fixes myself, rewritten history and commit message and merged as bbaa26652bf66897a3e3b0ae69e1baa7c5d26150
Thanks for working on this. :)

rchl:
And I've just realized it's a bit premature as we don't have tests yet. Do you want to add some still?

Issue #359, opened 2019-07-12
Child routes have no language prefix for non default languages


popescuandrei9mai:
### Version

[5.12.7](https://github.com/nuxt-i18n/releases/tag/5.12.7)

### Reproduction link

[https://codesandbox.io/s/codesandbox-nuxt-mpc6k](https://codesandbox.io/s/codesandbox-nuxt-mpc6k)

### Steps to reproduce

Press the Child 1 button from home page. Switch the language to English.

### What is expected ?

Route should change from '/despre/primul-copil'  to '/en/about/first-child'

### What is actually happening?

Route changes from '/despre/primul-copil' to '/about/first-child'

### Additional comments?

Note that the default language is romanian so child routes for english should get the '/en' prefix.
The problem seems to come from this line https://github.com/nuxt-community/nuxt-i18n/blob/master/src/helpers/routes.js#L116

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c259">#c259</a>)</em></sub></div>

woutercouvaras:
Hi @popescuandrei9mai, I did you checkout the code sanbox example here:

https://nuxtjs.org/examples/i18n#official-code-i18n-module-code-

The plugin and middleware files have some details around redirecting...it seems they redirect the local not to have a prefix, but if you want one, I would imagine you could use the same sort of logic to redirect to the default prefix if that's the behaviour you're after?

popescuandrei9mai:
Hi @woutercouvaras. This can be a fix for the issue but it looks more like a workaround. There should be no difference between parent and child routes regarding the prefix.

PR #462, opened 2019-09-24
fix: Route name missing for routes with children + Locale prefixes missing for child routes with custom paths

### LINKED ISSUE #356: Routes with children have no name thus localePath(route with children) always returns the root path
### Version

[5.12.7](https://github.com/nuxt-i18n/releases/tag/5.12.7)

### Reproduction link

[https://codesandbox.io/s/codesandbox-nuxt-8go7n](https://codesandbox.io/s/codesandbox-nuxt-8go7n)

### Steps to reproduce

On the home page, press the About button.

### What is expected ?

If the romanian language is selected, the button should redirect us to '/despre'.
If the english language is selected, the button should redirect us to 
'/en/about'

The button, no matter what language is selected, redirects us to '/'

### Additional comments?

The problem comes from this line: https://github.com/nuxt-community/nuxt-i18n/blob/master/src/helpers/routes.js#L74

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c258">#c258</a>)</em></sub></div>

rchl:
PR includes two separate (but in the same family) bugs, in two separate commits.
Refer to individual commits messages for more info.

Resolves #356, #359

### DIFF
--- a/src/helpers/routes.js
+++ b/src/helpers/routes.js
@@ -72,12 +72,13 @@ exports.makeRoutes = (baseRoutes, {
         continue
       }
 
-      // Make localized route name
-      localizedRoute.name = name + routesNameSeparator + locale
+      // Make localized route name. Name might not exist on parent route if child has same path.
+      if (name) {
+        localizedRoute.name = name + routesNameSeparator + locale
+      }
 
       // Generate localized children routes if any
       if (route.children) {
-        delete localizedRoute.name
         localizedRoute.children = []
         for (let i = 0, length1 = route.children.length; i < length1; i++) {
           localizedRoute.children = localizedRoute.children.concat(buildLocalizedRoutes(route.children[i], { locales: [locale] }, true, isExtraRouteTree))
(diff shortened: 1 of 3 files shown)


src/helpers/routes.js
```diff
@@ -77,7 +77,6 @@ exports.makeRoutes = (baseRoutes, {
 
       // Generate localized children routes if any
       if (route.children) {
-        delete localizedRoute.name
```
rchl: Actually I found out what issue removing route name was solving. In the case of directory structure like:
```
pages:
 - posts/index.vue
 - posts.vue
```
Nuxt will generate routes:
```
{
    path: '/posts',
    component: _f3e2cabc,
    children: [{
      path: '',
      component: _50b11fa5,
      name: 'posts'
    }]
  }
```

So the parent route doesn't have a name in this case, and if it had, VueRouter would trigger warning that child route is not accessible using `:to({ name: 'posts' })`.

So I need to look more into it. I guess the name should only be removed when there will be no child routes that would "shadow" parent's name.

rchl: Addendum: This is an issue *when combined with* the fact that we are always setting name for extra routes added for locales. Then we can end up with multiple identical names that shadow each other.

rchl: I believe I have it figured out now.

PR #465, opened 2019-09-27
Chore/extend test coverage

rchl:


src/helpers/routes.js
```diff
@@ -65,13 +64,6 @@ exports.makeRoutes = (baseRoutes, {
       let { path } = route
       const localizedRoute = { ...route }
 
-      // Skip if locale not in module's configuration
-      if (locales.indexOf(locale) === -1) {
-        // eslint-disable-next-line no-console
-        console.warn(`[${MODULE_NAME}] Can't generate localized route for route '${name}' with locale '${locale}' because locale is not in the module's configuration`)
-        continue
-      }
```
paulgv: Why are we removing this check?

rchl: I've mentioned the reason in the commit - https://github.com/nuxt-community/nuxt-i18n/pull/465/commits/01bc84d799988184d54d98d1abf79e77d66fda35

Basically, seems like with current logic it's impossible to trigger it. With earlier code merging and filtering locales, I couldn't find a way to trigger it.

paulgv: All right sounds good!

PR #473, opened 2019-10-04
docs: add dynamic custom route example

Iffabled:
This pull request adds an example of how to configure a custom path for a dynamic route using in-component options.


paulgv:
Thank you @Iffabled !

PR #478, opened 2019-10-08
docs: improve README.md

Iffabled:
`README.md` is the file newcomers see when they first visit the module's repository or NPM page. Currently, it gives absolutely no information about how the module works. This pull request aims to improve by adding information about features, setup and basic usage of this module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,44 +1,103 @@
-<h1 align="center">nuxt-i18n</h1>
-<p align="center">i18n features for your Nuxt project</p>
+# nuxt-i18n
 
-<p align="center">
-<a href="https://nuxt-community.github.io/nuxt-i18n/">Documentation</a> | <a href="./CHANGELOG.md">Release notes</a>
-</p>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,53 +114,68 @@ git clone git@github.com:<username>/nuxt-i18n.git modules/nuxt-i18n
...
 
-- Install the dependencies
+- Install the dependencies:
 
 ```sh
-cd nuxt-i18n/
-yarn
+cd nuxt-i18n
+yarn # yarn
+npm i # npm
```
rchl: Maybe let's skip whole clone/install deps steps for `Documentation` header since those are already listed for the project above?

rchl:
I like the changes. Thanks. :)

README.md
Iffabled: Then it would be unclear which of these steps apply to contributing to documentation and which don't.

rchl: But it's all the same repo. There is no difference in contributing to code or documentation really, apart from running `docs:dev` script to develop documentation.

But if you are insisting, I won't hold it against this review. It's not that important.

If keeping it, there is a bit of incosistency as it mentions `npm` step but other places only mention `yarn`.

rchl:
Thank you.

PR #488, opened 2019-10-15
fix: make switchLocalePath work from Nuxt plugin or middleware

### LINKED ISSUE #480: switchLocalePath() doesn't work in plugins
### Version

### Reproduction link

[https://codesandbox.io/embed/nuxt-i18n-switchlocalepath-bug-dbi7h](https://codesandbox.io/embed/nuxt-i18n-switchlocalepath-bug-dbi7h)

### Steps to reproduce

Create a Nuxt plugin and try using `switchLocalePath()` in it.

### What is expected ?

Function working: returning the locale path for the desired locale.

Function crashes with error:
> Cannot destructure property `params` of 'undefined' or 'null'.
> const { params, ...routeCopy } = this.$route

### Additional comments?

I'm using nuxt 2.10.0 and nuxt-i18n 6.3.0

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c318">#c318</a>)</em></sub></div>

rchl:
switchLocalePath didn't work from a Nuxt plugin or middleware
(from context.app) as it accessed properties on `this` that were not
accessible from there.

Fixed by creating a proxy interface that is consistent for both
Vue instances and Nuxt context so that switchLocalePath can access
same properties from it and get expected results.

Also:
  - Updated `differentDomains` tests to not use snapshot testing.
    Those actually were wrongly tested due to bug in test utils but
   the failure wasn't caught in snapshot (it was supposed to have
    SEO disabled).

Resolves #480

### DIFF
--- a/src/plugins/routing.js
+++ b/src/plugins/routing.js
@@ -12,130 +12,153 @@ import {
   vuex
 } from './options'
 
-function localePathFactory (i18nPath, routerPath) {
-  return function localePath (route, locale) {
-    // Abort if no route or no locale
-    if (!route) return
+function localePath (route, locale) {
+  // Abort if no route or no locale
+  if (!route) return
 
-    const i18n = this[i18nPath]
+  const { i18n } = this
 
-    if (strategy === STRATEGIES.NO_PREFIX && locale && locale !== i18n.locale) {
-      // eslint-disable-next-line no-console
-      console.warn(`[${MODULE_NAME}] Passing non-current locale to localePath is unsupported when using no_prefix strategy`)
-    }
+  if (strategy === STRATEGIES.NO_PREFIX && locale && locale !== i18n.locale) {
+    // eslint-disable-next-line no-console
+    console.warn(`[${MODULE_NAME}] Passing non-current locale to localePath is unsupported when using no_prefix strategy`)
+  }
 
-    locale = locale || i18n.locale
+  locale = locale || i18n.locale
 
-    if (!locale) return
+  if (!locale) return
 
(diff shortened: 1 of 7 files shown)


src/plugins/routing.js
```diff
@@ -12,148 +12,173 @@ import {
...
+  const name = this.getRouteBaseName()
+  if (!name) {
+    return ''
+  }
 
-    const { params, ...routeCopy } = this.$route
-    let langSwitchParams = {}
-    if (vuex && this.$store) {
-      langSwitchParams = this.$store.getters[`${vuex.moduleName}/localeRouteParams`](locale)
+  const route = this.route
```
paulgv: **nitpick:** Let's use object-destructuring here:

```suggestion
  const { route, store } = this
```

(and let's remove `const store = this.store` below)

src/plugins/routing.js
```diff
@@ -12,148 +12,173 @@ import {
...
-    return path
   }
+  return path
 }
 
-function getRouteBaseNameFactory (contextRoute) {
-  function givenOrCurrent (route) {
-    return route || this.$route
+function getRouteBaseName (route) {
+  route = this.route
```
paulgv: Should we remove the `route` argument from `getRouteBaseName`'s definition? It seems superfluous since we're assigning `route` here.

rchl: Yes! :)

PR #502, opened 2019-10-22
Translate doc to spanish

chefjuanpi:
Hi,
We translate the docs to spanish.


rchl:
Thank you. I appreciate the effort.

Just not sure how to handle cases when we update or add something in English docs. We would have to rely on you or the community to update translation too. And I can see this getting hard to synchronize...

chefjuanpi:
I not sure, if vuepress has some 'internationalization coverage control" or similar, because you are right the docs and the lib continue to evolve means today the docs can be 100% translate and up to day but tomorrow no.

I will search how others teams using vuepress manage this part.

chefjuanpi:
I create this issue in vuepress https://github.com/vuejs/vuepress/issues/1978 maybe it becomes interesting.

PR #507, opened 2019-10-25
refactor: avoid storing route in a property

rchl:
I didn't realize that fact when creating __route property, but the
context object stays the same during lifetime of the application and
the `route` property is updated to match current route so we can use
that instead of creating extra from middleware.


paulgv:
LGTM!

PR #510, opened 2019-10-28
fix: Direct navigation to URL in SPA with vue-router in hash mode

### LINKED ISSUE #490: page infinitely loading in spa mode
### Version

[v6.3.0](https://github.com/nuxt-i18n/releases/tag/v6.3.0)

### Reproduction link

[https://github.com/divisey/nuxt-i18n-path-issue](https://github.com/divisey/nuxt-i18n-path-issue)

### Steps to reproduce

yarn 

yarn dev

then visit http://localhost:3000/#/es/ in browser

### What is expected ?

it should work well when url is either http://localhost:3000/#/es, http://localhost:3000/#/es/ or http://localhost:3000/#/es/?a=1

it works when url is  http://localhost:3000/#/es, but page infinitely loading when you visit http://localhost:3000/#/es/, or when you have a 
query like http://localhost:3000/#/es?a=1

### Additional comments?

this issue does not appear in v6.0.0, maybe somewhere has a bug in the later version.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c322">#c322</a>)</em></sub></div>

rchl:
This fixes bug that triggers in quite specific situation - SPA mode
with Vue-router set to use hash mode.

In that setup if user navigates directly to an URL that will trigger
certain route and that route's `path` does not *exactly* match the
path user navigated to (for example due to extra trailing slash
or query parameter in user's URL), that triggers redirect that fails
due to target route being the same. That leaves current route in a
broken state. I assume because that route wasn't committed yet.

Fix by comparing routes in more sophisticated way than just comparing
route's path. Instead use VueRouter to resolve route from path and then
compare that with current route.

Resolves #490

### DIFF
--- a/src/plugins/main.js
+++ b/src/plugins/main.js
@@ -18,7 +18,13 @@ import {
   vueI18n,
   vuex
 } from './options'
-import { getLocaleDomain, getLocaleFromRoute, syncVuex, validateRouteParams } from './utils'
+import {
+  getLocaleDomain,
+  getLocaleFromRoute,
+  isSameRoute,
+  syncVuex,
+  validateRouteParams
+} from './utils'
 
 Vue.use(VueI18n)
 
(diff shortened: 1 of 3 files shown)


rchl:
The code to compare routes is taken from VueRouter (with some minor modifications). Unfortunately, VueRouter doesn't export those so we can't use them directly.

src/plugins/main.js
```diff
@@ -161,13 +167,13 @@ export default async (context) => {
     await syncVuex(store, newLocale, app.i18n.getLocaleMessage(newLocale))
 
     if (!initialSetup && strategy !== STRATEGIES.NO_PREFIX) {
+      const redirectPath = app.switchLocalePath(newLocale) || app.localePath('index', newLocale)
```
rchl: This one line replaces the previous two lines. Basically switched to using `switchLocalePath` as we can now use it from this context thanks to the other fix that made it work in plugins.

Issue #152, opened 2018-11-05
Dynamic routes can hijack more specific routes


rchl:
### Version

[v5.3.0](https://github.com/nuxt-i18n/releases/tag/v5.3.0)

### Reproduction link

[https://codesandbox.io/s/ly9pnqvp79](https://codesandbox.io/s/ly9pnqvp79)

### Steps to reproduce

1. Open codesandbox testcase
2. Add `/us` to the URL

### What is expected ?

Page shows 'US' in big letters

### What is actually happening?

Page shows 'FR'

### Additional comments?

This error happens with dynamic nuxt routes. Those are named with `_.vue` (or `_slug.vue` but I haven't checked those).

The bug seems to be in the order in which routes are generated. For example in this case:

```
['nuxt-i18n', {
            locales: [
                {code: 'pt', iso: 'pt-PT', file: 'pt.json'},
                {code: 'fr', iso: 'fr-FR', file: 'fr.json'},
                {code: 'us', iso: 'us-US', file: 'us.json'},
                // ...
            ],
            // ...
            defaultLocale: 'fr',
        }],
```

routes are generated in the order in which languages are defined in `locales` array. With this setting, generated routes will look like this:

```
routes: [
        {
                path: "/pt/*",
                component: _865a92ea,
                name: "all___pt"
        },
        {
                path: "/*",
                component: _865a92ea,
                name: "all___fr"
        },
        {
                path: "/us/*",
                component: _865a92ea,
                name: "all___us"
        }
],
```

As you can see, the 'catch-all' route (default language) was placed before more specific 'us' route. With this setup, 'us' route is inacessible due to routes being matched by iterating routes in the order they are defined.

Problem can be fixed by generating routes in proper order.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c168">#c168</a>)</em></sub></div>

rchl:
More mature implementations like ui-router for angular sort the rules by specificity. It's quite complex algorithm though and I'm not sure exactly about specifics.

rchl:
This is quite a gotcha so should be fixed.

pimlie:
Is case-insensitivity a must for the locale part of routes? If not, you could use similar routes as we do in nuxtpress: https://github.com/nuxt/press/blob/develop/packages/core/src/blueprint/index.js#L435-L440

[...]

PR #518, opened 2019-11-04
Add warning for wildcards with prefix_except_default strategy

SHoogland:
Let me know if this is the correct way of documenting this behaviour, but it took me a few hours to figure this out, and I hope I can prevent that wasted time for someone else.


rchl:
It's a valid and useful comment but I would like to fix #152 so that this is not necessary. I'll try to look into that soon.

rchl:
I've merged slighly altered variant of this warning at 56ae21a27bab9bc03dffbe7015cd85a843ce2d88. Thanks.

PR #544, opened 2019-11-27
Use Node 12.1.0 for testing instead of latest

rchl:
Trying to fix "Segmentation fault (core dumped)" error.


rchl:
It seems to have fixed the crash. There is also a separate issue with browser testing

PR #554, opened 2019-12-27
feat: support paths in localePath()

sugoidesune:
support passing in path as string to localePath()
`localePath('/app/profile')`

[Stackoverflow Conversation](https://stackoverflow.com/questions/58968237/nuxtjs-i18n-vue-router-route-with-name-about-us-en-does-not-exist/59497900#59497900)


rchl:
Thanks but I would require tests for this too.

sugoidesune:
> Thanks but I would require tests for this too.

Alright.  Found where the tests are. Will try and add a commit soon!

rchl:
Thanks!

PR #566, opened 2020-01-17
Feature/nuxt property decorator

moleCuleFFF:
This is to fix https://github.com/nuxt-community/nuxt-i18n/issues/564
But also addresses adding "In-component" options as a Class Property for better integration into nuxt-property-decorator

This is a ruff first draft so expect some conversation ;)


moleCuleFFF:
I'll adjust tests to increase coverage

husayt:
@moleCuleFFF let me know if anything can be done on [nuxt property decorator](https://github.com/nuxt-community/nuxt-property-decorator) side to help

PR #570, opened 2020-01-21
fix: remove substring to support localCode more than 2 digits

hitripod:
`localeCodes` is defined in `nuxt.config.js`, so we don't need to trim the locale code to 2 digits.
Otherwise, the detected language would be wrong if the locale code has more than 2 digit.

For example: `zh-TW` and `zh-CN` are totally two different charsets. If it's still limited to 2 digits, there is no way to distinguish from them.

### DIFF
--- a/src/plugins/main.js
+++ b/src/plugins/main.js
@@ -233,9 +233,9 @@ export default async (context) => {
         // Exclude 1 for backwards compatibility and fallback when fallbackLocale is empty
       } else if (process.client && typeof navigator !== 'undefined' && navigator.language) {
         // Get browser language either from navigator if running on client side, or from the headers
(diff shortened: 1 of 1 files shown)


rchl:
- This is a breaking change since it can break detection for people who currently use two-letter codes in their configuration
- It can and should be done but this is not the way to do it. It needs to be more complex logic that tries to match full code but fallbacks to two-letter code if not found. Then it could be introduced as non-breaking change.

MaySoMusician:
Is there any plan to accept this even in a future version?
I'm facing the same problem: separating `zh-CN` from `zh-TW`

rchl:
It's a bug and should be fixed but the fix needs to be done in a different way.

rchl:
I have made a fix that should address that in #651 (not yet integrated at the time of writing).

PR #578, opened 2020-01-27
chore(deps): migrate vue-i18n-loader package name

YoshinoriN:
Hello. Thank you for awesome package :)

The vue-i18n-loader package name & repo has been changed. Please see the following.

> https://github.com/intlify/vue-i18n-loader/pull/69/.

This PR update following.

* dependency
* docs
* loader in `index.js`

### DIFF
--- a/docs/es/vue-i18n-loader.md
+++ b/docs/es/vue-i18n-loader.md
@@ -1,6 +1,6 @@
 # vue-i18n-loader
 
-Si desea habilitar [vue-i18n-loader](https://github.com/kazupon/vue-i18n-loader), simplemente configure la opción `vueI18nLoader` en `true`.
+Si desea habilitar [vue-i18n-loader](https://github.com/intlify/vue-i18n-loader), simplemente configure la opción `vueI18nLoader` en `true`.
 
 ```js
 // nuxt.config.js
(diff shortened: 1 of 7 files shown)


rchl:
Thank you for this PR.
Unfortunately, this needs to be a major change since some might have configured vue-i18n-loader using those options https://nuxt-community.github.io/nuxt-i18n/vue-i18n-loader.html#yaml
So I'm gonna merge this when I feel like it's a good time to release a new major version.

YoshinoriN:
@rchl 
Thank you for your review :)
Please let me know when you merge this, I will rebase this PR to master before you merge.

rchl:
Actually I was thinking we should handle migration automatically. From the `extendBuild` hook we can check `config.module.rules` and modify instances of `@kazupon/vue-i18n-loader` with `@intlify/vue-i18n-loader`. We just have to remember that `loader` can be either a string or array. Then we can release it as minor change.

Would you be able to do that?

YoshinoriN:
OK. I will try it out this weekend :)

rchl:
I've looked into it.

I don't think we need migration for it as documentation example was outdated anyway and using it that way would have not worked anyway. vue-i18n-loader has had [support for yaml for a while now](https://github.com/intlify/vue-i18n-loader/commit/36c62a5) and it's supposed to be used with `lang="yaml"` attribute now.

I've updated documentation and added a test for yaml block in 23856656d55e69e43a029d0789cd05d3b1247464. You can cherry pick that commit (it's in main repo) and then I'll merge it.

YoshinoriN:
Sorry, so the late reply.
I cherry-pick  https://github.com/nuxt-community/nuxt-i18n/commit/23856656d55e69e43a029d0789cd05d3b1247464 and force-pushed from the current master branch.

rchl:
Thank you :)

PR #593, opened 2020-02-14
fix: use the synced locale on client if set during SSR

matthieusieben:
If `vuex.syncLocale` is true, then any locale determined on the server during SSR should be used as initial value by the client.

This also allows:
- Not to determine the language twice when a page is loaded (once on the server, then again in the client)
- Other plugins (e.g. user session in a universal app) to alter the locale during bootstrap (`await app.i18n.setLocale('xx')`) and make sure that that locale is re-used on the client.

Note that I think that this should be default behavior for this module.

### DIFF
--- a/src/plugins/main.js
+++ b/src/plugins/main.js
@@ -206,7 +206,9 @@ export default async (context) => {
 
   let locale = app.i18n.defaultLocale || null
 
-  if (app.i18n.differentDomains) {
+  if (vuex && vuex.syncLocale && store && store.state[vuex.moduleName].locale !== '') {
+    locale = store.state[vuex.moduleName].locale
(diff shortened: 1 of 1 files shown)


rchl:
I think it makes sense but would need some tests for it too

matthieusieben:
I have no idea how to test that. Can you help ?

rchl:
Can help when I find some time.

rchl:
Sorry for holding it off for so long.
I haven't really find a good way to test it (to assert that locale from store is used on the client), but since tests run with `syncLocale` enabled, it should be fine already. I'll only add some tests with vuex disabled to make sure that that is tested.

PR #597, opened 2020-02-16
feat: additional catchall hreflang tags

manniL:
Resolves https://github.com/nuxt-community/nuxt-i18n/issues/522

### DIFF
--- a/docs/seo.md
+++ b/docs/seo.md
@@ -30,7 +30,7 @@ An example without selected catchall locale:
       iso: 'en-US' // Will be used as catchall locale by default
     },
     {
-      code: 'engb',
+      code: 'gb',
       iso: 'en-GB'
     }
   ]
@@ -48,7 +48,7 @@ Here is how you'd use `isCatchallLocale` to selected another language:
       iso: 'en-US'
     },
     {
-      code: 'engb',
+      code: 'gb',
       iso: 'en-GB',
       isCatchallLocale: true // This one will be used as catchall locale
     }
@@ -64,7 +64,7 @@ In case you already have an `en` language iso set, it'll be used as the catchall
 ['nuxt-i18n', {
   locales: [
     {
-      code: 'engb',
+      code: 'gb',
       iso: 'en-GB'
     },
     {
@@ -121,7 +121,7 @@ You must also set the `baseUrl` option to your production domain in order to mak
 
 
 To enable this feature everywhere in your app, set `seo` option to `true`. 
-**This comes with a performance drawback though**. More infos below.
+**This comes with a performance drawback though**. More information below.
 
 ```js
 // nuxt.config.js
--- a/docs/seo.md
+++ b/docs/seo.md
@@ -1,15 +1,92 @@
 # SEO
 
-With `seo` option enabled, **nuxt-i18n** attempts to add some metadata to improve your pages SEO. Here's what it does:
 
-* Add a _lang_ attribute containing current locale's ISO code to the `<html>` tag.
-* Generate `<link rel="alternate" hreflang="x">` tags for every language configured in `nuxt.config.js`. For each language, the ISO code is used as `hreflang` attribute's value. [More on hreflang](https://support.google.com/webmasters/answer/189077)
-* Generate `og:locale` and `og:locale:alternate` meta tags as defined in the [Open Graph protocol](http://ogp.me/#optional)
-* When using `prefix_and_default` strategy, generate `rel="canonical"` link on the default language routes containing the
+## Benefits
+
+When the `seo` option is enabled, **nuxt-i18n** attempts to add some metadata to improve your pages SEO. Here's what it does.
+
+### `lang` attribute for `<html>` tag.
+
+With `seo` enabled, nuxt-i18n will set the correct `lang` attribute, equivalent to the current locale's ISO code, in the `<html>` tag.
+
+### Automatic hreflang generation
+
+nuxt-i18n will generate `<link rel="alternate" hreflang="x">` tags for every language configured in `nuxt.config.js`. 
+The language's ISO codes are used as `hreflang` values. 
+
+Since version [v6.6.0](https://github.com/nuxt-community/nuxt-i18n/releases/tag/v6.6.0), a catchall locale hreflang link
+is provided for each language group (e.g. `en-*`) as well. 
+By default, it is the first language provided but another language can be selected by setting `isCatchallLocale` to `true`
+on that specific language object in your `nuxt.config.js`. [More on hreflang](https://support.google.com/webmasters/answer/189077)
+
+An example without selected catchall locale:
+```js
+// nuxt.config.js
+
+['nuxt-i18n', {
+  locales: [
+    {
+      code: 'en',
+      iso: 'en-US' // Will be used as catchall locale by default
+    },
+    {
+      code: 'engb',
+      iso: 'en-GB'
+    }
+  ]
+}]
+```
+
+Here is how you'd use `isCatchallLocale` to selected another language:
+```js
+// nuxt.config.js
+
+['nuxt-i18n', {
+  locales: [
+    {
+      code: 'en',
+      iso: 'en-US'
+    },
+    {
+      code: 'engb',
+      iso: 'en-GB',
+      isCatchallLocale: true // This one will be used as catchall locale
+    }
+  ]
+}]
+```
+
+In case you already have an `en` language iso set, it'll be used as the catchall without doing anything
+
+```js
+// nuxt.config.js
+
+['nuxt-i18n', {
+  locales: [
+    {
+      code: 'engb',
+      iso: 'en-GB'
+    },
+    {
+      code: 'en',
+      iso: 'en' // will be used as catchall locale
+    }
+  ]
+}]
+```
+
+### OpenGraph Locale tag generation
+
+nuxt-i18n will also generate `og:locale` and `og:locale:alternate` meta tags as defined in the [Open Graph protocol](http://ogp.me/#optional).
+
+### Canonical links for `prefix_and_default`
+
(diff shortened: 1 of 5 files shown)


rchl:
I thought the idea was to add that more-generic meta tag *in addition to* original one, rather than replace it.

Also, I would like this to be more automatic. I was thinking to include the list of all locales in the module and derive one automatically (from first locale-country code seen, for example). Of course, it would still be possible to override that with an extra property.

As for the property name, maybe call it something closer to [documentation](https://support.google.com/webmasters/answer/189077) - for example `isCatchallLocale`.

And the new property should be documented in `seo.md` and `options-reference.md`.

manniL:
> I thought the idea was to add that more-generic meta tag _in addition to_ original one, rather than replace it.


I don't think it's intended to use the same link/url for a specific language *and* a catch-all. At least I can't find an example (not in the guidelines nor on 3rd party SEO-related pages like Moz, yoast, ...) reflecting that.

> Also, I would like this to be more automatic. I was thinking to include the list of all locales in the module and derive one automatically (from first locale-country code seen, for example). Of course, it would still be possible to override that with an extra property.

Can't be automated easily in case the first point is valid.

> As for the property name, maybe call it something closer to [documentation](https://support.google.com/webmasters/answer/189077) - for example `isCatchallLocale`.

Will change ☺️

> And the new property should be documented in `seo.md` and `options-reference.md`.

True!

rchl:
> I don't think it's intended to use the same link/url for a specific language _and_ a catch-all. At least I can't find an example (not in the guidelines nor on 3rd party SEO-related pages like Moz, yoast, ...) reflecting that.

How about that part (see emphasis)?
> If you have several alternate URLs targeted at users with the same language but in different locales, it's a good idea also to provide a catchall URL for geographically unspecified users of that language. For example, you may have specific URLs for English speakers in Ireland (en-ie), Canada (en-ca), and Australia (en-au), but should also provide a generic English (en) page for searchers in, say, the US, UK, and all other English-speaking locations. **It can be one of the specific pages, if you choose.**
https://support.google.com/webmasters/answer/189077

manniL:
@rchl Yup, you are right! Already correct my psot above after seeing another confirming tweet from John Mueller ☺️

test/module.test.js
```diff
@@ -49,6 +49,18 @@ describe('basic', () => {
...
+        tagName: 'link',
+        rel: 'alternate',
+        href: 'nuxt-app.localhost/',
+        hreflang: 'en'
+      },
+      {
+        tagName: 'link',
+        rel: 'alternate',
+        href: 'nuxt-app.localhost/fr',
+        hreflang: 'fr'
```
rchl: I'd like to also have a test where locale iso code is initially two letter locale. I believe this is still a valid ISO code and people might be using such configuration already. In that case, no extra meta tags should be added for that locale.

So configuration like:
```
locales: [{ code: 'en', iso: 'en' }, { code 'fr', iso: 'fr-FR' }]
```

src/templates/seo-head.js
```diff
@@ -30,22 +30,46 @@ export const nuxtI18nSeo = function () {
...
+
+      const forcedCatchallLocale = localesWithLanguage.find(locale => locale.isCatchallLocale) || localesWithLanguage[0]
+      const catchAllLocaleIso = forcedCatchallLocale[LOCALE_ISO_KEY]
+      const catchAllLanguage = languageFromLocaleIso(catchAllLocaleIso)
+      const catchallLocale = { ...locales.find(locale => locale.hreflang === catchAllLocaleIso) }
+      catchallLocale.hid = `alternate-hreflang-${catchAllLanguage}`
+      catchallLocale.hreflang = catchAllLanguage
+
+      return catchallLocale
+    }).filter(Boolean)
```
rchl: Find this code kinda hard to read with all the maps, filters, finds and multiple uses of `languageFromLocaleIso`.

I've come up with this instead which I think works as well (at least tests think so):
```js
  if (strategy !== STRATEGIES.NO_PREFIX) {
    const localeMap = new Map()

    for (const locale of this.$i18n.locales) {
      const localeIso = locale[LOCALE_ISO_KEY]

      if (!localeIso) {
        // eslint-disable-next-line no-console
        console.warn(`[${MODULE_NAME}] Locale ISO code is required to generate alternate link`)
        continue
      }

      const [language, region] = localeIso.split('-')

      if (language && region && (locale.isCatchallLocale || !localeMap.has(language))) {
        localeMap.set(language, locale)
      }

      localeMap.set(localeIso, locale)
    }

    for (const [iso, locale] of localeMap.entries()) {
      link.push({
        hid: `alternate-hreflang-${iso}`,
        rel: 'alternate',
        href: baseUrl + this.switchLocalePath(locale.code),
        hreflang: iso
      })
    }
  }
```

Also, we could think about moving this generation to the module's code as it needs to be done only once really. The `switchLocalePath` would still have to be done at runtime though.

manniL: Really like this code and added it together with my other refactoring in the latest commits.

Regarding extracting the majority of the setup (where possible) to the build time part - this would totally make sense, as long as the complexity of the runtime part with `switchLocalePath` won't be increased ☺️

test/module.test.js
manniL: Will cover that in a test case. Should I add a new fixture or alter the current one?

src/templates/seo-head.js
manniL: We should also consider that `Map` might need a polyfill, so moving that part to runtime could also remove the need for that ☺️

src/templates/seo-head.js
```diff
@@ -21,71 +21,127 @@ export const nuxtI18nSeo = function () {
...
-    if (canonicalPath && canonicalPath !== this.$route.path) {
-      // Current page is not the canonical one -- add a canonical link
-      link.push({
-        hid: 'canonical-lang-' + currentLocaleData[LOCALE_CODE_KEY],
-        rel: 'canonical',
-        href: baseUrl + canonicalPath
-      })
-    }
-  }
+  addHreflangLinks.bind(this)(currentLocale, this.$i18n.locales, link)
```
manniL: I tried to restructure the code and went with functions + explicit binding instead of arrow functions.

I've also thought about returning a part of the `link`/`meta` array and eventually concatenating them instead of altering/pushing directly

Would love to get your feedback on that. ☺️

test/module.test.js
manniL: Covered it in the same fixture for now

src/templates/seo-head.js
rchl: It looks fine. :)

test/module.test.js
rchl: You don't have to create new fixture but you can override settings for the existing one. Like in `differentDomains enabled` test.

test/module.test.js
```diff
@@ -42,13 +42,25 @@ describe('basic', () => {
...
         tagName: 'link',
         rel: 'alternate',
         href: 'nuxt-app.localhost/fr',
         hreflang: 'fr-FR'
+      },
+      {
+        tagName: 'link',
+        rel: 'alternate',
+        href: 'nuxt-app.localhost/',
+        hreflang: 'en'
```
rchl: So now there is a duplicate tag for 'en'? If so, it sounds like a bug.

rchl:
We should also have test that uses `IsCatchallLocale`.

test/module.test.js
manniL: Whoops, should just be a false positive because the two objects are the same. Will update the test and assure it reflects the correct changes

manniL: Was an FP ☺️

manniL:
> We should also have test that uses `IsCatchallLocale`.

Covered in a separate fixture! ☺️

> Also a friendly reminder about documentation. :)

Done now as well, thanks for the reminder 👍🏻

docs/seo.md
```diff
@@ -1,15 +1,92 @@
...
+
+In case you already have an `en` language iso set, it'll be used as the catchall without doing anything
+
+```js
+// nuxt.config.js
+
+['nuxt-i18n', {
+  locales: [
+    {
+      code: 'engb',
```
rchl: I feel it should just be `gb` code. I'm not saying it's wrong but `gb` is kinda more standard. And when using with date formatting, that value is actually used as a parameter to `Intl.DateTimeFormat` so it can make a difference (set side whether it's correct to pass that value actually).

docs/seo.md
```diff
@@ -43,7 +120,8 @@ You should also set the `baseUrl` option to your production domain in order to m
 ```
 
 
-To enable this feature everywhere in your app, set `seo` option to `true`:
+To enable this feature everywhere in your app, set `seo` option to `true`. 
+**This comes with a performance drawback though**. More infos below.
```
rchl: Just "info"? :)

test/module.test.js
```diff
@@ -320,6 +330,113 @@ describe('basic', () => {
   })
 })
 
+describe('hreflang', () => {
+  let nuxt
+
+  beforeAll(async () => {
+    const override = {
```
rchl: This might in the end give same result but merging these locales with base config ends up with duplicate locales:

```
[ { code: 'en', iso: 'en', name: 'English' },                                                                                                                               
         { code: 'fr', iso: 'fr-FR', name: 'Français' },
         { code: 'en', iso: 'en', name: 'English' },
         { code: 'fr', iso: 'fr-FR', name: 'Français' },
         { code: 'es', iso: 'es-ES', name: 'Spanish (Spain)' },
         { code: 'esVe',
           iso: 'es-VE',
           name: 'Spanish (Venezuela)',
           isCatchallLocale: true } ]
```

because deepmerge will merge array of locales. I'd prefer to not have duplicates in the first place by just loading original config and then overriding locales manually. Like in `lazy loading` testsuite for example.

manniL: Totally makes sense! Thanks for the heads up ☺️

Issue #555, opened 2020-01-02
Is there a way to set cookie's domain


bayun2:
### What problem does this feature solve?

For example, my site is a.xxx.com,  i want to set cookie in root domain(xxx.com) for other site like b.xxx.com can get the same cookie 

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c360">#c360</a>)</em></sub></div>

thnima:
@rchl Is there any update on this?
We actually have the same situation: The app is hosted on a different subdomain than the website. To share the cookie we would need to set the cookie domain to the root domain.
As this is very important: Is there any workaround to achieve this behaviour right now?

Thank you for your efforts on this repo!

Best,
Thomas

rchl:
It would be easy to add that functionality.
Unfortunately, I'm quite busy right now, but maybe I'll find some time this week but anyone is welcome to contribute.

thnima:
Setting a `cookieDomain` property looks pretty easy indeed. I will have a look at it.

rchl:
This was implemented by looked PR

PR #599, opened 2020-02-18
feat: Added option to exclude subdomain from cookie domain

thnima:
According to #555

### DIFF
--- a/docs/es/options-reference.md
+++ b/docs/es/options-reference.md
@@ -62,6 +62,10 @@ Aquí están todas las opciones disponibles al configurar el módulo y sus valor
     // preferred language to prevent subsequent redirections
     // Set to false to redirect every time
     useCookie: true,
+    // Enabled by default, which means the cookie domain includes the subdomain
+    // (e.g. "www.mydomain.com")
+    // If disabled, the subdomain is excluded (e.g. "mydomain.com")
+    includeCookieSubdomain: true
     // Cookie name
     cookieKey: 'i18n_redirected',
     // Set to always redirect to value stored in the cookie, not just once
--- a/docs/es/options-reference.md
+++ b/docs/es/options-reference.md
@@ -62,7 +62,7 @@ Aquí están todas las opciones disponibles al configurar el módulo y sus valor
     // preferred language to prevent subsequent redirections
     // Set to false to redirect every time
     useCookie: true,
-    // Specifies the domain of the cookie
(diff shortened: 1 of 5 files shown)


docs/options-reference.md
```diff
@@ -62,6 +62,10 @@ Here are all the options available when configuring the module and their default
     // preferred language to prevent subsequent redirections
     // Set to false to redirect every time
     useCookie: true,
+    // Enabled by default, which means the cookie domain includes the subdomain
+    // (e.g. "www.mydomain.com")
+    // If disabled, the subdomain is excluded (e.g. "mydomain.com")
+    includeCookieSubdomain: true
```
rchl: I'm not sure about the naming since what would it mean in the case when the page is on `mydomain.com`? Then one would think that this option is irrelevant while it's not the case - toggling it to false would still change behavior as the behavior of cookie with explicit domain set is different from cookie without explicit domain set.

I think I would prefer to have option `cookieDomain` that could be set to an explicit value and that would 1:1 reflect the value passed as cookie option.

Also this solution is not as flexible as explicit domain would be. User might be running sites on `a.foo.bar.com` and `b.foo.bar.com` and want to set cookie domain to `foo.bar.com`.

thnima: I initially had the same thought, but decided against it, as the cookieDomain can only be the actual host. But I see the argument with domains like `a.foo.bar.com`. I'll adapt it to use the `cookieDomain` property

types/nuxt-i18n.d.ts
```diff
@@ -26,6 +26,7 @@ declare namespace NuxtVueI18n {
 
     interface DetectBrowserLanguageInterface {
       useCookie?: boolean
+      cookieDomain?: string
```
rchl: `string | null` since with strict null checking the `?` means undefined so it wouldn't match default value.

docs/es/options-reference.md
```diff
@@ -62,6 +62,8 @@ Aquí están todas las opciones disponibles al configurar el módulo y sus valor
     // preferred language to prevent subsequent redirections
     // Set to false to redirect every time
     useCookie: true,
+    // Specifies the domain of the cookie
+    cookieDomain: null
```
rchl: Maybe:
```
Set to override the default domain of the cookie. Defaults to host of the site.
```

rchl:
Thank you

PR #600, opened 2020-02-19
chore: switch to `@nuxtjs/eslint-config-typescript` for eslint configuration

rchl:
Also fix some types to be visible as methods rather than properties.


kevinmarrec:
👌

PR #622, opened 2020-03-03
docs: Update branch-language-detection

progvb:
Added description where the language code comes from.


rchl:
I think it would be better to put it after first configuration block. And it could be a 
```
::: tip
text
:::
```
block

progvb:
@rchl Thank you for fast review. I've updated the PR as you suggested.

rchl:
Thank you for PR :)

PR #623, opened 2020-03-06
add prefix if no domain defined for locale.

chrissyast:
allows locale code prefix to be added if `differentLocales` is true but locale does not have `domain` defined.


rchl:
- What would be the use case for this change?
- There are two other places that handle locale and do an "early return" of sort when `differentDomains` is enabled. I would expect those places to be sort of broken with this change.
- We would need to add at least one test for this. That would also help in understanding the use case.

chrissyast:
Apologies, this PR was sent to your main repo instead of our fork. Full documentation and tests will be provided in due course

chrissyast:
@rchl I believe I am at a stage where I'm ready to submit a proper PR for this feature now. Could you help me with how to configure the tests?

PR #627, opened 2020-03-11
Get X-Forwarded-Port

maxm86545:
Get X-Forwarded-Port for use SSR with different port

### DIFF
--- a/src/templates/utils.js
+++ b/src/templates/utils.js
@@ -135,10 +135,10 @@ export function isSameRoute (a, b) {
  * @return {String} x-forwarded-host & x-forwarded-port
  */
 const getForwarded = function (req) {
-    if (process.client) {
-        return window.location.href.split('/')[2]
-    }
-    return (req.headers['x-forwarded-host'] ? req.headers['x-forwarded-host'] : req.headers.host) +
+  if (process.client) {
+    return window.location.href.split('/')[2]
+  }
+  return (req.headers['x-forwarded-host'] ? req.headers['x-forwarded-host'] : req.headers.host) +
       (req.headers['x-forwarded-port'] ? (':' + req.headers['x-forwarded-port']) : '')
 }
 
--- a/src/templates/utils.js
+++ b/src/templates/utils.js
@@ -130,13 +130,17 @@ export function isSameRoute (a, b) {
 }
 
 /**
- * Get x-forwarded-host
+ * Get x-forwarded-host & x-forwarded-port
  * @param  {object} req
- * @return {String} x-forwarded-host
+ * @return {String} x-forwarded-host & x-forwarded-port
  */
-const getForwarded = req => (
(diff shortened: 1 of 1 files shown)


src/templates/utils.js
```diff
@@ -130,13 +130,17 @@ export function isSameRoute (a, b) {
...
  */
-const getForwarded = req => (
-  process.client ? window.location.href.split('/')[2] : (req.headers['x-forwarded-host'] ? req.headers['x-forwarded-host'] : req.headers.host)
-)
+const getForwarded = function (req) {
+  if (process.client) {
+    return window.location.href.split('/')[2]
+  }
+  return (req.headers['x-forwarded-host'] ? req.headers['x-forwarded-host'] : req.headers.host) +
+      (req.headers['x-forwarded-port'] ? (':' + req.headers['x-forwarded-port']) : '')
```
rchl: I'd suggest this to make things more readable and "dry":
```suggestion
  const xForwardedHost = req.headers['x-forwarded-host']
  const xForwardedPort = req.headers['x-forwarded-port']
  return (xForwardedHost ? xForwardedHost : req.headers.host) + (xForwardedPort ? (':' + xForwardedPort) : '')
```
But I'll also take some time to think a bit about the logic.

rchl: From the logic's perspective, this is a breaking change since while before one would get `foo.com`, now one will likely get something like `foo.com:443` (for HTTPS request). So then that value will no longer match `domain` configured in `locales` object.

I understand that you want to use different locale depending on the port used, right?

I suppose we can make this a non-breaking change by adding a logic where if configured `LOCALE_DOMAIN_KEY` doesn't contain a colon, we match only host part from `getForwarded`, otherwise we match both.

rchl: Another thing to consider is that this will make behavior between client and server even more inconsistent.

For example:
 - On the client the URL is `https://foo.com/page` and so code returns `foo.com` host.
 - On the server we compute `foo.com:443` from `x-forwarded-*` headers for HTTPS or `foo.com:80` for HTTP
 - client and server don't match and domain matching only works for one case.

That issue is due to browser no including standard ports in the URL.

maxm86545: > I understand that you want to use different locale depending on the port used, right?

No, I use same port for different locales, but I use different domains. I use proxy via Nginx with different port. I am having problems with SSR, locale is undefined on server side.

rchl: Not quite sure I understand why the port matters.
I would imagine you have Nginx configured so that only one of the ports goes to your Nuxt app. Why would you need to be so specific about the port then?

maxm86545: > ignore port when matching locale domain but use the port specified by the user in `domain` (if any) when redirecting

Maybe this is the best solution?
https://github.com/nuxt-community/nuxt-i18n/pull/630

rchl: Yes, we can move into that.

PR #630, opened 2020-03-12
Remove forwardedHost parameter

maxm86545:
Remove forwardedHost for simplify. Ignore port when matching locale domain (it's a solution for working SSR: https://github.com/nuxt-community/nuxt-i18n/pull/627#discussion_r391029596).

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -114,12 +114,6 @@ Instance of [VueI18n class](http://kazupon.github.io/vue-i18n/api/#vuei18n-class
 
   Whether `differentDomains` option is enabled.
 
-#### forwardedHost
-
-  - **Type**: `boolean`
-
-  Whether `forwardedHost` option is enabled.
-
 #### beforeLanguageSwitch
 
   - **Type**: `Function`
--- a/docs/es/api/README.md
+++ b/docs/es/api/README.md
@@ -114,12 +114,6 @@ Instance of [VueI18n class](http://kazupon.github.io/vue-i18n/api/#vuei18n-class
 
   Whether `differentDomains` option is enabled.
 
-#### forwardedHost
-
-  - **Type**: `boolean`
-
-  Whether `forwardedHost` option is enabled.
-
 #### beforeLanguageSwitch
 
   - **Type**: `Function`
--- a/docs/es/options-reference.md
+++ b/docs/es/options-reference.md
@@ -77,10 +77,6 @@ Aquí están todas las opciones disponibles al configurar el módulo y sus valor
   // as an array of objects, each containing a domain key
   differentDomains: false,
 
-  // If using different domains, set this to true to get hostname from X-Forwared-Host
-  // HTTP header instead of window.location
-  forwardedHost: false,
-
   // If true, SEO metadata is generated for routes that have i18n enabled.
   // Note that performance can suffer with this enabled and there might be compatibility
   // issues with some plugins. Recommended way is to set up SEO as described in:
--- a/docs/options-reference.md
+++ b/docs/options-reference.md
@@ -90,10 +90,6 @@ Here are all the options available when configuring the module and their default
   // as an array of objects, each containing a domain key
   differentDomains: false,
 
-  // If using different domains, set this to true to get hostname from X-Forwared-Host
-  // HTTP header instead of window.location
-  forwardedHost: false,
-
   // If true, SEO metadata is generated for routes that have i18n enabled.
   // Note that performance can suffer with this enabled and there might be compatibility
   // issues with some plugins. Recommended way is to set up SEO as described in:
--- a/src/helpers/constants.js
+++ b/src/helpers/constants.js
@@ -41,7 +41,6 @@ exports.DEFAULT_OPTIONS = {
     fallbackLocale: null
   },
   differentDomains: false,
-  forwardedHost: false,
   seo: false,
   baseUrl: '',
   vuex: {
--- a/src/plugins/main.js
+++ b/src/plugins/main.js
@@ -8,7 +8,6 @@ import {
   defaultLocale,
   detectBrowserLanguage,
   differentDomains,
-  forwardedHost,
   lazy,
   localeCodes,
   locales,
@@ -188,7 +187,6 @@ export default async (context) => {
   app.i18n.locales = locales
   app.i18n.defaultLocale = defaultLocale
   app.i18n.differentDomains = differentDomains
-  app.i18n.forwardedHost = forwardedHost
   app.i18n.beforeLanguageSwitch = beforeLanguageSwitch
   app.i18n.onLanguageSwitched = onLanguageSwitched
   app.i18n.setLocaleCookie = setLocaleCookie
--- a/src/templates/utils.js
+++ b/src/templates/utils.js
@@ -130,31 +130,13 @@ export function isSameRoute (a, b) {
 }
 
 /**
- * Get x-forwarded-host
- * @param  {object} req
- * @return {String} x-forwarded-host
- */
-const getForwarded = req => (
-  process.client ? window.location.href.split('/')[2] : (req.headers['x-forwarded-host'] ? req.headers['x-forwarded-host'] : req.headers.host)
-)
-
-/**
- * Get hostname
- * @param  {object} req
- * @return {String} Hostname
- */
-const getHostname = req => (
-  process.client ? window.location.href.split('/')[2] : req.headers.host
-)
-
-/**
  * Get locale code that corresponds to current hostname
  * @param  {VueI18n} nuxtI18n Instance of VueI18n
  * @param  {object} req Request object
  * @return {String} Locade code found if any
  */
 export const getLocaleDomain = (nuxtI18n, req) => {
-  const hostname = nuxtI18n.forwardedHost ? getForwarded(req) : getHostname(req)
+  const hostname = process.client ? window.location.hostname : (req.headers['x-forwarded-host'] || req.headers.host)
   if (hostname) {
     const localeDomain = nuxtI18n.locales.find(l => l[LOCALE_DOMAIN_KEY] === hostname)
     if (localeDomain) {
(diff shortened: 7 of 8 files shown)


rchl:
This change has the potential to break someone's use case so I wonder if it should be a breaking change...

I guess it will only change behavior when someone has `x-forwarded-host` header but didn't want to use it before. But that would likely result in broken functionality with `differentDomains` so I will probably just treat it as a bug fix.

I'm not sure how does it fix your case though. It seems equivalent to the old code when it comes to the server-side.
For client-side there will be a change again, as `location.href.split('/')[2]` returned port (didn't notice that before) while `location.hostname` won't.

So before merging this, I would like to understand your case. Are your public URLs include the port (for example https://foo.com:888 ?). How would this fix help here for `differentDomains` functionality?

maxm86545:
> I guess it will only change behavior when someone has x-forwarded-host header but didn't want to use it before.

Someone doesn't want to use x-forwarded-host header? Does this person exist?

> For client-side there will be a change again, as location.href.split('/')[2] returned port (didn't notice that before) while location.hostname won't.

Yes, `location.href.split('/')[2]` return port on client-side, and `req.headers['x-forwarded-host']` or `req.headers.host` don't return port on server-side. I did the same behavior on client and server.

> So before merging this, I would like to understand your case. Do your public URLs include the port (for example https://foo.com:888 ?). How would this fix help here for differentDomains functionality?

Yes, my public URLs include the port. For example: foo.com:888 - English, fr.foo.com:888 - French, es.foo.com:888 - Spanish.

rchl:
Still, don't understand how this helps. You still won't pass the port when redirecting.

maxm86545:
I specify domains in the nuxt.config.js without ports:

> i18n: {
        locales: [
            {
                name: 'English',
                domain: 'foo.com',
                ...
            },
            {
                name: 'French',
                domain: 'fr.foo.com',
                ...
            },
            {
                name: 'Spanish',
                domain: 'es.foo.com',
                ...
            },
        ],
         ...
}

rchl:
And so nuxt-i18n will redirect without port included and it won't work as far as I can see.

maxm86545:
This solution works for me. For me the port is not important, because it does not change locale. Only different domains change locale.

maxm86545:
> And so nuxt-i18n will redirect without port included and it won't work as far as I can see.

Do you mean switchLocalePath redirect without port?

rchl:
No, i mean different domains functionality that will still not work properly.
Or maybe tell me what was the issue without this fix? Exactly. How did it behave and when

maxm86545:
> Or maybe tell me what was the issue without this fix? Exactly. How did it behave and when

Ок.
My public URLs include the port. For example: foo.com:888 - English, es.foo.com:888 - Spanish.

My nuxt.config.js is:
> i18n: {
        locales: [
            {
                name: 'English',
                code: 'en',
                iso: 'en',
                domain: 'foo.com:888',
            },
            {
                name: 'Spanish',
                code: 'es',
                iso: 'es',
                domain: 'es.foo.com:888',
            },
        ],
        differentDomains: true,
        forwardedHost: true, // I tried using true and false
         ...
}

I use Nginx proxy:
foo.com:888 -> (with headers `Host: foo.com` and  `X-Forwarded-Host: foo.com`) node.js (nuxt) server
es.foo.com:888 ->  (with headers `Host: es.foo.com` and  `X-Forwarded-Host: es.foo.com`) node.js (nuxt) server

I have problems with translation on server-side, SSR is not working. Translation works on client-side.

rchl:
Well, no you are saying something else than before (previously you said you don't have ports in configuration).

But anyways, even with the fix I still don't see how it will work. The `getLocaleDomain` will return null (as `domain` setting and `x-forwarded-host` won't match) and locale won't match the domain.

maxm86545:
With this fix I specify domains in the `nuxt.config.js` without ports. The `getLocaleDomain` return correct locale domain.

rchl:
This change changes nothing for that case, as far as I can see. At least for the server-side requests.

With this configuration:
```
i18n: {
        locales: [
            {
                name: 'English',
                code: 'en',
                iso: 'en',
                domain: 'foo.com:888'
            },
            {
                name: 'Spanish',
                code: 'es',
                iso: 'es',
                domain: 'es.foo.com:888'
            },
        ],
        differentDomains: true,
        forwardedHost: true, // I tried using true and false
         ...
}
```
Both before this change and after, the `x-forwarded-host` would not batch `domain` and the locale wouldn't be applied according to domain.

Please, give me full details instead of partial responses. I feel like we are running in circles.

rchl:
And the same if you don't include ports in Nuxt configuration. The behavior should still be the same both before and after this change.

While I don't mind this change (I think it simplifies things), I don't understand how to fixed your issue. I want to understand that as it would be good to also have test for that.

maxm86545:
Your comment: https://github.com/nuxt-community/nuxt-i18n/pull/630#issuecomment-598053277
> location.href.split('/')[2] returned port (didn't notice that before) while location.hostname won't.

https://github.com/nuxt-community/nuxt-i18n/blob/eeb63bb701d1b76804790e87ef8cabe9149fdb9a/src/templates/utils.js#L157
On client-side the `nuxtI18n.forwardedHost ? getForwarded(req) : getHostname(req)` return host with port, and on server-side return host without port. `const hostname` on client-side not equals on server-side. This creates a problem with SSR. My fix solves this problem. And I find the forwardedHost parameter redundant, so I suggest removing it.

rchl:
OK.
I would think that creates problem on the client (SPA) and not server (SSR) but OK, at least I understand your problem now.

maxm86545:
Additionally, maybe we can add the key `port` for locales? For example:

```
locales: [
     {
          name: 'English',
          code: 'en',
          iso: 'en',
          domain: 'foo.com',
          port: 888
    },
    {
          name: 'Spanish',
          code: 'es',
          iso: 'es',
          domain: 'es.foo.com:888',
          port: 888
    },
    ...
]
```

This nee

rchl:
I've also added some tests in 3f4d13577bbca02e3b8d4bdd156a7854a95c7f0d and added deprecation warning for that setting.

jensgeffken:
Is this warning only temporary ? Till the next iteration ? Because i have this warning in an universal ssr nuxt app since over 2 weeks and i cant figure the instructions given to resolve this. Thanks

rchl:
@jensgeffken Just don't set `forwardedHost` in options passed to the module.

jensgeffken:
> @jensgeffken Just don't set `forwardedHost` in options passed to the module.

you mean in nuxt.config.js ?

I did this ...

['nuxt-i18n', {
      detectBrowserLanguage: {
        useCookie: true,
        cookieKey: 'language_redirect',
        alwaysRedirect: true,
        fallbackLocale: 'en'
      },
      defaultLocale: 'de',
      baseUrl: process.env.NODE_ENV === 'development' ? 'http://localhost:8888' : 'https://www.example.com',
      seo: false,
      locales: [
        {
          code: 'en',
          iso: 'en-US',
          file: 'en-US.js',
          name: 'English'
        },
        {
          code: 'de',
          iso: 'de-DE',
          file: 'de-DE.js',
          name: 'Deutsch'
        }
      ],
      lazy: true,
      langDir: 'languages/'
    }]

still the warning ... dang

rchl:
Which warning are you talking about? With that config, you wouldn't get any warning about `forwardedHost`.

jensgeffken:
> Which warning are you talking about? With that config, you wouldn't get any warning about `forwardedHost`.

mmh nope still getting the error in universal ssr app without forwardedHost or forwardedHost set to false in nuxt.config.js

rchl:
Please copy-paste exact erro.

jensgeffken:
> Please copy-paste the exact warning you are getting.

 WARN  [undefined] The forwardedHost option is deprecated. You can safely remove it. See: https://github.com/nuxt-community/nuxt-i18n/pull/630.

My config is like the one i showed, running: nuxt (v2.12.1) universal ssr, in dev mode, when starting the app with npm run dev, i ll get the mentioned warning ...

rchl:
The warning is harmless but I don't see how it would trigger unless you have some weird issues with build files (like module's code partially being new and partially old).

I'd recommend deleting `node_modules`, deleting `.nuxt` build directory and then reinstalling dependencies and starting next again.

If you have an issue with the warning then that might be sign of some deeper issue that could cause other problems so I recommend you try what I wrote above.

jensgeffken:
ok, thanks a lot

sowinski:
Hi guys,

updated nuxt-i18n from version 6.4.1 to 6.11.1 and I am using x-forwarded-host.

Nuxt-i18n is currently not picking up the correct translation. I do not understand what I have to change. Can someone explain it? Or is this somewhere documented?

rchl:
There is no option anymore because `x-forwarded-host` is always taken into consideration now.

If something doesn't work then your best bet is to create an issue and provide more information in it.

jensgeffken:
the message still appears 👎 
` WARN  [undefined] The forwardedHost option is deprecated. You can safely remove it. See: https://github.com/nuxt-community/nuxt-i18n/pull/630.`
guess its one of those things, i will ignore it, everything else works fine

rchl:
@jensgeffken This message triggers when there is a `forwardedHost` key in options. And the module itself doesn't add that key at any place in the code. So it's coming from your project somehow.

You can look at generated file in `.nuxt/nuxt-i18n/options.js` to check what the options you are passing to the module include. If you have `forwardedHost` there then it's coming from somewhere in your project.

jensgeffken:
> @jensgeffken This message triggers when there is a `forwardedHost` key in options. And the module itself doesn't add that key at any place in the code. So it's coming from your project somehow.
> 
> You can look at generated file in `.nuxt/nuxt-i18n/options.js` to check what the options you are passing to the module include. If you have `forwardedHost` there then it's coming from somewhere in your project.

hello @rchl thanks for the reply, i found it in my config and successfully removed it. My mistake thanks again.

Issue #245, opened 2019-03-11
Add ability to alter status code with rootRedirect option.


elisiondan:
### What problem does this feature solve?

By default, the redirect function sets the HTTP response code to **302 Found**. However, there are situations when it is desirable to set a different status. Specifically, this change avoids the issue with google crawlers which mark the root page and the target page of the *rootRedirect* property as duplicates. This occurs because the crawler expects one of the pages to marked as canonical. In case when the situation is solved with a redirect, the crawler [expects](https://support.google.com/webmasters/answer/139066?hl=en) status code 301. This simple change improves the website SEO when using rootRedirect.

### What does the proposed changes look like?

I attach the updated code and a provided example in the documentation [here](https://github.com/elisiondesign/nuxt-i18n/commit/868b96d60aa46f56d1a3d9a857203fc36a8fab17).

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c212">#c212</a>)</em></sub></div>

gaelreyrol:
This issue is real :disappointed:

gaelreyrol:
Does his suggestion suitable ? Do you need a PR @rchl ?

rchl:
I think it makes sense. PR welcome.

rchl:
Resolved by #639

PR #639, opened 2020-03-24
feat: allow customization of rootRedirect status code

gaelreyrol:
Hi,

This PR address the feature request from #245.

Hope it suits !

Regards

### DIFF
--- a/src/templates/middleware.js
+++ b/src/templates/middleware.js
@@ -1,5 +1,5 @@
 import middleware from '../middleware'
-import options from './options'
+import { detectBrowserLanguage, rootRedirect } from './options'
 import { getLocaleFromRoute } from './utils'
 
 middleware.nuxti18n = async (context) => {
@@ -10,18 +10,20 @@ middleware.nuxti18n = async (context) => {
   }
 
   // Handle root path redirect
-  const rootRedirect = <%= JSON.stringify(options.rootRedirect) %>
   if (route.path === '/' && rootRedirect) {
-    if (rootRedirect.path && rootRedirect.statusCode) {
-      redirect(rootRedirect.statusCode, '/' + rootRedirect.path, route.query)
-      return
+    let statusCode = 302
+    let path = rootRedirect
+
+    if (typeof rootRedirect !== 'string' && rootRedirect.path && rootRedirect.statusCode) {
+      statusCode = rootRedirect.statusCode
+      path = rootRedirect.path
     }
 
-    redirect('/' + rootRedirect, route.query)
+    redirect(statusCode, '/' + path, route.query)
     return
   }
 
-  if (options.detectBrowserLanguage && await app.i18n.__detectBrowserLanguage()) {
+  if (detectBrowserLanguage && await app.i18n.__detectBrowserLanguage()) {
     return
   }
 
--- a/src/templates/middleware.js
+++ b/src/templates/middleware.js
@@ -14,7 +14,7 @@ middleware.nuxti18n = async (context) => {
     let statusCode = 302
     let path = rootRedirect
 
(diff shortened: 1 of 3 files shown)


src/templates/middleware.js
```diff
@@ -1,5 +1,5 @@
 import middleware from '../middleware'
-import { detectBrowserLanguage, rootRedirect } from './options'
+import options from './options'
```
rchl: Please only import things you are using (as before).

src/templates/middleware.js
```diff
@@ -10,12 +10,18 @@ middleware.nuxti18n = async (context) => {
   }
 
   // Handle root path redirect
+  const rootRedirect = <%= JSON.stringify(options.rootRedirect) %>
```
rchl: You don't need to use template. Just plain JS will work.

src/templates/middleware.js
```diff
@@ -10,12 +10,18 @@ middleware.nuxti18n = async (context) => {
...
   if (route.path === '/' && rootRedirect) {
+    if (rootRedirect.path && rootRedirect.statusCode) {
+      redirect(rootRedirect.statusCode, '/' + rootRedirect.path, route.query)
+      return
+    }
+
```
rchl: Maybe refactor so that you don't have to repeat `redirect` call twice:

```js
let statusCode = 302
let path = rootRedirect

if (typeof rootRedirect !== 'string') {
  { statusCode, path } = rootRedirect
}

redirect(statusCode, '/' + path, route.query)
```

src/templates/middleware.js
```diff
@@ -11,7 +11,15 @@ middleware.nuxti18n = async (context) => {
 
   // Handle root path redirect
   if (route.path === '/' && rootRedirect) {
-    redirect('/' + rootRedirect, route.query)
+    let statusCode = 302
+    let path = rootRedirect
+
+    if (typeof rootRedirect !== 'string' && rootRedirect.path && rootRedirect.statusCode) {
```
rchl: What's the purpose of checking that properties are set? If they are not, we're gonna pass `rootRedirect` object as path and given that it's not a string, it's not gonna work anyway.

src/templates/middleware.js
```diff
@@ -11,7 +11,15 @@ middleware.nuxti18n = async (context) => {
...
+      statusCode = rootRedirect.statusCode
+      path = rootRedirect.path
```
rchl: Is there a specific reason for my suggested, shorter way? Just wondering.

gaelreyrol: I get an error from the linter: `Declaration or statement expected.`

src/templates/middleware.js
gaelreyrol: Well I did this because of my implementation in the condition block

rchl: We can't recover from a situation when `rootRedirect` is not a string and those properties are missing so we are as good (or bad) without those checks.

gaelreyrol: If rootRedirect is an object and does not contains both statusCode and path it will throw an undefined error. I though we might not want to throw but just fallback on the default behavior. But if you think it is unnecessary I'll change it :)

rchl: It will not throw, it will set both `statusCode` and `path` to `undefined`.
And what's the other option anyway, setting `path` to empty object? That won't work either.

rchl:
Thank you :)
(I'll add some test for it before releasing)

gaelreyrol:
@rchl You released a new version with this feature but it has not been published on npm

rchl:
Indeed. Fixed now.
Thanks for noticing.

PR #651, opened 2020-03-31
feature: match against full language code provided by the browser

### LINKED ISSUE #563: Translations doesn't work when switching language with disabled useCookie property
### Version

[v6.4.1](https://github.com/nuxt-community/nuxt-i18n/releases/tag/v6.4.1)

### Reproduction link

[https://codesandbox.io/s/thirsty-lake-m7vwr](https://codesandbox.io/s/thirsty-lake-m7vwr)

### Steps to reproduce

Click on the link with the label: "Deutsch"

### What is expected ?

The string "English string" should get translated to "Deutscher string"

Clicking the link changes the route to "/de", which is fine. But the text doesn't get translated.
If you set `useCookie: true` in the nuxtConfig again, it works as expected.
Same behaviour can be observed, when you set `useCookie: true` AND disable the cookies in the browser

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c370">#c370</a>)</em></sub></div>

### LINKED ISSUE #650: Using 2-Letter Country/Region Code instead of Language Code in detectBrowserLanguage

When there is a need to distinguish Chinese-Simplified and Chinese-Traditional, we will be happy to use country / region codes like CN or TW or HK or MO or SG as locale "codes" in our applications. Currently, main.js sees only the first 2-letter language code, for instance, 'zh' for entire Chinese locales that are apparently different one another. We all need to know that end-users in these areas are quite sensitive to distinguish their locales and that they dislike ambiguity in their mother tongues.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c430">#c430</a>)</em></sub></div>

rchl:
For detecting and redirecting to browser language the module only
looked at first two letters of the language codes provided by the
browser so it wasn't able to match up more specific locale codes like
en-GB or zh-CN.

Fix by matching checking against full language code first and then
falling back to only first two letters if no higher-ranked code
was found.

Resolves #650
Resolves #563

### DIFF
--- a/docs/browser-language-detection.md
+++ b/docs/browser-language-detection.md
@@ -6,6 +6,7 @@ By default, **nuxt-i18n** attempts to redirect users to their preferred language
 // nuxt.config.js
 
 ['nuxt-i18n', {
+  // ...
   detectBrowserLanguage: {
     useCookie: true,
     cookieKey: 'i18n_redirected'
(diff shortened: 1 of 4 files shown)


mcdmaster:
Looks fine. Can you please go ahead with this PR before long? Thank you!

mcdmaster:
Thank you!

PR #652, opened 2020-04-01
Update main.js

mcdmaster:
For enabling use of country/region codes instead of language codes in locales[].code so that ambiguous locales can be distinguished e.g. zh-cn vs. zh-tw


rchl:
See #651

PR #654, opened 2020-04-03
Hotfix: compare locales with suffix as case-insensitive

mcdmaster:
Comparison of locale codes in complete fashion i.e. ISO-631 Language Code and ISO-3166 Country / Region Code concatenated with a hyphen in between needs to be done as case-insensitive.

Not only have many web developers but also we ourselves been figured out that they used to use "complete" language codes in entirely lower case letters like zh-cn, zh-tw etc.

### DIFF
--- a/src/templates/utils-common.js
+++ b/src/templates/utils-common.js
@@ -24,7 +24,7 @@ export const matchBrowserLocale = (appLocales, browserLocales) => {
 
   // First pass: match exact locale.
   for (const [index, browserCode] of browserLocales.entries()) {
-    if (appLocales.includes(browserCode)) {
+    if (appLocales.toLowerCase().includes(browserCode.toLowerCase())) {
       matchedLocales.push({ code: browserCode, score: 1 - index / browserLocales.length })
       break
     }


mcdmaster:
Related issue: tokyo-metropolitan-gov/covid19#2765

src/templates/utils-common.js
```diff
@@ -24,8 +24,14 @@ export const matchBrowserLocale = (appLocales, browserLocales) => {
...
   for (const [index, browserCode] of browserLocales.entries()) {
-    if (appLocales.includes(browserCode)) {
-      matchedLocales.push({ code: browserCode, score: 1 - index / browserLocales.length })
+    if (appLocales.forEach(appLocale => {
+      if (appLocale.toLowerCase().includes(browserCode.toLowerCase())) {
+        matchedLocales.push({ code: browserCode, score: 1 - index / browserLocales.length })
+        return true
+      } else {
+        continue
+      }
```
rchl, line 33: You need to push app locale since browser locale case might not match and then that will not match in the calling code.

Also here is suggested cleaner code that doesn't use functions (forEach):
```suggestion
    for (const appCode of appLocales) {
      if (appCode.toLowerCase() === browserCode.toLowerCase()) {
        matchedLocales.push({ code: appCode, score: 1 - index / browserLocales.length })
        break;
      }
```

rchl: Also please add a test for comparing locales in case-insensitive manner (in `unit.test.js`).

src/templates/utils-common.js
```diff
@@ -24,9 +24,11 @@ export const matchBrowserLocale = (appLocales, browserLocales) => {
...
   for (const [index, browserCode] of browserLocales.entries()) {
-    if (appLocales.includes(browserCode)) {
-      matchedLocales.push({ code: browserCode, score: 1 - index / browserLocales.length })
-      break
+    for (const appCode of appLocales) {
+      if (appCode.toLowerCase() === browserCode.toLowerCase()) {
+        matchedLocales.push({ code: appCode, score: 1 - index / browserLocales.length })
+        break
+      }
     }
```
rchl, line 32: Actually we should break out of the outer loop so maybe instead:

```suggestion
    const matchedCode = appLocales.find(appCode => appCode.toLowerCase() === browserCode.toLowerCase())
    if (matchedCode) {
      matchedLocales.push({ code: matchedCode, score: 1 - index / browserLocales.length })
      break
    }
```

src/templates/utils-common.js
mcdmaster: Thank you! I much appreciate your helping.

test/unit.test.js
```diff
@@ -34,21 +34,36 @@ describe('matchBrowserLocale', () => {
     const { matchBrowserLocale } = await import('../src/templates/utils-common')
 
     // Both locales match first browser locale - full locale should win.
-    const appLocales = ['en', 'en-US']
+    const appLocales = ['en', 'en-US', 'en-us']
     const browserLocales = ['en-US', 'en']
 
     expect(matchBrowserLocale(appLocales, browserLocales)).toBe('en-US')
+
+    const appLocales = ['zh-CN', 'zh-cn']
```
rchl, line 42: Please create separate test blocks for tests related to case instead of mixing with existing tests.

mcdmaster: Got it. Thanks

test/unit.test.js
```diff
@@ -34,21 +34,36 @@ describe('matchBrowserLocale', () => {
...
 
     // Both locales match first browser locale - full locale should win.
-    const appLocales = ['en', 'en-US']
+    const appLocales = ['en', 'en-US', 'en-us']
     const browserLocales = ['en-US', 'en']
 
     expect(matchBrowserLocale(appLocales, browserLocales)).toBe('en-US')
+
+    appLocales = ['zh-CN', 'zh-cn']
+    browserLocales = ['zh-CN', 'zh']
```
rchl, line 43: We probably only need one new test that tests that full code is compared case-insensitively.
Eventually, we could also test a case-insensitive match on short (two-letter) locale.
No need to create pointless tests that don't test anything specific.

mcdmaster: Intended to go as far as I can. Thanks anyway

test/unit.test.js
```diff
@@ -34,18 +34,18 @@ describe('matchBrowserLocale', () => {
     const { matchBrowserLocale } = await import('../src/templates/utils-common')
 
     // Both locales match first browser locale - full locale should win.
-    appLocales = ['en', 'en-US', 'en-us']
-    browserLocales = ['en-US', 'en']
+    string[] appLocales = ['en', 'en-US', 'en-us']
+    string[] browserLocales = ['en-US', 'en']
```
rchl, line 38: Sorry, but the changes you are making make no sense. You don't seem to have an understanding of javascript. Maybe you want me to just take over that fix?

mcdmaster: Haha, indeed I was a programmer decades ago but I am doing it to fight against COVID-19!

mcdmaster: If you take over the change, I much appreciate it. Thank you very much!

MaySoMusician: @mcdmaster He might want you to leave the tests that have already been as is, and to create just new tests about case-insentive comparison

rchl:
I'll take over in few hours when I have free time

test/unit.test.js
mcdmaster: @rchl Can you please be handed over to create testing?

@MaySoMusician I think that this implementation needs deep dive into existing behavior,, which should not be spoiled. I am convinced that @rchl is the most aware of what `nuxt-i18n` should be with our request implemented. Hence, I wanted to hand this over to him. Thank you anyway

mcdmaster:
> I'll take over in few hours when I have free time

Thank you @rchl, it helps us fight against COVID-19 in Japan but we are having multilingual, multinational and multicultural neighbors. I would like to meet up with you when we can overcome this panic situation around the globe! Thanks again

rchl:
Resolved with e614e8e31010a7a0114cd9ea9d01588d7c26e510

mcdmaster:
Thank you very much!

mcdmaster:
Hi @rchl,

Do you mind being added to our [Contributors' List](https://github.com/tokyo-metropolitan-gov/covid19/blob/master/CONTRIBUTORS.md)?
For sure, I would like to recommend our team to do so.
Can you please let me hear your opinion? Thanks in advance

rchl:
I don't mind. Go ahead. :)

PR #664, opened 2020-04-13
fix: allow detecting locale from path, when it fails from name

badihi:
I suggest removing `else` from getLocaleFromRoute to allow detecting locale from path, when it is not possible to be detected from name.

I will be useful when we want to have some pages, dedicated to a locale, like what is already available in basic test fixture (page under `pages/fr/`). In this case, the locale will be detected from url, and in case that the page uses some components, they will use the appropriate locale.

### DIFF
--- a/src/templates/utils.js
+++ b/src/templates/utils.js
@@ -163,7 +163,8 @@ export const getLocaleFromRoute = (route = {}) => {
     if (matches && matches.length > 1) {
       return matches[1]
     }
-  } else if (route.path) {
+  }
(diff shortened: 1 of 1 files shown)


rchl:
I think it makes sense but I would like to also see test that fails without this change and passes with.

Also documentation in `routing.md` is not entirely correct by saying:
```
To completely disable i18n on a given page:

export default {
  nuxtI18n: false
}
```

We could say instead something like `To completely disable generation of locale-specific routes for given page:`.

PR #665, opened 2020-04-13
WIP: feature:  support for pages for an specific locale

badihi:
I propose a new feature, to enable users to have fully customized pages for specific locales. This can be very useful for content pages, that using translation tools will not be beneficial for them, say FAQ or about-us.

Note that using some tricks, like creating a directory for the locale (like what is done in basic fixture) will not always work, especially in scenarios including nested pages.

Using this feature, localized page 'sth.vue' can be overridden for a specific locale 'xx', using the file 'sth.xx.vue' near the original file.

### DIFF
--- a/src/helpers/routes.js
+++ b/src/helpers/routes.js
@@ -44,6 +44,17 @@ exports.makeRoutes = (baseRoutes, {
       ...pageOptions,
       ...routeOptions
     }
+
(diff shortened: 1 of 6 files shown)


rchl:
Hmm, I suppose I'm not entirely against it but does it work correctly with [dynamic routes](https://nuxtjs.org/guide/routing#dynamic-routes) like `_foo.vue` or `_.vue`? I guess not?

PR #684, opened 2020-04-24
chore(docs): add example for wildcard route configuration

kavrillon:
As I spent some times to understand how to customize i18n urls for wildcard pages (`_.vue`), I suggest an example in the documentation to illustrate it.

### DIFF
--- a/docs/es/routing.md
+++ b/docs/es/routing.md
@@ -151,9 +151,10 @@ pages/
 ├── _nested/
 ├──── _route/
 ├────── index.vue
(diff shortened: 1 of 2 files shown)


rchl:
Could you also update ES docs? And I don't expect you to know Spanish, I would like you to just replace changes sentences with update English ones. Someone who knows Spanish will have to update them later.

rchl:
Awesome. Thank you.

kavrillon:
@rchl You're welcome!

Issue #689, opened 2020-04-27
Enabling "fallbackLocale" in "detectBrowserLanguage" intereferes with static page generation by breaking prerendering


Louvki:
### Version
nuxt-i18n: 6.10.1
nuxt: 2.12.2

### Reproduction Link
https://codesandbox.io/s/romantic-neumann-y1bd1?file=/nuxt.config.js

### Steps to reproduce
I want to prerender 6 pages for the route of "pages\gen\_id\index.vue".
In the page I console.log the payload once it's received in the async data method.

1. Run "npm generate" 
2. Check the console. You will see it only logged out the payload for pages with the default locale. ("en/gen/1", "en/gen/2"). Meaning only 2 pages of the initial 6 will be pre-rendered.

1. Now comment out the "fallbackLocale" option in the i18n object in nuxt.config.js
2. Run "npm generate" 
3. Check the console. Now it will have logged out the payload for every page

### What is Expected?
"fallbackLocale" option should not interefere with static page generation.

### What is actually happening?
When "fallbackLocale" option is enabled, pages which are not of default locale do not get pre-rendered.

rchl:
I can see this happening as the module will redirect to default locale if there is no cookie set yet and I'm assuming that during generation cookies are not sent or saved. This will probably also happen with `alwaysRedirect: true`.

It probably makes sense to special-case generation of pages. It would just need to be reliably detectable from a plugin. I will have a look at some point.

rchl:
Resolved by #718 (spinned of from #715)

PR #715, opened 2020-05-18
fix: ignore browser-detection by nuxt generate

ThornWalli:
This commit tries to fix issue [#689](https://github.com/nuxt-community/nuxt-i18n/issues/689).

With `nuxt generate` the browser detection is ignored.
Language of the route will not be changed and routes can be rendered.

### DIFF
--- a/src/templates/plugin.main.js
+++ b/src/templates/plugin.main.js
@@ -127,7 +127,7 @@ export default async (context) => {
 
   const doDetectBrowserLanguage = () => {
     // Browser detection is ignored if it is a nuxt generate.
-    if (context.isStatic && context.ssrContext) {
+    if (process.static && process.server) {
       return false
     }
 
--- a/src/templates/plugin.main.js
+++ b/src/templates/plugin.main.js
@@ -126,6 +126,11 @@ export default async (context) => {
   }
 
   const doDetectBrowserLanguage = () => {
+    // Browser detection is ignored if it is a nuxt generate.
+    if (context.isStatic && context.ssrContext) {
+      return false
+    }
+
     const { alwaysRedirect, fallbackLocale } = detectBrowserLanguage
 
     let matchedLocale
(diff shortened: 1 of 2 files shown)


src/templates/plugin.main.js
```diff
@@ -126,6 +126,11 @@ export default async (context) => {
   }
 
   const doDetectBrowserLanguage = () => {
+    // Browser detection is ignored if it is a nuxt generate.
+    if (context.isStatic && context.ssrContext) {
```
rchl, line 130: These are deprecated ways of checking generated/server mode. See https://nuxtjs.org/guide/plugins#client-side-only for suggested ones.

ThornWalli: I updated the check. Hope this is enough. 

Added a test that checks for file content. 
A direct call to '/fr/index.html' is necessary for the check.
Unfortunately I could not do this with the test browser.

rchl: > Unfortunately I could not do this with the test browser.

What's the reason you couldn't do it? There is already test using `generate` there so I wonder why made it not feasible.

rchl: OK, I see now. Because we are testing pre-rendered content.

ThornWalli: With the existing test there is no initial call of the non-default locale route.

If I open the /fr/index.html directly with a browser and would notice that it is empty.

With the test browser this case is unfortunately not to be adjusted, here the internal router seems to cover it.

It seems that if a fallback locale is defined in detectBrowserLanguage, the rendering of non-default local routes generates empty pages.

rchl:
I'm gonna integrate this through https://github.com/nuxt-community/nuxt-i18n/pull/718 as I've done some small changes to tests and added one more.

Also noticed that generate + detectBrowserLanguage doesn't work well currently (on the client) as the code that redirects doesn't run on initial loading. This is due to `middleware` not running for generated pages on initial load.
I will have to fix that separately.

rchl:
Thanks.

PR #724, opened 2020-05-21
fix: trigger language detection on initial load in generated mode

rchl:
With Nuxt generate & universal mode Nuxt doesn't trigger middleware
on initial loading so we need to do it manually so that possible
redirection is handled.


maxerbox:
Waiting for the fix to release, it is causing weird redirection in production

PR #731, opened 2020-05-25
fix: remove "encodePaths" option and don't encode paths by default

### LINKED ISSUE #712: How can I redirect translated url to the same page
I want to translate url, but have an issue in ssr loading.  
here is my configuration `
```
// nuxt.config.js
module.exports = {
...
modules: [
   [
      'nuxt-i18n',
      {
        seo: true,
        lazy: true,
        detectBrowserLanguage: {
          useCookie: true,
          cookieKey: 'i18n_redirected',
          alwaysRedirect: false,
          fallbackLocale: 'hy'
        },
        locales: [
          {
            name: 'English',
            code: 'en',
            iso: 'en-US',
            file: 'en.json'
          },
          {
            name: 'Հայերեն',
            code: 'hy',
            iso: 'hy-AM',
            file: 'hy.json'
          }
        ],
        langDir: 'locales/',
        defaultLocale: 'hy',
        parsePages: false,
         pages: {
           about: {
             en: '/about',
             hy: '/մեր֊մասին'
           },
       }
   ]
}
```
it works in client side when i link to localePath('about') . but when i refresh my page it cant find  http://localhost:4000/մեր֊մասին

rchl:
This could be considered a breaking change but since I only know of it
breaking things rather than helping anything and there are no tests that
would prove me otherwise, I think it's safe to remove without releasing
a major version.

Also, Nuxt doesn't encode paths either so this change aligns with it.

Resolves #712

### DIFF
--- a/docs/es/routing.md
+++ b/docs/es/routing.md
@@ -229,19 +229,6 @@ Debería configurar su propiedad `pages` de la siguiente manera:
 
 Si falta una ruta personalizada para una de las configuraciones locales, se usa la ruta personalizada `defaultLocale` si está establecida..
 
-### Expresión regular
-
-Por defecto, todas las rutas personalizadas están codificadas para manejar caracteres no latinos en la ruta. Esto convertirá rutas con expresiones regulares como `/foo/:slug-:id(\\d+)` a `/foo/:slug-:id(%5Cd+)`.
-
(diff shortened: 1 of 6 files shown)


rchl:
Paths encoding was fixing issue #7 but that one doesn't reproduce in latest Nuxt anymore so I'm even more confident in getting rid of `encodePaths` option.

PR #747, opened 2020-06-01
Add missing closing brace

seishun:


rchl:
Can you also do the same change in `es` docs?

rchl:
Pushed manually with added fix in `es`. Thanks.

Issue #557, opened 2020-01-08
this.$i18n.locales undefined when specifying i18n options through component option


coin-au-carre:
### Version

[v5.3.0](https://github.com/nuxt-i18n/releases/tag/v5.3.0)

### Reproduction link

[https://nuxt-community.github.io/nuxt-i18n/lang-switcher.html](https://nuxt-community.github.io/nuxt-i18n/lang-switcher.html)

### Steps to reproduce

Follow the steps on the first section of [Lang Switcher](https://nuxt-community.github.io/nuxt-i18n/lang-switcher.html#lang-switcher)

### What is expected ?

this.$i18n.locales should contain an array of 3 objects which are defined in nuxt.js.config. 

### What is actually happening?

`this.$i18n.locales` is undefined 

```
TypeError
Cannot read property 'filter' of undefined
```

### Additional comments?

this.$i18n.availableLocales returns an array `['en','es','fr]`

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c365">#c365</a>)</em></sub></div>

rchl:
It works in my test sandbox: https://codesandbox.io/s/i18n-plugin-i18nlocales-f82pn

(Just remember that you have to navigate the `/en` path, for example, as there is no index page generated for default `prefix` strategy.)

coin-au-carre:
Ok thank you this is weird if I do this in your code :

```
 availableLocales() {
      console.log(this.$i18n)
      return this.$i18n.locales.filter(i => i.code !== this.$i18n.locale);
    }
```

I can indeed see that `this.$i18n.locales` is well defined. Something is preventing that on my side...

coin-au-carre:
Ok I think I got the problems. Putting a pages/test.vue files works. However if I put this code in layouts/default.vue, it does not work. My vue knowledge is newbish I guess I try to do things I am not supposed to do in layouts/ ?

rchl:
If you can reproduce this problem in codesandbox then I can have a look.

Holi0317:
Codesandbox sample: https://codesandbox.io/s/i18n-plugin-i18nlocales-it7n5

The `$i18n` property of current component will get override without `nuxt-i18n` augmentation if there is an `<i18n>` tag exist in the SFC with `vueI18nLoader` enabled in `nuxt.config.js`

[...]

PR #748, opened 2020-06-01
Fix switchLocalePath

freewhite4:
DifferentDomains is undefined.
I fixed the switchLocalePath function. Passing the correct i18n properties.

### DIFF
--- a/src/templates/plugin.routing.js
+++ b/src/templates/plugin.routing.js
@@ -156,7 +156,7 @@ const VueInstanceProxy = function (targetFunction) {
   return function () {
     const proxy = {
       getRouteBaseName: this.getRouteBaseName,
-      i18n: this.$i18n,
+      i18n: this.$router.app.$i18n,
       localePath: this.localePath,
       req: process.server ? this.$ssrContext.req : null,
       route: this.$route,


freewhite4:
Fix switchLocalePath

src/templates/plugin.routing.js
```diff
@@ -156,7 +156,7 @@ const VueInstanceProxy = function (targetFunction) {
   return function () {
     const proxy = {
       getRouteBaseName: this.getRouteBaseName,
-      i18n: this.$i18n,
+      i18n: this.$router.app.$i18n,
```
rchl, line 159: This doesn't make sense to me. Why would we need to go through `$router` to get $i18n`. These two are unrelated.

And obviously `this.$i18n` is not undefined in general case or most of the tests would fail.
Please provide reproduction steps (or repo) so that I can check what is *actually* wrong there.

maxm86545: We work with @freewhite4 on one project. Our nuxt.config.js is https://github.com/nuxt-community/nuxt-i18n/pull/630#issuecomment-598068821. We have a problem: switchLocalePath returns a link without a domain. @freewhite4 fixed the problem.

rchl: I need to understand why it fixes the problem. This is surely not the way to fix it, whatever it is. Please create a minimal repo that reproduces.

rchl: I can see how this change can make some sort of difference since `this.$router.app.$i18n` is a root `$i18n` instance (more direct path to it would be `this.$root.$i18n`) while `this.$i18n` can be a component-local instance if component defines some `i18n` options.

But it's not correct to use root instance always as it might not reflect the local state so this fix seems wrong.

The very latest version (https://github.com/nuxt-community/nuxt-i18n/releases/tag/v6.12.1) includes a fix for #557 that might actually fix your problem too. Have you tried that latest version?

freewhite4: We used 6.11.1 version and had this problem. We migrated to 6.12.1 and the problem was resolved. 
Thank you!

Issue #717, opened 2020-05-19
trailingSlash option in router does impact i18n root page


miteyema:
### Version
nuxt-i18n: 6.11.1
nuxt: 2.12.2

#### [mode](https://nuxtjs.org/api/configuration-mode): 
 - [x] universal
 - [ ] spa

### Reproduction Link
https://github.com/miteyema/nuxt-i18n-demo

### Steps to reproduce
Configure trailingSlash in router to `false`

### What is Expected?
When I click on any other language, root page url should not have trailing slash, e.g. https://nuxt-i18n-demo.netlify.app/fr

### What is actually happening?
When I click on any other language, root page url contains trailing slash, e.g. https://nuxt-i18n-demo.netlify.app/fr/

rchl:
I've released a beta version with the fix. You can give it a go: `npm i nuxt-i18n@beta` if you want.

simplenotezy:
@rchl I just tested this with the beta; I can confirm it still happens with:

    <nuxt-link :to="localePath('/')">

Turns into:

    /da/

Live test-case with the beta version from tonight: https://thejewelleryroom.com

simplenotezy:
But it seems to work on some links (e.g. the footer link), but not the header logo that takes you to the frontpage.

simplenotezy:
I made a temporary workaround:

```
Vue.prototype.$removeTrailingSlash = (string) => {
	if (typeof string === 'string' && string.substr(-1) === '/' && string !== '/') {
		string = string.substr(0, (string.length - 1));
	}

	return string;
}
```

And then:

	<nuxt-link :to="$removeTrailingSlash(localePath('/'))">
		<svg>
	</nuxt-link>

rchl:
It's always better to use route names in `localePath` rather than paths. That works in more cases and in cases like this more correctly (kinda regret now that support for paths was added recently).

So if you use `localePath('index')` then it should work as expected.

I'll have a look if I can fix paths handling also.

simplenotezy:
I see. Thanks a lot Rafał! @rchl

rchl:
I wonder how would your server handle a file structure like so this one (I suppose it might be like that already):

```
| - da/
| - da/foo.html
| - da.html
| - index.html
```

And user visits `/da`.

Would it then load the root `da.html` file?
Because various node servers that I'm using for testing would kinda fail on it as they see the `da` directory and try to load `index.html` from it but since there is none they return 404. They don't fall back to the file in that case.

[...]

PR #756, opened 2020-06-05
fix: handling of trailing slash in localePath

rchl:
Respect "router.trailingSlash" option when using "localePath" with
an absolute path as an argument. The trailing slash is now normalized
according to that option (true: add trailing slash,
false or undefined: remove trailing slash). That excludes the root path
which always includes trailing slash.

An extra fix for #717

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
   "scripts": {
     "dev:basic": "nuxt -c ./test/fixture/basic/nuxt.config.js",
     "dev:basic:generate": "nuxt generate -c ./test/fixture/basic/nuxt.config.js",
+    "start:dist": "node -r esm ./test/utils/http-server-internal.js --port 8080 -v dist",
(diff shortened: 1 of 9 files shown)


rchl:
To go forward with this, I need to figure out if this change makes sense.

Since I have problems making tests work with a simple HTTP server, it might also be a problem for production servers like Apache or nginx.

The problem is when Nuxt generates such directory structure:
```
| - da/
| - da/foo.html
| - da.html
| - index.html
```

This can happen with those options set:
```
  router: {
    trailingSlash: false
  },
  generate: {
    subFolders: false,
  },
```

Then, supposedly, `http://localhost/da` should load the root `da.html` file and `http://localhost/foo.html` should load the `da/foo.html` file but at least the simple server would fail here as it would see the directory and since directory does not contain `index.html`, it would return 404.

PR #763, opened 2020-06-11
add acorn-walk

vodniciarv:
issue https://github.com/nuxt-community/nuxt-i18n/issues/137 is back in version 6.12.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "@babel/traverse": "^7.5.5",
     "@intlify/vue-i18n-extensions": "^1.0.1",
     "@intlify/vue-i18n-loader": "^1.0.0",
+    "acorn-walk": "^7.1.1",
     "cookie": "^0.4.0",
     "is-https": "^2.0.0",
     "js-cookie": "^2.2.1",


rchl:
What in this project uses `acorn-walk`?

Looking at the lock file, apart from jsdom which is not included in production, also `@nuxt/types` uses `acorn-walk` indirectly through `webpack-bundle-analyzer`.

Given that, I don't think it's the responsibility of this package to include `acorn-walk` but rather `@nuxt/types`' (although it would be a bit weird for types package to include such things). Maybe @kevinmarrec has an opinion.

rchl:
No, sorry, checked wrong. The dependency tree looks like this:

nuxt => @nuxt/webpack => webpack-bundle-analyzer => acorn-walk

So this package itself shouldn't depend on `acorn-walk`, I believe.

rchl:
Well, I guess I shouldn't be checking what explicitly depends on it but rather what uses it without explicitly depending on it.

So I did:
```
rm -rf node_modules
yarn --production
```
And search for acron-walk in all files in `node_modules` and didn't see anything using `acorn-walk`.

So please explain the exact error you are getting and why you think it's this package that is missing dependency.

rchl:
I'll reopen if you provide a valid reason for this.

PR #775, opened 2020-06-24
fix: Allow locale in path override browser detected one (#455)

stlhrt:
After stumbling on #455 this is quick solution. Doesn't break existing functionality and works as an opt-in feature.

### DIFF
--- a/src/helpers/constants.js
+++ b/src/helpers/constants.js
@@ -35,7 +35,8 @@ exports.DEFAULT_OPTIONS = {
     cookieDomain: null,
     cookieKey: 'i18n_redirected',
     alwaysRedirect: false,
-    fallbackLocale: ''
+    fallbackLocale: '',
+    routeOverride: false
   },
   differentDomains: false,
   seo: false,
--- a/src/templates/plugin.main.js
+++ b/src/templates/plugin.main.js
@@ -126,17 +126,19 @@ export default async (context) => {
     return redirectPath
   }
 
-  const doDetectBrowserLanguage = () => {
+  const doDetectBrowserLanguage = (routeLocale) => {
     // Browser detection is ignored if it is a nuxt generate.
     if (process.static && process.server) {
       return false
     }
 
-    const { alwaysRedirect, fallbackLocale } = detectBrowserLanguage
+    const { alwaysRedirect, fallbackLocale, routeOverride } = detectBrowserLanguage
 
     let matchedLocale
 
-    if (useCookie && (matchedLocale = app.i18n.getLocaleCookie())) {
+    if (routeOverride && routeLocale) {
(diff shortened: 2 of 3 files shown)


rchl:
What's the difference between this new option and `detectBrowserLocale: false`?

In any case, this solution is very hacky and doesn't consider many cases (differentDomains, server-detected locale, stored locale cookie, ...) so it won't be accepted as is. (And why is a locale variable called `routePath`?)

Also, I'm not willing to accept new options unless absolutely necessary as that logarithmically increases the number of setting combinations and that makes all possible combinations hard to test and keep working.

The original issue can IMO be fixed without introducing new options as it should only give benefits but this is not the way to do it.

stlhrt:
`detectBrowserLocale: false` doesn't handle this specific scenario correctly.
I want to be able to have it enabled for users coming to my site as it improves user experience, at the same time I don't want to use cookies to store locale as when user is navigating through the site locale is known and may not be the same as browser default.
`nuxt-i18n` doesn't allow that and forces browser detection always when no cookie is present thus breaking user locale choice.

As for considering other options, they all work as they did before.
And the naming, right my bad, it was a quick hack so it skipped my mind.

rchl:
> `detectBrowserLocale: false` doesn't handle this specific scenario correctly.
> I want to be able to have it enabled for users coming to my site as it improves user experience,

Can you point out a specific example (with nuxt-i18n configuration and steps to reproduce)?

If I knew the actual issue maybe I could suggest some generic change rather than introducing a new option. As I said, I'd rather not introduce new options. Especially one that makes things more confusing with already confusing options available in `detectBrowserLocale`. It's not clear how it interacts with other options right now.

Also:
 - missing documentation
 - the test doesn't have to be a browser test but can be just pure `module` test
(but you don't have to address those until we figure out if this change makes sense)

divine:
> If I knew the actual issue maybe I could suggest some generic change rather than introducing a new option. 

Well, it's better to apply minor change without breaking up anything.

My suggestion would be that @rchl _could_ prepare minor (🙏) fix with exluded 'no_prefix' as we had discussion already in https://github.com/nuxt-community/nuxt-i18n/pull/603

stlhrt:
Easiest to reproduce is to edit the test I wrote and set pathOverride to
false. It perfectly shows what is happening.

śr., 24 cze 2020, 15:37 użytkownik Rafał Chłodnicki <
notifications@github.com> napisał:

rchl:
> Easiest to reproduce is to edit the test I wrote and set pathOverride to false. It perfectly shows what is happening.

In which case this will behave differently from `detectBrowserLanguage: false`?

stlhrt:
Argh... you're right. I didn't test that case. This requires more attention :)

PR #781, opened 2020-06-29
Fix language detection from route path

divine:
Fix language detection from route path on all strategies except no_prefix

Not happy with this code block but somehow we need to change the cookie as well.

```
if (strategy !== STRATEGIES.NO_PREFIX && route.path !== '/') {
	if (useCookie) {
	  app.i18n.setLocaleCookie(finalLocale)
	} else {
	  return finalLocale
	}
}
```

@rchl what do you think?

Thanks!

### DIFF
--- a/src/templates/plugin.main.js
+++ b/src/templates/plugin.main.js
@@ -136,7 +136,9 @@ export default async (context) => {
 
     let matchedLocale
 
-    if (useCookie && (matchedLocale = app.i18n.getLocaleCookie())) {
+    if (strategy !== STRATEGIES.NO_PREFIX && (matchedLocale = getLocaleFromRoute(route)) && route.path !== '/') {
+      // Get preferred language from locale route
+    } else if (useCookie && (matchedLocale = app.i18n.getLocaleCookie())) {
       // Get preferred language from cookie if present and enabled
     } else if (process.client && typeof navigator !== 'undefined' && navigator.languages) {
       // Get browser language either from navigator if running on client side, or from the headers
@@ -149,7 +151,13 @@ export default async (context) => {
 
     // Handle cookie option to prevent multiple redirections
     if (finalLocale && (!useCookie || alwaysRedirect || !app.i18n.getLocaleCookie())) {
-      if (finalLocale !== app.i18n.locale) {
+      if (strategy !== STRATEGIES.NO_PREFIX && route.path !== '/') {
+        if (useCookie) {
+          app.i18n.setLocaleCookie(finalLocale)
+        } else {
+          return finalLocale
+        }
+      } else if (finalLocale !== app.i18n.locale) {
         return finalLocale
       } else if (useCookie && !app.i18n.getLocaleCookie()) {
         app.i18n.setLocaleCookie(finalLocale)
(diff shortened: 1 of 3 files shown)


src/templates/plugin.main.js
```diff
@@ -149,7 +151,13 @@ export default async (context) => {
 
     // Handle cookie option to prevent multiple redirections
     if (finalLocale && (!useCookie || alwaysRedirect || !app.i18n.getLocaleCookie())) {
-      if (finalLocale !== app.i18n.locale) {
+      if (strategy !== STRATEGIES.NO_PREFIX && route.path !== '/') {
+        if (useCookie) {
+          app.i18n.setLocaleCookie(finalLocale)
+        } else {
+          return finalLocale
+        }
```
rchl, line 159: It should be possible to handle it better by not saving the cookie on almost every navigation and also this function should return false if the locale is not supposed to change.

src/templates/plugin.main.js
```diff
@@ -136,7 +136,9 @@ export default async (context) => {
 
     let matchedLocale
 
-    if (useCookie && (matchedLocale = app.i18n.getLocaleCookie())) {
+    if (strategy !== STRATEGIES.NO_PREFIX && (matchedLocale = getLocaleFromRoute(route)) && route.path !== '/') {
+      // Get preferred language from locale route
+    } else if (useCookie && (matchedLocale = app.i18n.getLocaleCookie())) {
```
rchl, line 141: I'm not even sure if the `doDetectBrowserLanguage` function is the right place to handle this. There is some logic before `doDetectBrowserLanguage` is called that determines locale to use based on other factors and settings and this change will potentially override that previous decision.

I will have to graph all possible flows in my head and try to test them to figure out what's the best place to handle that.

test/module.test.js
```diff
@@ -716,6 +716,17 @@ describe('prefix_and_default strategy', () => {
     const links = dom.querySelectorAll('head link[rel="canonical"]')
     expect(links.length).toBe(0)
   })
+
+  test('default locale from locale path with en locale cookie', async () => {
```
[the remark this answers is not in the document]
divine: Okay, I've just tried to help, new to nuxt so accept my apologies for any mistakes made 🤔

src/templates/plugin.main.js
divine: @rchl did you get a chance to take a look if not I might try to something again.

Thanks!

rchl: I've started making tests. I will put out a PR so you can track the progres.
I'm on vacation next two weeks so won't have the time to work on it probably.

divine: Hi @rchl, I do really hope that you'd a wonderful vacation and felling good!

I'm just patiently waiting for that PR, maybe there is something that I could be helpful, please just let me know, thanks!

rchl: Sorry for the long wait. I'll try to address it as soon as I can.
I've decided to implement it as an option and potentially make it a default in the future as I'm not confident that it won't break some use cases.

PR #793, opened 2020-07-15
fix(types): fix type of detectBrowserLanguage to allow false

shgtkshruch:
Enables to set `detectBrowserLanguage` to be `false` in case of disabling redirection.

https://github.com/nuxt-community/nuxt-i18n/blob/master/docs/options-reference.md


rchl:
Thanks.

PR #801, opened 2020-07-22
chore(docs): update repository name and documentation links

rchl:
The most important change is updating links to the documentation which
has changed and old ones don't work anymore.

The package was not renamed from "nuxt-i18n" to "@nuxtjs/i18n" as that
would be a breaking change, requiring a change of the module's name in
the modules array.


.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ executors:
         default: false
```
pi0: As we started migrating all repos to Github actions, it is required to  stick with circleci? (I really love circleci suggestion is to keep consistency)

pi0 (approved):

.circleci/config.yml
rchl: I'm fine with disabling circleci as from my experience github actions were working equally fine if not better (faster).

The browser tests are disabled on circleci as they didn't work properly. Don't remember exact details but it was an unresolved issue on their side, I believe.

rchl: My commit message:
```
It's randomly failing chrome browser testing and since github
is also testing everything, we'll use that for now.
```

rchl: I don't have admin rights to disable building on circleci. Can you do that @pi0? The link should be https://app.circleci.com/settings/project/github/nuxt-community/i18n-module?return-to=https%3A%2F%2Fapp.circleci.com%2Fpipelines%2Fgithub%2Fnuxt-community%2Fi18n-module

PR #820, opened 2020-08-02
fix(lazy): include default locale in the main bundle

### LINKED ISSUE #818: Default locale should be loaded from main bundle
If the user has specified `defaultLocale` and is using `lazy` option to load messages, the lang file for default locale should be loaded from the main bundle as it's always needed (either for the current locale or to be used as fallback).

This was brought up in both #486 and #815.

rchl:
This avoid extra request for the locale that is always needed anyway.

Resolves #818

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -35,6 +35,29 @@ module.exports = function (userOptions) {
     return
   }
 
+  let defaultLangFile
+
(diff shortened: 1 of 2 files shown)


rchl:
Dup

PR #823, opened 2020-08-03
fix(lazy): make client use translations loaded on the server

### LINKED ISSUE #486: Preload lang chunks

Hello,

I'm working on a 30-lang website and it would be great to preload some lang chunks.

Ex:
- On http://website.com/it/ ('it' and 'en' chunks are preloaded, as 'en' is the fallback language)
- On http://website.com/fr/ ('fr' and 'en' chunks are preloaded)

Btw, is it possible to do so today?

Thanks,

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c319">#c319</a>)</em></sub></div>

### LINKED ISSUE #663: Dynamic content in locale file is loaded in SSR then reloaded on the client during hydration
### Version

[v6.8.1](https://github.com/nuxt-i18n/releases/tag/v6.8.1)

### Reproduction link

[https://none](https://none)

### Steps to reproduce

- Enable `lazy: true` in nuxt.config.js.
- Create an `/locales/en.js`
- In the en.js file, add a `message: Math.random()` translation
- In a template, display `{{ $t('message') }}`
- Reload the browser
- Observe the page loads with the translations already applied in SSR, showing a random number
- Observe that a few moments later, hydration happens and the random number is replaced with a new number

### What is expected ?

Only the SSR version of the translations would be applied and hydration would not actually update the translations.

Hydration is updating the translations, changing the content from what SSR generates.

### Additional comments?

The reason this came up is we're using a lorem ipsum module to quickly mock up fake content in our translations. Having the content change suddenly after first load is jarring.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c436">#c436</a>)</em></sub></div>

rchl:
Instead of client importing and processing messages (translations)
itself on load, pass already loaded messages from the server through
"nuxtState". This will in most cases allow avoiding triggering an extra
network request to fetch lang file for the given locale.

Slight behavior change: The client will no longer import the lang file
itself for the initially used locale. That means that if the lang file
has exported a function, it will only be called on the server and not on
the client. This matches behavior of "asyncData" and I think it makes
more sense in general.

Resolves #486
Resolves #663

### DIFF
--- a/docs/es/seo.md
+++ b/docs/es/seo.md
@@ -1,5 +1,9 @@
 # Optimización para motores de búsqueda
 
+::: tip NOTE
+Using `seo` option (or alternatively the `$nuxtI18nSeo`-based solution - see [Improving Performance](#improving-performance)) requires that locales are configured as an array of objects and not strings.
+:::
+
 Con la opción `seo`  habilitada, **nuxt-i18n** intenta agregar algunos metadatos para mejorar sus páginas SEO. Esto es lo que hace:
 
 * Agregue un atributo _lang_ que contenga el código ISO del entorno local actual a la etiqueta  `<html>`.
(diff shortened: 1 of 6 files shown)


divine:
New release as well?

rchl:
Published in v6.13.3

tvld:
Great )).... 
Q: is this a breaking change for clients that load the language module proactively at present ?

rchl:
What exact configuration do you have in mind?
(This will only affect users that use `lazy: true`)

tvld:
We use ``lazy=true`` but, and I just checked, it seems not in the projects were we load the files ourselves so to work around the issue... So my question was premature; sorry)).
We will rewrite one of the projects with  v6.13.3 and will come back if any questions remain on that front )))
Thanks again ))

yakovyarmo:
for me it still doesn't work since process.server is undefined

Issue #735, opened 2020-05-26
How to translate routes without using locale prefix


jonalxh:
hI, I'm trying to do an app which works in Spanish and English, I need it can be accesed with page name and translated name, i.e: I have `about.vue `in `pages` folder and I need to access it with **/about** and **/nosotros**. I've tried with your plugin but it causes an **/es/nosotros** route and I need it to work without locale prefix.

Adding `no_prefix` strategy to` nuxt-i18n` config in `nuxt.config.js` doesn't work too.

I've set the following code in `about.vue` without success:

```
 export default {
	nuxtI18n: {
		paths: {
			es: '/nosotros',
			en: '/about'
		}
	},
}

```

nuxt.config.js:
```
modules: [
    ['nuxt-i18n', {
      strategy: 'no_prefix',
      lazy: true,
      parsePages: false, //Tried with true and does not work
      locales: [
        {
          name: 'Español',
          code: 'es',
          file: 'es/locale.json'
        },
        {
          name: 'English',
          code: 'en',
          file: 'en/locale.json'
        }
      ],
      langDir: '/locales/',
      defaultLocale: 'en'
    }]
]
```

rchl:
With `no_prefix` no prefixes are created.
The `about.vue` snippet also seems correct.

You must be doing some a simple mistake somewhere if that doesn't work.
Please create reproduction on github or https://codesandbox.io/ if you want me to look at it.

jonalxh:
Do you have a working example for me to try before creating the code snippet?
Thanks.

jonalxh:
I was reading [this article](https://github.com/nuxt-community/nuxt-i18n/blob/master/docs/es/routing.md)  and it says **no_prefix** does not work with **Custom Routers**. What should I do to correctly translate my routes?

rchl:
Here is some example you can fork and modify: https://codesandbox.io/s/i18n-plugin-i18nlocales-ouwhn

> I was reading [this article](https://github.com/nuxt-community/nuxt-i18n/blob/master/docs/es/routing.md) and it says **no_prefix** does not work with **Custom Routers**.

Ah yes, that's true.
To be fair not sure why. I think it might still be possible to support custom routes with `no_prefix` strategy. I could have a look at that at some point.

jonalxh:
[...]

PR #832, opened 2020-08-05
fix(differentDomains): Match domain properly on client if port provided

### LINKED ISSUE #831: Locale detection don't work on client side, with differentDomains enabled and vuex sync disabled
### Version
nuxt-i18n: 6.13.6 (older too)
nuxt: 2.13.1 (older too)

 - [x] universal
 - [] spa

### Nuxt-i18n configuration
```
["nuxt-i18n", {
    differentDomains: true,
    defaultLocale: config.LOCALE_DEFAULT,
    locales,
    parsePages: false,
    pages,
    vueI18n: {
        fallbackLocale: config.LOCALE_FALLBACK,
        silentFallbackWarn: true,
        messages,
    },
    vueI18nLoader: true,
    vuex: {
        syncMessages: false,
        syncLocale: false,
    },
}],
```

### Steps to reproduce
1. Create setup with 
```
{
    differentDomains: true,
    defaultLocale: "any-locale",
    vuex: {
        syncMessages: false,
        syncLocale: false,
    }
}
```
2. open in browser non-default locale, on server it will work as expected, but on client it will be default

rchl:
If locale's "domain" contained a port then that wasn't matched correctly
on the client as it compared with "location.hostname" that doesn't include
the port. Use "location.host" instead.

Also, force "no-prefix" strategy when using "differentDomains" as
otherwise duplicate routes are created with the same paths but different
names. That has then caused issues when detecting locale from the route
as it would detect different locale and override what was set on
"$i18n.locale" previously.

Resolves #831

### DIFF
--- a/docs/content/en/different-domains.md
+++ b/docs/content/en/different-domains.md
@@ -9,10 +9,7 @@ You might want to use a different domain name for each language your app support
 
 * Set `differentDomains` option to `true`
 * Configure `locales` option as an array of objects, where each object has a `domain` key which value is the domain name you'd like to use for that locale (including port if non-default)
-
-<alert type="info">
-
-You might want to set `detectBrowserLanguage` to `false`. When enabled (it is by default), user can get redirected to a different domain on first visit. Set to `false` if you want to ensure that visiting given domain always shows page in the corresponding locale.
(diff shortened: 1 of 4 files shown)


rchl:
Have to leave out the part about "forcing no-prefix strategy" as that would require a map of customized routes (#735) to work properly in all cases.

PR #855, opened 2020-08-20
Fixed matching browser locale when the only matching component is the language

koendeschacht:
Previously, nuxt-i18n would not find a match with app locales `['en-IN', 'hi-IN']` and browser locales: `['en-US', 'en-GB', 'en']`. Now the following cases:

app locales: ['en-IN', 'hi-IN'] browser locales: `['en-US', 'en-GB', 'en']`

and

app locales: ['en-IN', 'hi-IN'] browser locales: `['en-US', 'en-GB']`

result in the chosen locale `en-IN`.

### DIFF
--- a/src/templates/utils-common.js
+++ b/src/templates/utils-common.js
@@ -38,11 +38,20 @@ export const matchBrowserLocale = (appLocales, browserLocales) => {
   for (const [index, browserCode] of browserLocales.entries()) {
     if (browserCode.includes('-')) {
       // For backwards-compatibility, this is lower-cased before comparing.
-      const languageCode = browserCode.split('-')[0].toLowerCase()
+      const languageCodeBrowser = browserCode.split('-')[0].toLowerCase()
 
-      if (appLocales.includes(languageCode)) {
+      const matchedCode = appLocales.find(appCode => {
+        let languageCodeLocale
+        if (appCode.includes('-')) {
+          languageCodeLocale = appCode.split('-')[0]
+        } else {
+          languageCodeLocale = appCode
+        }
+        return languageCodeLocale.toLowerCase() === languageCodeBrowser
+      })
+      if (matchedCode) {
         // Deduct a thousandth for being non-exact match.
-        matchedLocales.push({ code: languageCode, score: 0.999 - index / browserLocales.length })
+        matchedLocales.push({ code: matchedCode, score: 0.999 - index / browserLocales.length })
         break
       }
     }
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -94,4 +94,11 @@ describe('matchBrowserLocale', () => {
 
     expect(matchBrowserLocale(appLocales, browserLocales)).toBe('en-gb')
   })
+
+  test('matches locale when only languages match', () => {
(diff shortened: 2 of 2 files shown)


src/templates/utils-common.js
```diff
@@ -38,11 +38,20 @@ export const matchBrowserLocale = (appLocales, browserLocales) => {
...
+      const languageCodeBrowser = browserCode.split('-')[0].toLowerCase()
 
-      if (appLocales.includes(languageCode)) {
+      const matchedCode = appLocales.find(appCode => {
+        let languageCodeLocale
+        if (appCode.includes('-')) {
+          languageCodeLocale = appCode.split('-')[0]
+        } else {
+          languageCodeLocale = appCode
+        }
```
rchl, line 49: This is shorter:
```suggestion
        const languageCodeLocale = appCode.split('-')[0].toLowerCase()
```

but I'm not sure I agree with the logic where

> app locales: ['en-IN', 'hi-IN'] browser locales: `['en-US', 'en-GB']` result in the chosen locale `en-IN`.

I'm not sure if those en-* variants are sufficiently similar to use them as a substitute. I mean, we know that English US and UK are very similar and would work but I'm not sure if that applies in all cases (for other countries).

rchl (changes requested):
Ops. Looks like I forgot to send this feedback. Sorry for the delay.

src/templates/utils-common.js
koendeschacht: Good point. I'm not sure either. As I was modifying the PR, I also started to doubt some other cases. Take for example:
nuxt-i18n config: 
```
{
locales:['en-CA', 'fr-CA'],
defaultLocale: 'en-CA'
}
```

and browser locales 
```
['en-UK', 'fr']
```

Which one should be matched here? Or

config: 
```
{
locales:['fr-CA', 'en-CA'],
defaultLocale: 'fr-CA'
}
```

and browser locales 
```
['fr', 'en-CA']
```

?? 

I was thinking however, the problem of matching browser locales has existed for such a long time, there should be specification or reference implementation out there somewhere. I'll already found these:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat/supportedLocalesOf
https://tc39.es/ecma402/#sec-supportedlocales 

The default matcher in Javascript seems to be 'best-fit', but if my quick googling is correct neither Firefox or the V8 engine actually implemented this:

https://searchfox.org/mozilla-central/source/js/src/builtin/intl/CommonFunctions.js#575
https://chromium.googlesource.com/v8/v8/+/4.3.49/src/i18n.js?autodive=0%2F%2F (search for bestFitSupportedLocalesOf)

The fallback match algorithm is 'lookup' but I could not quickly grasp the specification and apply it to this case. I'll continue on it later tonight, and maybe also see  whether there are some server packages (Spring? Rails?) that have an implementation of the 'best-fit' algo.

rchl: It seems that #965 is pretty much having the same effect but also adds support for matching ISO codes so I'll handle this issue through that other PR.

Thanks for your efforts though and sorry that it took forever.

PR #864, opened 2020-08-25
feat(docs): migrate docs to Nuxt Content

debs-obrien:
@pi0 migration of docs to content including image and spanish lang


rchl:
Wow, thank you. 👍 

I'll review after kid goes to bed.
I only wonder about whether the links will change and whether we should address that with some redirects if so.

debs-obrien:
no problem  @rchl although I must admit it was a time consuming one especially with multi lang. 

url's are all exactly the same so shouldn't be any problem. only difference is vuepress puts a .html at the end and content doesn't. so you shouldn't need to do any redirects.

rchl:
@debs-obrien Do you also have the full-size logo in SVG format?
I think we should also update logo used in the readme so that it's consistent.
We could handle that separately of course but it's nice to have SVG also for authoring.

rchl:
~~Is the docs Nuxt missing `@nuxtjs/tailwindcss` dependency?~~
Also getting:
> warning "@nuxt/content-theme-docs > @tailwindcss/typography@0.2.0" has unmet peer dependency "tailwindcss@^1.5.0".

rchl:
No worries about that warning if you don't know the answer. I've looked into it and it's a bit complicated system of dependencies. :)

rchl:
Looks good to me apart from https://github.com/nuxt/content/issues/412 but that's not a blocker and maybe we could also add SVG version of the icon.png to the repo so that we have it in case we need to generate different image sizes.

debs-obrien:
done. commit message should have said icon not logo

rchl:
No problem. It's all gonna be squashed anyway.

PR #868, opened 2020-08-26
fix(docs): Typo wrong route path

x-N0:
Wrong route to the spanish path to routing


rchl:
A better way to do it is to use a relative link.

But since I've found many other links that are broken, I've fixed them all together in 19c516c66b3e3f9a39af51b11591d965b945dc84.

Thanks for brining my attention to it. :)

PR #886, opened 2020-09-09
Update options-reference.md

geantas:
Fixed typo


divine:
Almost ready in this PR https://github.com/nuxt-community/i18n-module/pull/799/files#diff-25585ead48e3603bb79c736736f47982R65

rchl:
Thanks but I'll handle it through #799 soon.
(Also es has the same issue).

PR #891, opened 2020-09-12
fix(docs): add warnings regarding no_prefix strategy

Khalphion:
Added warning to the custom path and ignore routes section of the routing documentation

I've been spending a bit of time why my page was not found until I noticed that the custom path feature was in conflict with the `no_prefix` strategy, but nothing of that was mentioned in the corresponding section.

### DIFF
--- a/docs/content/en/routing.md
+++ b/docs/content/en/routing.md
@@ -106,6 +106,12 @@ If on `Nuxt` version lower than 2.10.2, and using strategy `prefix_except_defaul
 
 In some cases, you might want to translate URLs in addition to having them prefixed with the locale code. There are 2 ways of configuring custom paths for your pages: in-component options or via the module's configuration.
 
+<alert type="warning">
+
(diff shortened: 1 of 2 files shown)


docs/content/en/routing.md
```diff
@@ -106,6 +106,12 @@ If on `Nuxt` version lower than 2.10.2, and using strategy `prefix_except_defaul
 
 In some cases, you might want to translate URLs in addition to having them prefixed with the locale code. There are 2 ways of configuring custom paths for your pages: in-component options or via the module's configuration.
 
+<alert type="warning">
+
+Custom paths are not supported with the `no-prefix` [strategy](#strategy).
```
rchl, line 111: I think that linking to `strategy` might not be necessary. But I guess it's no big deal.

rchl (changes requested):
Similar change should be also made in a Spanish version (just put English text there, it's only important that all versions are in sync).

rchl:
Thanks.

PR #904, opened 2020-09-23
Added option useSortRoutes

smaknsk:
if change routes in nuxt.config.js
```
...
router: {
    extendRoutes(routes, resolve) {
      const index = routes.findIndex(item => item.name == 'profile-id');
      routes.splice(index, 1, {
        name: 'profile',
        path: '/profile-:id',
        component: resolve(__dirname, 'pages/profile/id.vue')
      })
}
...
```
nuxt-i18n reset my order routes
result:
```
/ru
/ru/:category
/ru/profile-:id
/:category
/profile-:id
/
```
bug: route `/ru/profile-:id` and `/profile-:id` never resolve

expected:
```
/profile-:id
/ru/profile-:id
/
/ru
/:category
/ru/:category
```
this fix added useSortRoutes as option
```
...
i18n: {
    useSortRoutes: false,
   ...
}
...
```

### DIFF
--- a/src/helpers/constants.js
+++ b/src/helpers/constants.js
@@ -51,6 +51,7 @@ exports.DEFAULT_OPTIONS = {
   },
   parsePages: true,
   pages: {},
+  useSortRoutes: true,
   beforeLanguageSwitch: () => null,
   onLanguageSwitched: () => null
 }
--- a/src/helpers/routes.js
+++ b/src/helpers/routes.js
@@ -13,7 +13,8 @@ exports.makeRoutes = (baseRoutes, {
   parsePages,
   routesNameSeparator,
   strategy,
-  trailingSlash
+  trailingSlash,
+  useSortRoutes
 }) => {
   locales = getLocaleCodes(locales)
   let localizedRoutes = []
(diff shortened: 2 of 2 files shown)


rchl:
I'd rather have the sorting bug fixed than introducing an option to workaround the bug.

The sorting function is in Nuxt.

FYI: @pimlie

smaknsk:
@rchl Yes and no
my case is a special case - yes, the problem is probably in sortRoutes
but nuxt makes it possible to do "custom" sorting 

Nuxt assumes calling sortRoutes manually as an option 
https://ru.nuxtjs.org/api/configuration-router/#extendroutes
`If you want to sort your routes, you can use the sortRoutes (routes) function`

i don't want to call sorting
but i18n-module calls sortRoutes hard and may break my "custom" sort

rchl:
But I'd argue that when using this module sorting is almost mandatory since without it you need to order locale in a certain way for all to be accessible. It's not obvious and difficult to explain.

Would fixing the issue with sortRoutes not fix your problem? If it would then I'd prefer that.

But if you insist on adding the setting then there are a couple of things left to do:
 - rename to `sortRoutes` (since `use` implies that we are using a function named in a certain way which is an internal detail that doesn't have to be communicated).
 - add documentation for the new option (in `options-reference.md`)
 - update typescript types for settings in `nuxt-i18n.d.ts`

PR #914, opened 2020-10-04
Update lazy-load-translations.md

uran1980:
fix typo


rchl:
Can you also update the equivalent `es` page?

rchl:
Thanks. Pushed to master with updated 'es' locale also.

PR #922, opened 2020-10-08
fix: add x-default hreflang tag for improved SEO

divine:
Add x-default hreflang to improve SEO.

https://webmasters.googleblog.com/2013/04/x-default-hreflang-for-international-pages.html

https://support.google.com/webmasters/answer/189077?hl=en#xdefault

### DIFF
--- a/src/templates/seo-head.js
+++ b/src/templates/seo-head.js
@@ -1,5 +1,6 @@
 import VueMeta from 'vue-meta'
 import {
+  defaultLocale,
   COMPONENT_OPTIONS_KEY,
   LOCALE_CODE_KEY,
   LOCALE_ISO_KEY,
@@ -72,6 +73,15 @@ function addHreflangLinks (locales, baseUrl, link) {
       rel: 'alternate',
       href: baseUrl + this.switchLocalePath(mapLocale.code),
       hreflang: iso
+    })
+  }
+
+  if (defaultLocale) {
+    link.push({
+      hid: 'alternate-x-default',
+      rel: 'alternate',
+      href: baseUrl + this.switchLocalePath(defaultLocale),
+      hreflang: 'x-default'
     })
   }
 }
--- a/test/module.test.js
+++ b/test/module.test.js
@@ -191,6 +191,12 @@ for (const trailingSlash of TRAILING_SLASHES) {
           rel: 'alternate',
           href: pathRespectingTrailingSlash('nuxt-app.localhost/fr'),
           hreflang: 'fr-FR'
+        },
+        {
+          tagName: 'link',
+          rel: 'alternate',
+          href: 'nuxt-app.localhost/',
+          hreflang: 'x-default'
         }
       ]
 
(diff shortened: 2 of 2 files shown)


divine:
@rchl is everything good with this PR?

Thanks!

rchl:
After reading about `x-default` I'm a bit uncertain whether it should be used by this module. The guides seem to emphasize that the URL marked with `x-default` is for when:

> page doesn’t target any specific language or locale and is the default page when no other page is better suited.

In our case, we are marking the page that targets a specific locale (typically `en`) with it.

For example, on https://webmasters.googleblog.com/2013/04/x-default-hreflang-for-international-pages.html they are listing those tags:
```
<link rel="alternate" href="http://example.com/en-gb" hreflang="en-gb" />
<link rel="alternate" href="http://example.com/en-us" hreflang="en-us" />
<link rel="alternate" href="http://example.com/en-au" hreflang="en-au" />
<link rel="alternate" href="http://example.com/" hreflang="x-default" />
```
which don't include something like:
```
<link rel="alternate" href="http://example.com/" hreflang="en" />
```
They only list it as the `x-default` URL.

I'll try to google more information about it tomorrow. Maybe someone had similar concerns and got a clear answer already.

divine:
@rchl I can understand your concerns, here is more information:

> Hreflang tags work on a best-match basis. In other words, Google returns the version of content that it deems to be the best match based on various signals like the user’s country and language settings.
> 
> To illustrate how this works, imagine that Google is returning a result for an English-speaking user located in Spain.
> 
> Google first looks for a language-country match (e.g., en-es) and returns that page if it exists.
> 
> If not, it looks for a language code match (e.g., en-us). Crucially, the country code doesn’t matter. Google simply looks for a matching language code.
> 
> If that doesn’t exist, Google will fallback to the x‑default version.
> Source: https://ahrefs.com/blog/hreflang-tags/

and here is how they're using it:
> Hreflang annotations for https://ahrefs.com/blog/zh/website-audit/:
> 
> ```
> <link rel="alternate" hreflang="x-default" href="https://ahrefs.com/blog/website-audit/" />
> <link rel="alternate" hreflang="en" href="https://ahrefs.com/blog/website-audit/" />
> <link rel="alternate" hreflang="ru" href="https://ahrefs.com/blog/ru/audit-saita/" />
> ```
> Source: https://help.ahrefs.com/en/articles/2754344-what-does-the-self-referencing-hreflang-annotation-missing-issue-in-site-audit-mean


Basically they're referencing x-default again while having `/en` already, this is basically used to mention default language for search engines if any other languages doesn't fit at all.

Also Google's article is from 2013 a little bit outdated without actual examples (sorry about not giving ahrefs link in the first place).

Edit: Here is exact answer https://moz.com/community/q/x-default-on-hreflang-tags


Thanks!

rchl:
OK. Let's try it.

rchl:
Released in Fixes released in https://github.com/nuxt-community/i18n-module/releases/tag/v6.15.2

PR #927, opened 2020-10-13
Fix various SEO issue with canonical tag

### LINKED ISSUE #912: Canonical URL for prefixed strategy

Currently prefixed strategy generating* `index.html` for each page which is solely used for redirection does create a duplicate content. It's not possible without `index.html` to detect and force redirection.

The problem here is that Google can't decide which is canonical URLs on it's own.

It's now detecting `/en` as duplicate of `/` (root) path.

See: https://support.google.com/webmasters/answer/182192?hl=en#3

Canonical URL enabled for prefixed strategy as well. It's only enabled for `PREFIX_AND_DEFAULT  `https://github.com/nuxt-community/i18n-module/blob/master/CHANGELOG.md#features-24

Thanks!

rchl:
Best to look at individual commits since each is a self-contained fix for a separate, small issue.

Resolves #912 

@divine Would be nice if you could check if this resolves all your issues.
And if there are some that are still unresolved then maybe consider creating a separate issue.

### DIFF
--- a/docs/content/en/api.md
+++ b/docs/content/en/api.md
@@ -26,12 +26,6 @@ All [Vue I18n properties and methods](http://kazupon.github.io/vue-i18n/api/#vue
 
   See also [Basic usage - nuxt-link](../basic-usage#nuxt-link).
 
-<alert type="warning">
-
-In `no_prefix` strategy, passing `locale` other than the current one is not supported.
-
-</alert>
-
 #### switchLocalePath
 
   - **Arguments**:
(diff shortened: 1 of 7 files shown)


divine:
@rchl I've just checked, it works perfectly and let me thank you for all the work that you've done!

Probably with all this SEO improvements we might see soon more nuxt sites within search results 🎉!

Thanks!

rchl:
(just rebased with last push, no changes)

PR #951, opened 2020-11-03
fix(docs): include "defaultLocale" in the example configuration

bottlehs:
defaultLocale added successfully. If defaultLocale is not added, it will not run if the code is written as illustrated.


rchl:
Thanks.

PR #952, opened 2020-11-04
Korean Langeuge add or Inertia content-en,es/basic-usage.md 🍀

bottlehs:
- Add Korean
- Inertia content-en,es/basic-usage.md


rchl:
There is really no need to add another locale to examples. These are just examples and having 2 or 3 there already is enough to get the idea of how to add your own.

PR #960, opened 2020-11-09
Update options-reference.md

unr:
Removed unwanted characters.

image


rchl:
There is the same issue in the `es` version. Could you fix that too?

rchl:
Fixes through bc9f76e. Thanks.

PR #963, opened 2020-11-15
feat: API for handling locale change during page transitions

### LINKED ISSUE #150: Page transition broken with nuxt-i18n
### Version

### Reproduction link

[https://github.com](https://github.com)

### Steps to reproduce

I have simple fade page transition

```css
.page-enter-active {
  backface-visibility: hidden;
  will-change: opacity;
  opacity: 1;
  transition: opacity 450ms linear;
}

.page-leave-active {
  backface-visibility: hidden;
  will-change: opacity;
  opacity: 1;
  transition: opacity 450ms linear;
}

.page-enter {
  opacity: 0;
}

.page-leave-active {
  opacity: 0;
}
```

The problem is, when I change locale, text content changes **instantly** while page transition is not complete. 
In result I have a blink of content. 

Is there'a a way to solve this? 

### What is expected ?

No blink, smooth transition

Transition broken, content blink

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c163">#c163</a>)</em></sub></div>

pmrotule:
Closes #150 

The issue is happening since `nuxt-i18n` is setting the locale from `onNavigate` being called in a middleware which won't wait for the page transition to fade out. Unfortunately, this cannot be fully solved from this module since it is not possible to set a global `pageTransition.beforeEnter` while making sure it won't be overwritten from a `transition.beforeEnter` defined in a page component.

That being said, the new option `skipSettingLocaleOnNavigate` would allow us to workaround the issue. If it is `true`, the locale will not be set on navigate which would let us setting it ourselves from a `beforeEnter` transition hook:

```js
// nuxt.config.js

export default {
  // ...
  plugins: ['~/plugins/router'],

  i18n: {
    // ...
    skipSettingLocaleOnNavigate: true,
  },
}
```

```js
// plugins/router.js

export default ({ app }) => {
  app.nuxt.defaultTransition.beforeEnter = () => {
    app.i18n.setPendingLocale()
  }
  // Optional: wait for locale before scrolling for a smoother transition
  app.router.options.scrollBehavior = async (to, from, savedPosition) => {
    // Make sure the route has changed otherwise `beforeEnter` will never be called
    if (to.name !== from.name) await app.i18n.waitForPendingLocale()
    return savedPosition || { x: 0, y: 0 }
  }
}
```

```js
// in any page component

export default {
  transition: {
    beforeEnter() {
      this.$i18n.setPendingLocale()
    },
  }
}
```

I'll wait for your feedback before adding test and documentation. I hope it makes sense.

### DIFF
--- a/docs/content/en/api.md
+++ b/docs/content/en/api.md
@@ -108,22 +108,6 @@ Instance of [VueI18n class](http://kazupon.github.io/vue-i18n/api/#vuei18n-class
 
   Switches locale of the app to specified locale code. If `useCookie` option is enabled, locale cookie will be updated with new value. If prefixes are enabled (`strategy` other than `no_prefix`), will navigate to new locale's route.
 
-#### setPendingLocale
-
-  - **Arguments**:
-    - no arguments
-  - **Returns**: `Promise<undefined>`
-
-  Switches to the pending locale that would have been set on navigate, but was prevented by the option [`skipSettingLocaleOnNavigate`](./options-reference#skipsettinglocaleonnavigate). See more information in [Wait for page transition](./lang-switcher#wait-for-page-transition).
-
-#### waitForPendingLocale
-
-  - **Arguments**:
-    - no arguments
-  - **Returns**: `Promise<undefined>`
-
-  Returns a promise that will be resolved once the pending locale is set.
-
 #### getBrowserLocale
 
   - **Arguments**:
@@ -131,6 +115,22 @@ Instance of [VueI18n class](http://kazupon.github.io/vue-i18n/api/#vuei18n-class
   - **Returns**: `string | undefined`
 
   Returns browser locale code filtered against the ones defined in options.
+
+#### setPendingLocale <badge>v6.20.0+</badge>
+
+  - **Arguments**:
+    - no arguments
+  - **Returns**: `Promise<undefined>`
+
+  Switches to the pending locale that would have been set on navigate, but was prevented by the option [`skipSettingLocaleOnNavigate`](./options-reference#skipsettinglocaleonnavigate). See more information in [Wait for page transition](./lang-switcher#wait-for-page-transition).
+
+#### waitForPendingLocale <badge>v6.20.0+</badge>
+
+  - **Arguments**:
+    - no arguments
+  - **Returns**: `Promise<undefined>`
+
+  Returns a promise that will be resolved once the pending locale is set.
 
 ### Properties
 
--- a/docs/content/en/api.md
+++ b/docs/content/en/api.md
@@ -116,7 +116,7 @@ Instance of [VueI18n class](http://kazupon.github.io/vue-i18n/api/#vuei18n-class
 
   Returns browser locale code filtered against the ones defined in options.
 
-#### setPendingLocale <badge>v6.20.0+</badge>
+#### finalizePendingLocaleChange <badge>v6.20.0+</badge>
 
   - **Arguments**:
     - no arguments
@@ -124,7 +124,7 @@ Instance of [VueI18n class](http://kazupon.github.io/vue-i18n/api/#vuei18n-class
 
   Switches to the pending locale that would have been set on navigate, but was prevented by the option [`skipSettingLocaleOnNavigate`](./options-reference#skipsettinglocaleonnavigate). See more information in [Wait for page transition](./lang-switcher#wait-for-page-transition).
 
-#### waitForPendingLocale <badge>v6.20.0+</badge>
+#### waitForPendingLocaleChange <badge>v6.20.0+</badge>
 
   - **Arguments**:
     - no arguments
--- a/docs/content/en/api.md
+++ b/docs/content/en/api.md
@@ -108,6 +108,22 @@ Instance of [VueI18n class](http://kazupon.github.io/vue-i18n/api/#vuei18n-class
 
   Switches locale of the app to specified locale code. If `useCookie` option is enabled, locale cookie will be updated with new value. If prefixes are enabled (`strategy` other than `no_prefix`), will navigate to new locale's route.
 
+#### setPendingLocale
+
+  - **Arguments**:
+    - no arguments
+  - **Returns**: `Promise<undefined>`
+
+  Switches to the pending locale that would have been set on navigate, but was prevented by the option [`skipSettingLocaleOnNavigate`](./options-reference#skipsettinglocaleonnavigate). See more information in [Wait for page transition](./lang-switcher#wait-for-page-transition).
+
+#### waitForPendingLocale
+
+  - **Arguments**:
+    - no arguments
+  - **Returns**: `Promise<undefined>`
+
+  Returns a promise that will be resolved once the pending locale is set.
+
 #### getBrowserLocale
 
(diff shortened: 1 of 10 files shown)


src/templates/plugin.main.js
```diff
@@ -221,6 +242,8 @@ export default async (context) => {
     i18n.setLocaleCookie = locale => setLocaleCookie(locale, res, { useCookie, cookieDomain, cookieKey, cookieSecure, cookieCrossOrigin })
     i18n.getLocaleCookie = () => getLocaleCookie(req, { useCookie, cookieKey, localeCodes })
     i18n.setLocale = (locale) => loadAndSetLocale(locale)
+    i18n.setPendingLocale = setPendingLocale
+    i18n.waitForPendingLocale = waitForPendingLocale
     i18n.__baseUrl = app.i18n.__baseUrl
   }
```
rchl: Should also copy those internal properties that you've added (`__pendingLocale` and other).

src/templates/plugin.main.js
```diff
@@ -174,11 +175,31 @@ export default async (context) => {
...
+    if (!app.i18n.__pendingLocale) {
+      return
+    }
+    await app.i18n.setLocale(app.i18n.__pendingLocale)
+    app.i18n.__resolvePendingLocalePromise()
+    app.i18n.__pendingLocale = null
+  }
+
+  const waitForPendingLocale = async () => {
+    return !app.i18n.__pendingLocale || (await app.i18n.__pendingLocalePromise)
```
rchl: Would prefer this as it would be more correct from the perspective of the types (I'm planning to enable more strict type-checking at some point):
```suggestion
    if (app.i18n.__pendingLocale) {
      await app.i18n.__pendingLocalePromise
    }
```

rchl (changes requested):
It's not the prettiest solution but I don't know better (better would probably require some changes in Nuxt).

So I'm fine with it for now.

MartinMa:
Very interesting solution.

@pmrotule Is there a way to let i18n-module call the `setPendingLocale()` function on its own? You mentioned `beforeEnter` but there can only be one `beforeEnter` listener I suppose? Is `beforeEnter` the best possible time to change the locale? When exactly is it called (it isn't actually specified in the docs)? Right before the new page begins to transition in? There would still be an edge case when the old page transitions out (say slide to the left) and the new page transitions in at the same time (say slide in from the right). In this edge case both translations should be visible at the same time. I mean, maybe we just have to live with the fact that this doesn't work.

pmrotule:
> Very interesting solution.
> 
> @pmrotule Is there a way to let i18n-module call the `setPendingLocale()` function on its own? You mentioned `beforeEnter` but there can only be one `beforeEnter` listener I suppose? Is `beforeEnter` the best possible time to change the locale? When exactly is it called (it isn't actually specified in the docs)? Right before the new page begins to transition in? There would still be an edge case when the old page transitions out (say slide to the left) and the new page transitions in at the same time (say slide in from the right). In this edge case both translations should be visible at the same time. I mean, maybe we just have to live with the fact that this doesn't work.

@MartinMa `beforeEnter` is indeed the exact time where you need to change it since it is called by the `<transition>` wrapping up your `<router-view />` which differs from the `router.beforeEach` hook. Basically, when you call `switchLocalePath`, it triggers a navigation which will call up to 3 different hooks at the same time:
- `router.beforeEach` will be called
- Any global or matching middleware as well
- The router-view `<transition>` will start

The problem is, `nuxt-i18n` is switching the locale from a middleware so it's switched right away without waiting for the transition to fade out the current page. Unfortunately, `setPendingLocale()` cannot be called from `nuxt-i18n` because it has to be called when the new page transitions in which is happening on `beforeEnter`, but `beforeEnter` can be set globally as well as from page components and in this case, there is really no way of accessing it. The only way I could see it working (like @rchl quickly mentioned) would be by adding a new config in Nuxt allowing to define a `pageTransition.beforeEnter` that couldn't be overwritten from a page component (something like `beforeAnyRouteEnter`).

By the way, sorry about the delay, I should find time in the next few days to wrap up this PR.

pmrotule:
@rchl I finally addressed your feedback and updated the documentation. I tried to add e2e tests, but it wasn't working for me locally. Let me know how you think we should move on with this fix.

AgustinBanchio:
Looking forward to this PR

MartinMa:
@pmrotule Thank you very much. Looks well thought out and well documented.

pmrotule:
@rchl Any feedback? I just rebased again. Should be good to go.

rchl:
I will have a look as soon as possible (hopefully this evening).

docs/content/en/api.md
```diff
@@ -108,6 +108,22 @@ Instance of [VueI18n class](http://kazupon.github.io/vue-i18n/api/#vuei18n-class
 
   Switches locale of the app to specified locale code. If `useCookie` option is enabled, locale cookie will be updated with new value. If prefixes are enabled (`strategy` other than `no_prefix`), will navigate to new locale's route.
 
+#### setPendingLocale
```
rchl: I fear that it's called too similar to `setLocale` and could confuse people. Maybe it should be renamed to something else? Maybe `flushPendingLocale`, or `finalizePendingLocaleChange`?

Then I guess the `waitForPendingLocale` should be changed too, for parity. Maybe `finalizePendingLocaleChange` and `waitForPendingLocaleChange`?

docs/content/en/lang-switcher.md
```diff
@@ -98,3 +98,46 @@ export default {
 **nuxt-i18n** won't reset parameters translations for you, this means that if you use identical parameters for different routes, navigating between those routes might result in conflicting parameters. Make sure you always set params translations in such cases.
 
 </alert>
+
+## Wait for page transition
+
+By default, the locale will be changed right away when calling [`switchLocalePath(path)`](./api#switchlocalepath) which means that if you have a page transition, it will fade out the page with the text already switched to the new language and fade back in with the same content.
```
rchl: > By default, the locale will be changed right away when calling `switchLocalePath(path)`

This statement feels wrong as `switchLocalePath` doesn't change locale but only generates an URL that can be used to change locale through navigating to it.

I've changed it to this in my commit:

> By default, the locale will be changed right away when navigating to a route with a different locale...

Let me know if you disagree.

rchl (commented):
Just some small changes and comments.

docs/content/en/lang-switcher.md
pmrotule: I totally agree. Thanks for cleaning that up.

docs/content/en/api.md
pmrotule: Makes sense. I went for `finalizePendingLocaleChange` and `waitForPendingLocaleChange` as you suggested.
https://github.com/nuxt-community/i18n-module/pull/963/commits/a4c0808d7993ec9686a5cf232695fcaf65973b2f

rchl (approved):

rchl:
Great work! Thanks.

Venegrad:
7.3 not working

Issue #499, opened 2019-10-21
iso code for locales config not used


TimonPeng:
https://github.com/nuxt-community/nuxt-i18n/blob/3a64fdad274571f160046f08f861aa7c3c4d7a15/src/helpers/utils.js#L15

TimonPeng:
https://github.com/nuxt-community/nuxt-i18n/issues/37#issuecomment-419415526

TimonPeng:
https://github.com/nuxt-community/nuxt-i18n/blob/master/src/plugins/main.js#L243

Here check the browser language is in `localeCodes` array or not.

https://github.com/nuxt-community/nuxt-i18n/blob/3a64fdad274571f160046f08f861aa7c3c4d7a15/src/index.js#L43

`localeCodes` defined here.

https://github.com/nuxt-community/nuxt-i18n/blob/3a64fdad274571f160046f08f861aa7c3c4d7a15/src/helpers/utils.js#L15

But didn't use iso code.

rchl:
Code looks fine to me. Please provide reproducion on https://codesandbox.io because I don't understand what problem do you see here.

TimonPeng:
@rchl Review please. https://codesandbox.io/s/codesandbox-nuxt-rng17

When don't set a cookie, it always choose English not depending on the http header `accept-language`.

You can use the following command for environment replication.

```
curl 'https://rng17.sse.codesandbox.io/' -H 'accept-language: zh-CN,zh;q=0.9,ja;q=0.8' --compressed -L -v
```

You can see two 302 redirects happend.

- The first time is depend on `rootRedirect: "cn"`
- The second time is depend on `detectBrowserLanguage`? I think.

rchl:
In a locale defined this way:
```
      {
        name: "简体中文",
        code: "cn",
        iso: "zh-CN",
        file: "cn.js"
      }
```
The `code` property is used to match the browser language. In your case, it doesn't match because nuxt-i18n matches it against `zh` from the `accept-language` header, not `cn`. So I think the issue here is with configuration - you should change `code` property to `zh`. Same for `rootRedirect`.

rchl:
The `iso` code is only used for some SEO functionality, not for matching browser locale.

TimonPeng:
@rchl Ok, thank you.

Would you mind adding an alias configuration for router path or domain name? 

Like apple.com:

- https://www.apple.com/jp/
- https://www.apple.com/hk/
[...]

PR #965, opened 2020-11-19
fix(detectBrowserLanguage): use ISO code if available

owenandrews:
This PR updates `matchBrowserLocale` to use the ISO code for a locale when performing browser language detection. If the ISO code isn't set then the usual locale code is used, as is the current functionality.

This solves some of the issues mentioned in #499 where one may not want to use ISO codes in the URL, but does wish to make use of the browser detection and SEO features.

This is my first attempt at contributing to this project so I'm not 100% sure that the way I've gone about it is correct and I would welcome feedback on how this PR can be made better :)


src/templates/utils-common.js
```diff
@@ -25,26 +25,29 @@ export const matchBrowserLocale = (appLocales, browserLocales) => {
   /** @type {{ code: string, score: number }[]} */
   const matchedLocales = []
 
+  // Normalise appLocales input
```
rchl, line 28: The types of `appLocales` in the JSDoc comment above should be updated.

rchl (commented):
Please delete `package-lock.json` and use `yarn` in this project.
(Or keep using npm but don't commit the lock file.)

rchl:
Sorry to keep you waiting. I haven't reviewed it properly yet, will do it ASAP, when I have more time.

rchl:
This would need some new tests added around here: https://github.com/nuxt-community/i18n-module/blob/fee0e9375ef050e05ffc9a29d69fe0216fe50e03/test/module.test.js#L1166-L1192

Base config used by those tests is here: https://github.com/nuxt-community/i18n-module/blob/fee0e9375ef050e05ffc9a29d69fe0216fe50e03/test/fixture/base.config.js#L33-L44

rchl:
Actually, tests can be unit-tested here: https://github.com/nuxt-community/i18n-module/blob/6bac69c1cd3ca51f0ab675460a63be4509d19c01/test/unit.test.js#L31-L31
It's easier and faster to work with those.

owenandrews:
Thanks @rchl, I added a unit test, does it need to be comprehensive or is just the one ok in this instance?

src/templates/utils-common.js
```diff
@@ -17,34 +17,37 @@ export const parseAcceptLanguage = input => {
...
  * @param {string[]} browserLocales The locales to match against configured.
  * @return {string?}
  */
 export const matchBrowserLocale = (appLocales, browserLocales) => {
   /** @type {{ code: string, score: number }[]} */
   const matchedLocales = []
 
+  // Normalise appLocales input
+  appLocales = appLocales.map(appLocale => ({
+    code: typeof appLocale === 'string' ? appLocale : appLocale.code,
```
rchl, line 30: We don't seem to be using the `code` property so why create it?

test/unit.test.js
```diff
@@ -94,4 +94,11 @@ describe('matchBrowserLocale', () => {
 
     expect(matchBrowserLocale(appLocales, browserLocales)).toBe('en-gb')
   })
+
+  test('matches ISO locale code', () => {
+    const appLocales = [{ code: 'cn', iso: 'zh-CN' }, { code: 'en', iso: 'en-US' }]
+    const browserLocales = ['zh', 'zh-CN']
+
+    expect(matchBrowserLocale(appLocales, browserLocales)).toBe('cn')
+  })
```
rchl, line 103: Let's also add something more tricky:

```
  test('matches full ISO code', () => {
    const appLocales = [{ code: 'us', iso: 'en-US' }, { code: 'gb', iso: 'en-GB' }]
    const browserLocales = ['en-GB', 'en']

    expect(matchBrowserLocale(appLocales, browserLocales)).toBe('gb')
  })
```

EDIT: And also this one (taken from another PR that is currently (that is without this fix) failing):
```
  test('matches locale when only languages match', () => {
    const appLocales = ['en-GB', 'en-US']
    const browserLocales = ['en-IN', 'en']

    expect(matchBrowserLocale(appLocales, browserLocales)).toBe('en-GB')
  })
```

And there is also this tricky case in which none of the matches is exact:

```
  test('matches locale when only languages match', () => {
    const appLocales = ['en-CA', 'fr-CA']
    const browserLocales = ['en-UK', 'fr']

    expect(matchBrowserLocale(appLocales, browserLocales)).toBe('en-CA')
  })
```

The expected result for this one is not entirely clear but let's have it test the current behavior anyway. Maybe add a comment to that effect?

rchl (approved):
Looks good to me. Just please update with this one extra test.

rchl:
Can you also update the `options-reference.md` (https://i18n.nuxtjs.org/options-reference#locales) to mention that the `iso` property is also used for matching browser locales when using the `detectBrowserLanguage` option? If you do, please update both `en` and `es` pages (both with English text).

test/unit.test.js
rchl: Pushed myself.

rchl:
Thanks!

owenandrews:
Thanks @rchl, sorry you beat me to it.

PR #966, opened 2020-11-20
Update lang-switcher.md

flozero:
Update the documentation. In my case saying slug made sense to use it everywhere, I didn't understand in the first place it will be the dynamic parameters name


rchl:
Thanks. I'll handle it through https://github.com/nuxt-community/i18n-module/pull/967, hopefully making it more clear.

PR #981, opened 2020-12-04
refactor: replace clonedeep with klona

divine:
klona is a tiny and performant library for deep cloning.

See: https://github.com/lukeed/klona#benchmarks


rchl:
Thank you :)

PR #992, opened 2020-12-09
feat(fallbackLocale): support vue-i18n decision map fallback

DktRemiDoolaeghe:
Add some support for `vue-i18n` locale fallback decision map: https://kazupon.github.io/vue-i18n/guide/fallback.html#explicit-fallback-with-decision-maps


docs/content/en/locale-fallback.md
```diff
@@ -0,0 +1,64 @@
+---
```
rchl, line 1: We'd also need a copy of that page for the `es` locale. Can be in English of course.

docs/content/en/locale-fallback.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Locale fallback
+
+**nuxt-i18n** takes advantage of **vue-i18n** ability. It is possible to define a single fallback locale, or a decision map for more specific needs.
+
+### One fallback
+
+The simplest way is defining a single fallback, whatever the wanted locale was. In this example, **en** fallback will be chosen for any missing locale.
+
+```
```
rchl: Can you use a `js {}[nuxt.config.js]` meta info to match other files?

docs/content/en/locale-fallback.md
```diff
@@ -0,0 +1,64 @@
+---
+title: Locale fallback
+description: "How a fallback gets selected when a translation is missing"
+position: 11
+category: Guide
+---
+
+## Locale fallback
+
+**nuxt-i18n** takes advantage of **vue-i18n** ability. It is possible to define a single fallback locale, or a decision map for more specific needs.
```
rchl: I have an issue with duplicating the contents of the `vue-i18n` documentation and also with it being not as detailed as the original documentation (for example doesn't mention an array format like `['en', 'fr']`).

I'd say that we don't need to explain any details here but just show an example configuration with a link to original documentation at https://kazupon.github.io/vue-i18n/guide/fallback.html .

So I would envision something like:

---

**nuxt-i18n** takes advantage of **vue-i18n** ability to handle localization fallback. It is possible to define a single fallback locale, an array of locales, or a decision map for more specific needs.

```js {}[nuxt.config.js]
modules: [
  'nuxt-i18n'
],

i18n: { 
  vueI18n: {
    fallbackLocale: 'en',
    // or
    fallbackLocale: ['en', 'fr'],
    // or
    fallbackLocale: {
      'de-CH':   ['fr', 'it'],
      'zh-Hant': ['zh-Hans'],
      'es-CL':   ['es-AR'],
      'es':      ['en-GB'],
      'pt':      ['es-AR'],
      'default': ['en', 'da']
    }
  }
}
```

More information in [vue-i18n documentation](https://kazupon.github.io/vue-i18n/guide/fallback.html).

---

docs/content/en/locale-fallback.md
```diff
@@ -0,0 +1,64 @@
+---
+title: Locale fallback
+description: "How a fallback gets selected when a translation is missing"
+position: 11
+category: Guide
+---
+
+## Locale fallback
```
rchl: This is redundant and results in two identical headers on this page.

test/module.test.js
```diff
@@ -1762,3 +1762,66 @@ describe('generate with prefix strategy', () => {
     expect(canonicalLink?.getAttribute('href')).toBe('nuxt-app.localhost/en')
   })
 })
+
+describe('Configuration with locale fallback decision map', () => {
+  let nuxt
```
rchl, line 1768: Please annotate the type with:

```
  /** @type {Nuxt} */
```

like in other cases.

test/module.test.js
```diff
@@ -1762,3 +1762,66 @@ describe('generate with prefix strategy', () => {
...
+  })
+
+  test('should be fallback translation by respecting the decision map', async () => {
+    const html = await get('/fr-FR/fallback')
+    const dom = getDom(html)
+
+    for (let level = 0; level <= 3; level++) {
+      const translationNode = dom.querySelector('[data-test="level' + level + '"]')
+      expect(translationNode).not.toBeNull()
+      expect(translationNode.textContent).toEqual('level ' + level)
```
rchl: I would prefer to remove the assert above and just do this here:

```js
      expect(translationNode?.textContent).toEqual('level ' + level)
```

since otherwise type-checking complains about `translationNode` being potentially null.

src/templates/plugin.main.js
```diff
@@ -90,10 +90,22 @@ export default async (context) => {
...
       const { loadLanguageAsync } = require('./utils')
-
-      // Load fallback locale.
-      if (app.i18n.fallbackLocale && newLocale !== app.i18n.fallbackLocale) {
-        await loadLanguageAsync(context, app.i18n.fallbackLocale)
+      const i18nFallbackLocale = app.i18n.fallbackLocale
+
+      // Load fallback locale(s).
+      if (i18nFallbackLocale) {
+        if (typeof i18nFallbackLocale === 'object' && i18nFallbackLocale[newLocale]) {
```
rchl: `typeof []` is also an `object`.

Since we should also support an array of strings, I think the logic should be:

```js
        if (Array.isArray(i18nFallbackLocale)) {
          await Promise.all(i18nFallbackLocale.map(locale => await loadLanguageAsync(context, locale)))
        } else if (typeof i18nFallbackLocale === 'object' && i18nFallbackLocale[newLocale]) {
```

Also, I'm wondering if we should explicitly avoid loading the currently loaded locale. It seems like we will do an extra `app.i18n.setLocaleMessage` call in that case so it would be better to avoid it.

src/templates/plugin.main.js
```diff
@@ -90,10 +90,22 @@ export default async (context) => {
...
+          let localesToLoadPromises = []
+          if (i18nFallbackLocale[newLocale]) {
```
rchl, line 101: `i18nFallbackLocale[newLocale]` is already checked above so we can just map right away.

rchl (changes requested):

docs/content/en/locale-fallback.md
DktRemiDoolaeghe: Job's done

docs/content/en/locale-fallback.md
DktRemiDoolaeghe: Job's done

src/templates/plugin.main.js
DktRemiDoolaeghe: Job's done

src/templates/plugin.main.js
DktRemiDoolaeghe: Ok, I've fixed the Array/Object question. Actually, I had forgotten the Array case. I've added a new unit test and the business code to make it green.

For the extra call to `app.i18n.setLocaleMessage`, I think this won't happen (but I may be wrong), as there is an `if` that skips the loading if already loaded:

```js
if (!app.i18n.loadedLanguages.includes(locale)) {
  ...
}
```

Or did I miss something?

test/module.test.js
DktRemiDoolaeghe: Job's done

test/module.test.js
DktRemiDoolaeghe: Job's done

src/templates/plugin.main.js
rchl: > For the extra call to `app.i18n.setLocaleMessage`, I think this won't happen (but I may be wrong), as there is an `if` that skips the loading if already loaded:

Yes but for the initial (SSR) request, when the client-side run this code, the languages will be available through NUXT_STATE but not present in `loadedLanguages`. So then there will be some redundant churn, I believe.

Maybe the original code should be more consistent and set `loadedLanguages` in this case, without actually triggering loading.

docs/content/en/locale-fallback.md
DktRemiDoolaeghe: Ok. I had a doubt on how to write a documentation correctly. Thanks for this fix, I have pushed it.

docs/content/en/locale-fallback.md
DktRemiDoolaeghe: Job's done

docs/content/en/locale-fallback.md
```diff
@@ -0,0 +1,33 @@
+---
+title: Locale fallback description: "How a fallback gets selected when a translation is missing"
```
rchl:
```
 WARN  Could not parse /Users/nuxt-i18n/docs/content/en/locale-fallback.md: incomplete explicit mapping pair; a key node is missed; or followed by a non-tabulated empty line at line 2, column 35
    title: Locale fallback description: "How a fallback gets selected  ... 
                                      ^
```

src/templates/plugin.main.js
```diff
@@ -90,10 +90,22 @@ export default async (context) => {
...
-      if (app.i18n.fallbackLocale && newLocale !== app.i18n.fallbackLocale) {
-        await loadLanguageAsync(context, app.i18n.fallbackLocale)
+      const i18nFallbackLocale = app.i18n.fallbackLocale
+
+      // Load fallback locale(s).
+      if (i18nFallbackLocale) {
+        let localesToLoadPromises = []
+        if (Array.isArray(i18nFallbackLocale)) {
+          localesToLoadPromises = i18nFallbackLocale.map(fbLocale => loadLanguageAsync(context, fbLocale))
+        } else if (typeof i18nFallbackLocale === 'object' && i18nFallbackLocale[newLocale]) {
```
rchl: This logic seems wrong?

If using the object notation and there is no explicit fallback for `newLocale` then we should check `i18nFallbackLocale.default`, not fallback to "string" notation below.

test/module.test.js
```diff
@@ -1762,3 +1762,92 @@ describe('generate with prefix strategy', () => {
...
+          fallbackLocale: {
+            'fr-FR': ['en-GB', 'es'],
+            default: ['de']
+          }
+        }
+      }
+    }
+
+    const localConfig = loadConfig(__dirname, 'locale-decision-map-fallback', override, { merge: true })
+    nuxt = (await setup(localConfig)).nuxt
```
rchl: I'd prefer to follow the existing convention and have fixture loading in `beforeAll` hooks. Then you just need to have separate `describe` blocks for your tests.

This is better because it allows adding more tests in those describe blocks without having to re-build the fixture for each test.

src/templates/plugin.main.js
DktRemiDoolaeghe: I must admit I have absolutely no knowledge of the internals of Nuxt.  Do you think this point should be included in this PR that is about fallback, or should it be part of another issue? I'm not even sure about the differences between running the code in SSR and in CSR and its subtilities in `nuxt-i18n`.

rchl: You can ignore it in this PR. I'll have a look after this is merged.

src/templates/plugin.main.js
DktRemiDoolaeghe: Not sure to understand. If we are in an object notation, and there is no default defined, then nothing more happens. The following else is related to the condition tree (array, object, string), no to the `if` about the default.

rchl: The code is:

```js
        } else if (typeof i18nFallbackLocale === 'object' && i18nFallbackLocale[newLocale]) {
          // ...
        } else if (newLocale !== i18nFallbackLocale) {
          localesToLoadPromises.push(loadLanguageAsync(context, i18nFallbackLocale))
        }
```

So if `i18nFallbackLocale[newLocale]` is falsey, it will go to the next `if` condition which will be true because `newLocale !== i18nFallbackLocale` and will pass the whole fallback object to `loadLanguageAsync`.

docs/content/en/locale-fallback.md
DktRemiDoolaeghe: Fixed, sorry that I missed it

test/module.test.js
DktRemiDoolaeghe: You're right, I've fixed it.

docs/content/en/locale-fallback.md
```diff
@@ -1,13 +1,15 @@
...
+category: Guide
 ---
 
 
-nuxt-i18n takes advantage of vue-i18n ability to handle localization fallback. It is possible to define a single fallback locale, an array of locales,
+**nuxt-i18n** takes advantage of **vue-i18n** ability to handle localization fallback. It is possible to define a single fallback locale, an array of locales,
 or a decision map for more specific needs.
 
-```js {}[nuxt.config.js]
+```js{}[nuxt.config.js]
```
rchl: Please add back a space.

DktRemiDoolaeghe: Are you sure? I've seen no reference with a space. All these occurrences have no space.

src/templates/plugin.main.js
DktRemiDoolaeghe: You're right. I've messed up the condition. I've changed it.

I have refactored the tests too, and added another one. Don't hesitate to tell be anything wrong.

docs/content/en/locale-fallback.md
rchl: There should be a space after `js` as otherwise, some editors won't recognize the block as using JS syntax. It will work either way for docs.

Can also skip `{}`.

DktRemiDoolaeghe: Do you want me to replace them all with `js [nuxt.config.js]`? It could be the occasion.

rchl: Feel free if you don't mind. :)

DktRemiDoolaeghe: No problem :) But whoops, makes a ton of conflict. If you prefer that I remove the commit, don't hesitate

rchl: Oh crap. I forgot that I've done that yesterday myself. I have too many repos to keep track of. ;)

Maybe just revert your commit then and just apply in the ones you've added.

DktRemiDoolaeghe: Job's done :)

rchl:
Thanks for implementing.

PR #1005, opened 2020-12-29
docs: fix i18n module declaration docs

Aierie:
module and its options should be enclosed within an array: 

https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-modules/


rchl:
I've taken the liberty to refactor it a bit more for hopefully even more clarity.

PR #1009, opened 2021-01-06
docs: fix two broken links in options-reference

iBlueDust:
Two links in [/options-reference](https://i18n.nuxtjs.org/options-reference/) were broken in the following ways:

**First link** in [this section](https://i18n.nuxtjs.org/options-reference/#locales) was to [`/options-reference/#detectbrowserlanguage`](https://i18n.nuxtjs.org/options-reference/#detectbrowserlanguage), but when clicked, did nothing. The link initially was to `#detectBrowserLanguage`, but the capitalization was enough to break the link. To fix this, it was changed to `#detectbrowserlanguage`.

**Second link** in [this section](https://i18n.nuxtjs.org/options-reference/#detectbrowserlanguage) was to [`/browser-language-detection`](https://i18n.nuxtjs.org/browser-language-detection/) but kept routing to [`/options-reference/browser-language-detection`](https://i18n.nuxtjs.org/options-reference/browser-language-detection), an invalid route (returns 404).

### DIFF
--- a/docs/content/en/options-reference.md
+++ b/docs/content/en/options-reference.md
@@ -62,7 +62,7 @@ List of locales supported by your app. Can either be an array of codes (`['en', 
 
 When using an object form, the properties can be:
 - `code` (**required**) - unique identifier of the locale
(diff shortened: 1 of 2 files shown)


rchl (approved):
Thanks!

PR #1010, opened 2021-01-06
fix: When the trailing slash option is the default, don't remove it on the error page.

ogawatti:
For example, when hosting with AWS Amplify (S3), if index.html does not exist, it will be redirected to the path with the trailing slash, resulting in a redirect loop.

The following settings will remove the trailing slash when a non-existent page is accessed, which should not be removed since the default Trailing Slash option does not redirect.

### Nuxt configuration

* `target` : static
* `router.trailingSlash` : undefined (**default**)

### Nuxt-i18n configuration
```
i18n: {
    strategy: 'prefix_and_default',
    lazy: true,
    ...
}
```

### DIFF
--- a/src/templates/plugin.routing.js
+++ b/src/templates/plugin.routing.js
@@ -62,7 +62,9 @@ function localeRoute (route, locale) {
         !i18n.differentDomains
 
     let path = (isPrefixed ? `/${locale}${route.path}` : route.path)
-    path = path.replace(/\/+$/, '') + (trailingSlash ? '/' : '') || '/'
+    if (trailingSlash !== undefined) {
+      path = path.replace(/\/+$/, '') + (trailingSlash ? '/' : '') || '/'
+    }
     localizedRoute.path = path
   } else {
     if (!route.name && !route.path) {
--- a/test/browser.test.js
+++ b/test/browser.test.js
@@ -965,7 +965,7 @@ describe(`${browserString} (onlyOnRoot + prefix)`, () => {
     const page = await browser.newPage()
     await page.goto(url('/'))
     expect(await (await page.$('body'))?.textContent()).toContain('locale: en')
-    expect(await getRouteFullPath(page)).toBe('/en')
+    expect(await getRouteFullPath(page)).toBe('/en/')
   })
 })
 
--- a/test/module.test.js
+++ b/test/module.test.js
@@ -921,6 +921,7 @@ describe('prefix_and_default strategy', () => {
     // Prefer unprefixed path for default locale:
     expect(window.$nuxt.localeRoute('/simple', 'en')).toMatchObject({ name: 'simple___en___default', fullPath: '/simple' })
     expect(window.$nuxt.localeRoute('/simple', 'fr')).toMatchObject({ name: 'simple___fr', fullPath: '/fr/simple' })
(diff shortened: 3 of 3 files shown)


src/templates/plugin.routing.js
```diff
@@ -62,7 +62,9 @@ function localeRoute (route, locale) {
         !i18n.differentDomains
 
     let path = (isPrefixed ? `/${locale}${route.path}` : route.path)
-    path = path.replace(/\/+$/, '') + (trailingSlash ? '/' : '') || '/'
+    if (trailingSlash !== undefined) {
+      path = path.replace(/\/+$/, '') + (trailingSlash ? '/' : '') || '/'
+    }
```
rchl, line 67: It's honestly quite hard for me to judge whether that change makes sense or not.

The fact that it fixes your issue seems somewhat coincidental to me since with explicit `trailingSlash=false` it would still fail.

You could say that when `trailingSlash` is undefined we should not do any trailing slash changes but then we are adding the locale prefix before that line so that already has to make a choice whether the trailing slash is added or not after locale. Currently, it adds a slash and that's what happens to be solving your issue but what it does now seems somewhat arbitrary since the idea was that the trailing slash will be "normalized" by the line you've changed.

Also with your change, `prefix` strategy and `target: 'server'` we'll redirect from `/` to `/en/` while at the same time we'll redirect from `/fr` to `/en`. So the trailing slash will be inconsistent.

So I'm not entirely convinced here... It's probably OK to make that change but it has the potential for breaking some cases also...

ogawatti: If you set trailingSlash to false and set it to S3 delivery, you will get a redirect loop.
That is correct from the Nuxt settings and S3 behavior.

However, as you said, going from `/` to `/en/` is not intended.
I would like to limit it to the case of an error, as the scope of influence is large, but I don't have a good idea for that ...

supachailllpay:
I ran into a similar problem with this one when using the function `localePath` with default nuxt `trailingSlash` option (undefined).

In my case I have to use both the path with trailing slash and without trailing slash which is seems impossible with current `localePath`.

Example:

```html
<nuxt-link to='/a'>...</nuxt-link>                 // => href=/a
<nuxt-link to='/b/'>...</nuxt-link>                // => href=/b/
<nuxt-link :to='localePath("/a")'>...</nuxt-link>  // => href=/en/a
<nuxt-link :to='localePath("/b/")'>...</nuxt-link> // => href=/en/b
```

I'm not sure how to fix this but I think we should respect the behavior of this `trailingSlash` option in the same way as nuxt do. To have a way that isn't just always append trailing slash or always remove it.

rchl:
I'm starting to realize that for static hosting, setting `route.trailingSlash` to `true` might be required and the module should either enforce that or strongly recommend it at least (see also #1037).

> In my case I have to use both the path with trailing slash and without trailing slash which is seems impossible with current `localePath`.

I'm curious why do you **need** to use paths with trailing slash and whether that's because you are statically hosting your generated code?

supachailllpay:
No, In my case it's a bit messy. I use universal mode run behind reverse proxy server which serve multiple applications. Some route that created by CMS may have a trailing slash. So I am concerned about this.

```html
<a :href='localePath("/a/")'>Go to another application</a>
```

I think we should at least have an option that can preserve trailing slash.

PR #1016, opened 2021-01-08
feat: expose localeProperties property

iifawzi:
Hi, as mentioned here https://github.com/nuxt-community/i18n-module/issues/916, @rchl decided to expose the current locale object, here's how i tackled it,  i hope i did it in the correct way since this's my first contribution!   

i think we can move the codeFromLocale function from seo-head to utils-common to use it here also, instead of duplication, if confirmed i will commit the change

### DIFF
--- a/docs/content/en/options-reference.md
+++ b/docs/content/en/options-reference.md
@@ -69,11 +69,7 @@ When using an object form, the properties can be:
 - `domain` (required when using `differentDomains`) - the domain name you'd like to use for that locale (including the port if used)
 - `...` - any custom property set on the object will be exposed at runtime. This can be used, for example, to define the language name for the purpose of using it in a language selector on the page.
 
-When using an array of objects, you can access the current locale properties using the `localeProperties` property. 
-  ```js
-  { code: 'en', iso: 'en-US', file: 'en.js' }
-```
-  - When using an array of codes, it will be set to an empty object.
+You can access all the properties of the current locale through the `localeProperties` property. When using an array of codes, it will only include the `code` property.
 
 ## `defaultLocale`
 
--- a/docs/content/en/options-reference.md
+++ b/docs/content/en/options-reference.md
@@ -69,6 +69,19 @@ When using an object form, the properties can be:
 - `domain` (required when using `differentDomains`) - the domain name you'd like to use for that locale (including the port if used)
 - `...` - any custom property set on the object will be exposed at runtime. This can be used, for example, to define the language name for the purpose of using it in a language selector on the page.
 
+## `localeProperties`
+
+- type: `object`
+- default: `{}`
+
+Object contains the properties of the current `Locale` as defined in the `Locales` array.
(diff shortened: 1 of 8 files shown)


rchl:
- The `locales` property can also be just an array of locales in which case your code will crash. If that's the case we can just set an empty object.
- Please initialize the new property also around line 263
- We'd need to add it to the documentation. In `docs/content/en/api.md` (and `es` version) for sure but ideally also mention it in the documentation of the `locales` option in `options-reference.md`.
- At least a single test that checks that this property exists and is set to the correct value should be added.

iifawzi:
Hi @rchl, i made all the mentioned points, changes are committed. 
i added tow tests to ensure that the property exists and is set to the correct value, when configuration is empty, and when there're an array of locales objects.

iifawzi:
Hi @rchl, i made all the mentioned points, changes are committed.
i added two tests to ensure that the property exists and is set to the correct value, when configuration is empty, and when array of locale objects is used.

docs/content/en/options-reference.md
```diff
@@ -69,6 +69,19 @@ When using an object form, the properties can be:
 - `domain` (required when using `differentDomains`) - the domain name you'd like to use for that locale (including the port if used)
 - `...` - any custom property set on the object will be exposed at runtime. This can be used, for example, to define the language name for the purpose of using it in a language selector on the page.
 
+## `localeProperties`
```
rchl: This is not an option itself so this shouldn't be listed on its own. Should be just mentioned as part of the `locales` option.

test/module.test.js
```diff
@@ -1960,3 +1965,25 @@ describe('Locale fallback array', () => {
...
+  })
+
+  test('localeProperties object exists and is set to the correct value', async () => {
+    const window = await nuxt.renderAndGetWindow(url('/'))
+    expect(window.$nuxt.$i18n.localeProperties).toEqual({
+      code: 'en',
+      iso: 'en',
+      name: 'English'
+    })
+  })
```
rchl: Can we attach this test to one of the existing describe blocks? Building a new fixture is time-consuming.

rchl (changes requested):

docs/content/en/options-reference.md
iifawzi: my bad, i committed the changes, idk if it's obvious right now or not, i added it to the bottom of the locales option.

test/module.test.js
iifawzi: i just can't find any appropriate describe block to attach it, what do you think?

iifawzi: update: i attached it to the `hreflang` block, all changes are committed.

rchl (approved):
Looks good from my side. I've done some small changes, please have a look and let me know if those look good to you.

iifawzi:
looks awesome from my side, well done!

PR #1021, opened 2021-01-11
docs: update lazy-load-translations.md

defaye:
The following changes are required for lazy loading to work properly.

### DIFF
--- a/docs/content/en/lazy-load-translations.md
+++ b/docs/content/en/lazy-load-translations.md
@@ -46,6 +46,7 @@ Configuration example:
   ],
   lazy: true,
   langDir: 'lang/',
+  defaultLocale: 'en',
   vueI18n: {
     fallbackLocale: 'en'
   }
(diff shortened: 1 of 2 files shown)


docs/content/en/lazy-load-translations.md
```diff
@@ -46,15 +46,17 @@ Configuration example:
   ],
   lazy: true,
   langDir: 'lang/',
-  defaultLocale: 'en'
+  vueI18n: {
+    defaultLocale: 'en'
+  }
```
rchl: This is not correct. `vue-i18n` doesn't support a `defaultLocale` option.

rchl:
Whatever changes are done in `en` have to be duplicated in `es` documentation also.

docs/content/en/lazy-load-translations.md
defaye: You're right. Sorry. This was meant to be a reference to `fallbackLocale`. I think that without it, it wasn't working for me when using lazy-loading.

defaye: Actually, I think it was working fine.  It seems that something else was at play, as some kind of caching possibly lead me to believe this was the critical change, but I've left the changes for the missing return statement.

PR #1022, opened 2021-01-11
feat: expose API for detecting browser locale

### LINKED ISSUE #1018: Revert to detecting browser's language after manually setting a locale?
Hello everyone,

In my application I'd like to have my locale switch have options like this: "English", "French", "Spanish"... and "Auto" which would revert the application to the default i18n state, where it detects the user's browser's locale.

I've tried using `seLocale()` or `setLocale('')` but without luck.

Is it even possible without deleting the cookie and refreshing? Thanks!

ThibaultNocchi:
I've exposed a method to get the browser detected locale from nuxt-i18n. This allows me to reset a custom set locale in my application (for instance when logging out). If I used `setLocaleCookie('')` I then needed to refresh the page to apply this "reset" (as intended ofc). So this way I can fetch this method and apply the found locale with `setLocale()`.

fix #1018

### DIFF
--- a/docs/content/en/api.md
+++ b/docs/content/en/api.md
@@ -99,6 +99,14 @@ Instance of [VueI18n class](http://kazupon.github.io/vue-i18n/api/#vuei18n-class
 
   Switches locale of the app to specified locale code. If `useCookie` option is enabled, locale cookie will be updated with new value. If prefixes are enabled (`strategy` other than `no_prefix`), will navigate to new locale's route.
 
+#### getBrowserLocale
+
+  - **Arguments**:
+    - no arguments
+  - **Returns**: `string | undefined`
+
+  Returns browser locale code filtered against the ones defined in options.
+
 ### Properties
 
 #### defaultLocale
@@ -112,7 +120,7 @@ Instance of [VueI18n class](http://kazupon.github.io/vue-i18n/api/#vuei18n-class
   - **Type**: `Array<string | LocaleObject>`
 
   List of locales as defined in options.
-  
+
 #### localeProperties
 
   - **Type**: `LocaleObject`
--- a/src/templates/plugin.main.js
+++ b/src/templates/plugin.main.js
@@ -200,6 +200,17 @@ export default async (context) => {
     return [null, null]
   }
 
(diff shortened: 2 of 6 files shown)


src/templates/plugin.main.js
```diff
@@ -200,6 +200,17 @@ export default async (context) => {
...
 
+  const getBrowserLocale = () => {
+    if (process.client && typeof navigator !== 'undefined' && navigator.languages) {
+      // Get browser language either from navigator if running on client side, or from the headers
+      return matchBrowserLocale(locales, navigator.languages)
+    } else if (req && typeof req.headers['accept-language'] !== 'undefined') {
+      return matchBrowserLocale(locales, parseAcceptLanguage(req.headers['accept-language']))
+    } else {
+      return undefined
+    }
```
rchl, line 211: Looks good but can we change the code so that only `string | undefined` types are possible (change `matchBrowserLocale` to return `undefined` instead of `null`)?

docs/content/en/api.md
```diff
@@ -99,6 +99,14 @@ Instance of [VueI18n class](http://kazupon.github.io/vue-i18n/api/#vuei18n-class
...
   Switches locale of the app to specified locale code. If `useCookie` option is enabled, locale cookie will be updated with new value. If prefixes are enabled (`strategy` other than `no_prefix`), will navigate to new locale's route.
 
+#### getBrowserLocale
+
+  - **Arguments**:
+    - no arguments
+  - **Returns**: `string | undefined | null`
+
+  Returns browser locale code filtered against the ones defined in options.
+
```
rchl: Please also update `es/api.md` in same way.

src/templates/plugin.main.js
ThibaultNocchi: I've done it this way, `matchBrowserLocale` now returns `undefined` and I've updated the tests. :)

docs/content/en/api.md
ThibaultNocchi: Done thanks to @ferferga!

rchl:
Thanks!

ThibaultNocchi:
Thank you too! :)

Issue #762, opened 2020-06-10
[Feature Request] Add dir attribute for each locale


HapLifeMan:
### Is your feature request related to a problem? Please describe.
I'm using this package to translate my app, and now I'm adding an RTL locale.
It would be great if nuxt-i18n could manage the body dir attribute.

### Describe the solution you'd like
I propose to add a `dir` property for each locale in the module (values would be `rtl`, `ltr` or `auto` [by default](https://developer.mozilla.org/fr/docs/Web/HTML/Attributs_universels/dir)). For example:

```js
// nuxt.config.js, inside nuxt-i18n module

locales: [
  { code: 'ar', iso: 'ar', file: 'ar/app.js', dir: 'rtl' },
  { code: 'en', iso: 'en-US', file: 'en/app.js', dir: 'ltr' },
  { code: 'fr', iso: 'fr-FR', file: 'fr/app.js', dir: 'ltr' },
],
```

Then this would be added in the HTML by setting a new `dir` attribute at a very top level of the document, for example in the `<html>` or `<body>` tag. I think it's a mandatory feature for an i18n repo 😊

### Describe alternatives you've considered
As for now I'm using a custom plugin and adding the dir property as soon as I can:

```js
// plugins/ltr-rtl.js

export default function({ app }, inject) {
  const dir = () => app.i18n.locales.find((x) => x.code === app.i18n.locale)?.dir;
  inject( 'dir', dir);
}
```

```vue
<!-- layouts/default.vue -->

<div id="app" :dir="$dir()">
  My app here...
</div>
```

What do you think? 😊

majdal:
RTL always has so little support, so I'm happy to see it being brought up! 

I have a similar issue that I'm trying to resolve. I tried two different solutions, and both are unsatisfactory. I'm still fairly new to Nuxt, so I'm not 100% sure this is the right approach. It would be good to create a more "standard" way of solving the rtl problem.

```js
export default function({ app }) {
  // beforeLanguageSwitch called right before setting a new locale
  app.i18n.beforeLanguageSwitch = (oldLocale, newLocale) => {
    app.head.htmlAttrs = {
      lang: newLocale,
      dir: newLocale == "ar" ? "rtl" : "ltr"
    }
  }
}
  ```
Nevertheless, it would be great to have a `$dir` property to use where needed.

HapLifeMan:
Hey @rch, I would really love to have your opinion 😊

rchl:
[...]

Issue #984, opened 2020-12-06
Store a map for customized paths


rchl:
### Is your feature request related to a problem? Please describe.

To solve issues like #641, #710, #735, #1254 and #1177, we would need to create and expose a list of customized paths. The knowledge of which paths are customized would allow short-cutting some cases or handle them in a special way.

### Describe the solution you'd like

Internally there should be a list of customized paths in a format that is quick to look up and be as compact as possible, size-wise.

### Describe alternatives you've considered

No alternative.

### Additional context

This concerns the `pages` option and custom component options specified through the `nuxtI18n` option.

divine:
Hello,

@rchl is using something like [flexsearch](https://github.com/nextapps-de/flexsearch#store) would be the best solution to store/search paths?

Thanks!

rchl:
At first sight it doesn't seem to be super fitting as I don't think we'll need full text capabilities like that here. We just need to be able to lookup route.name => route.path and route.path => route.name. At least as far as I remember this issue...

divine:
> At first sight it doesn't seem to be super fitting as I don't think we'll need full text capabilities like that here. 

There is a light version as well which is stripped down version (flexsearch.light.js 2.9 kb)

> We just need to be able to lookup route.name => route.path and route.path => route.name. At least as far as I remember this issue...

And custom settings for that route as well (in component option).

Thanks!

rchl:
#1254 concerns dynamic routes. Would need to investigate how would that have to be handled even when having a mapping of sorts.

divine:
@rchl I'd gladly help but not sure where to start and what needs to be done. Can you give you some ideas?

I will try my best.

Thanks!

rchl:
I've started working on it on the https://github.com/nuxt-community/i18n-module/commits/fix/no-prefix-rename branch but I remember leaving it in a pretty unfinished state.

I think there is quite a lot left to do but I won't give you any pointers as I don't remember much now. I'm not even sure if the approach I was experimenting with is even good and would solve all the issues...

BobbieGoede:
Closing this as v7 is not being actively worked on (critical hotfixes only), if this still applies to v8 and later please open a new issue!

PR #1023, opened 2021-01-12
feat: add dir property and defaultDirection option

iifawzi:
i've made some changes to manage the case when the seo is enabled, which required changing the name of `plugin.seo.js` for clarity. Current scenario is: if the dir property is defined it will be used in the html's dir attribute, if not, or the locales form is an array of codes, the dir property will be set to undefined and the attribute will not be added. 

discussion: #762


iifawzi:
Hey @rchl, can you review this if you got a time?

rchl:
Yes, sorry for the delay. Will do ASAP.

src/templates/seo-head.js
```diff
@@ -11,14 +12,15 @@ import {
...
+    !seo ||
     !(VueMeta.hasMetaInfo ? VueMeta.hasMetaInfo(this) : this._hasMetaInfo) ||
     !this.$i18n ||
     !this.$i18n.locale ||
     !this.$i18n.locales ||
     this.$options[COMPONENT_OPTIONS_KEY] === false ||
     (this.$options[COMPONENT_OPTIONS_KEY] && this.$options[COMPONENT_OPTIONS_KEY].seo === false)
   ) {
-    return {}
+    return { htmlAttrs: {} }
```
rchl: Does this change have a specific purpose?

src/templates/head-meta.js
```diff
@@ -0,0 +1,24 @@
...
+
+  let metaObject
+  if (seo) {
+    metaObject = nuxtI18nSeo.call(this)
+  } else {
+    metaObject = { htmlAttrs: {} }
+  }
+
+  const currentLocale = this.$i18n.localeProperties || {}
+  const currentLocaleDir = currentLocale[LOCALE_DIR_KEY]
```
rchl: Can drop the first line since `localeProperties` should be always defined.

rchl:
What I really don't like about this is that it uses a global mixin. It causes unnecessary work for both Vue and Vue-meta and is the reason why `seo: true` is not recommended in the documentation. Even though this logic is not as heavy as SEO logic, it still causes unnecessary processing for both Vue and vue-meta.

I would ask you to investigate https://vue-meta.nuxtjs.org/api/#meta-addapp instead. I was meaning to look into it for the SEO functionality but it would be nice to try it first on a smaller feature like this one. Judging from the documentation it should be pretty easy to register an app with it and then update it whenever the locale changes.

src/templates/seo-head.js
iifawzi: since in the `headMeta` function, if the seo is enabled, we have to ensure that the returned object has the `htmlAttrs` property. 

```JS
 let metaObject
  if (seo) {
    metaObject = nuxtI18nSeo.call(this)
  } else {
    metaObject = { htmlAttrs: {} }
  }
```

but yeah, it seems a little bit ambiguous this way, i think it might be better to revert back and return an empty object as before. And instead, check the existence of the property in the `headMeta` as following maybe: 

```JS
  let metaObject = {htmlAttrs: {}}
  if (seo) {
    const seoMeta = nuxtI18nSeo.call(this)
    if (Object.keys(seoMeta).length !== 0){
      metaObject = seoMeta;
    }
  }
```
what do you think?

src/templates/head-meta.js
iifawzi: when dropping it, the tests depends on `loader` fixture or `loader-yaml` will fail and it leads to crash `can't find dir property of undefined`, i investigated more and i've noticed something weird, `localeProperties` is not defined **ONLY** in the components where `<i18n>` component is used and the `vueI18nLoader` property is set to true for sure. 

i don't know how/why this happening, but if it can be considered as a bug, It needs to be fixed in general, since `localeProperties` will be undefined when used in those components.

iifawzi:
> What I really don't like about this is that it uses a global mixin. It causes unnecessary work for both Vue and Vue-meta and is the reason why `seo: true` is not recommended in the documentation. Even though this logic is not as heavy as SEO logic, it still causes unnecessary processing for both Vue and vue-meta.
> 
> I would ask you to investigate https://vue-meta.nuxtjs.org/api/#meta-addapp instead. I was meaning to look into it for the SEO functionality but it would be nice to try it first on a smaller feature like this one. Judging from the documentation it should be pretty easy to register an app with it and then update it whenever the locale changes.

make sense. i just implemented it this way to follow the convention used with SEO functionality. I will take a look on the vue-meta later today.

iifawzi:
as per docs, if we used `vue-meta` we will have to inject a function as same as the recommended solution for SEO, since `$meta()` can only be accessed through a component. is it the intended approach?

src/templates/seo-head.js
rchl: For the least amount of duplication maybe:

```js
  let metaObject = {}
  if (seo) {
    metaObject = nuxtI18nSeo.call(this)
  }
  if (!metaObject.htmlAttrs) {
     metaObject.htmlAttrs = {};
  }
```

But if you address my other comment then maybe this whole file won't be necessary.

src/templates/head-meta.js
rchl: That means that we're missing an assignment in `extendVueI18nInstance` that would copy that property on creating a new `vue-i18n` instance.

rchl (commented):
> as per docs, if we used `vue-meta` we will have to inject a function as same as the recommended solution for SEO, since `$meta()` can only be accessed through a component. is it the intended approach?

In theory it can be accessed from the plugin's code also with something like:

```
  const metaApp = Vue.prototype.$meta().addApp('i18n')
```

but I'm not sure if this is OK conceptually or more like a hack.

@pimlie maybe you could chime in here? Is that an OK way to add a VueMeta App? If not, maybe there should be some official way to access meta `$meta` from a Nuxt plugin?

src/templates/seo-head.js
iifawzi: looks much better. Yes it won't be necessary at all. **if** we decided later to go with this approach, i will then commit the changes.

src/templates/head-meta.js
iifawzi: copying? can u elaborate more? i will create a separate PR for this. 

i was investigating earlier today, and it worked when i assigned the property this way, but i think it's not the correct way, not sure. 

```JS
 const extendVueI18nInstance = i18n => {
    i18n.localeProperties = klona(locales.find(l => l[LOCALE_CODE_KEY] === i18n.locale) || { code: '' })
  }
```

rchl: Yes, something like that looks good.

`vue-i18n` creates a copied local instance of itself when a local `<i18n>` component is used. It copies all properties it knows about by itself but the Nuxt-specific ones have to be copied (or re-created) from that `extendVueI18nInstance` listener.

iifawzi: Awesome, thanks for the clarification.

iifawzi:
> > as per docs, if we used `vue-meta` we will have to inject a function as same as the recommended solution for SEO, since `$meta()` can only be accessed through a component. is it the intended approach?
> 
> In theory it can be accessed from the plugin's code also with something like:
> 
> ```
>   const metaApp = Vue.prototype.$meta().addApp('i18n')
> ```
> 
> but I'm not sure if this is OK conceptually or more like a hack.
> 
> @pimlie maybe you could chime in here? Is that an OK way to add a VueMeta App? If not, maybe there should be some official way to access meta `$meta` from a Nuxt plugin?

it will work, but we will not be able to **re-set** the meta when locale changes (from client side), or this at least what i noticed. 

```JS
import Vue from 'vue'
export const setHeadMeta = function (context) {
  const {app} = context;
  const metaApp = Vue.prototype.$meta().addApp("i18n");
  const currentDir = app.i18n.localeProperties.dir || undefined;
  metaApp.set({
    htmlAttrs: {
      dir: currentDir,
    }
  })
}
```

and get called when locale changes:  

```JS
    app.i18n.locale = newLocale
    app.i18n.localeProperties = klona(locales.find(l => l[LOCALE_CODE_KEY] === newLocale) || { code: newLocale })
    
    setHeadMeta(context)
```

it will work only in the initial render, after locale changes it won't, i read a bit in `vue-meta` docs/code, but couldn't find a way helps with our scenario.
did i miss anything?

rchl:
It seemed to work for me when I've tested before (in a slightly different way). Did you see where it goes wrong exactly? Does it seem like a bug in vue-meta?

iifawzi:
i haven't really reached any conclusions, but it seems like  `Vue.prototype.$meta()` is not working when used after the initial render, i used some silly logs in `vue-meta` methods to figure what really happens, but nothing appears as if it's not even called.

pimlie:
> If not, maybe there should be some official way to access meta `$meta` from a Nuxt plugin?

@rchl $meta should also be available from the Nuxt Context which you have access to in Nuxt modules and plugins

rchl:
> > If not, maybe there should be some official way to access meta `$meta` from a Nuxt plugin?
> 
> @rchl $meta should also be available from the Nuxt Context which you have access to in Nuxt modules and plugins

Tried different variations including:

```
context.$meta
undefined
context.app.$meta
undefined
```

and it doesn't seem to be available. It's also not visible in `@nuxt/types`.

pimlie:
Hmm, yeah. So if you want to use addApp your best solution would probably be to get the component instance from the current page through vue-router matched components. You might have to wait for that instance to become available. But addApp is centered around Vue instances indeed and just doesnt work statically.

Alternatively you could try to overwrite `app.head` to inject/merge your own meta stuff, thats what `nuxt-jsonld` does: https://github.com/ymmooot/nuxt-jsonld/blob/master/src/mergeHead.ts

rchl:
> Alternatively you could try to overwrite `app.head` to inject/merge your own meta stuff, thats what `nuxt-jsonld` does: [https://github.com/ymmooot/nuxt-jsonld/blob/master/src/mergeHead.ts](https://github.com/ymmooot/nuxt-jsonld/blob/master/src/mergeHead.ts?rgh-link-date=2021-01-26T11%3A15%3A41Z)

That one appears to be just adding a [global mixin](https://github.com/ymmooot/nuxt-jsonld/blob/62f36a9fd274de4c8dde9261a3ed99f6f5be64be/src/index.ts#L17-L21) rather than working on `app.head`. Something I'd like to avoid.

iifawzi:
so, what's your decision for this @rchl?

rchl:
I'm still against a global mixin so the only idea I have left is exposing a function like `$nuxtI18nHead()` that the user would opt-in to use in the `head()` in the layout. That would be similar to the `$nuxtI18nSeo()` function that we already have.

In the future, in a breaking version, I would probably remove the global SEO mixin completely and require the user to use `$nuxtI18nHead()` function from within the layout which would also return SEO meta tags if `seo` is enabled. That's why I'd use the `$nuxtI18nHead()` name for it instead of the more specific `$nuxtI18nDir()`, for example.

iifawzi:
this would be much better, i will commit the required changes tomorrow morning. 
EDIT: what if someone wanna use `$nuxtI18nSeo()` and `$nuxtI18nHead()` in the same time? (before the breaking version).




also, i'm thinking about adding a `defaultDirection` option in the module options, maybe with default value `auto` or `ltr` to be used when the dir property is not specified for any locale, or in case if someone wants to use `rtl` for multi-lang or dialect and `en` for specific one, then defaultDirection would help not specifying the dir in each locale, do you agree?

rchl:
> EDIT: what if someone wanna use `$nuxtI18nSeo()` and `$nuxtI18nHead()` in the same time? (before the breaking version).

Simple destructuring like:

```
return {
  ...this.$nuxtI18nSeo(),
  ...this.$nuxtI18nHead(),
}
```
should work since there are no overlapping properties in those two.

> also, i'm thinking about adding a `defaultDirection` option in the module options,

Sure

iifawzi:
> should work since there are no overlapping properties in those two.

Since we're adding the `dir` to `htmlAtrrs` there will be an overlapping. Maybe we can get over it by adding the `dir` to `bodyAttrs` (for now at least, until the breaking version).

rchl:
> > should work since there are no overlapping properties in those two.
> 
> Since we're adding the `dir` to `htmlAtrrs` there will be an overlapping. Maybe we can get over it by adding the `dir` to `bodyAttrs` (for now at least, until the breaking version).

Why do you think there would be an overlap? The SEO functionality doesn't set anything in `htmlAtrrs`.

EDIT: No sorry, it does. Was searching for `htmlAtrrs` with a typo instead of `htmlAttrs`. Hmm...

iifawzi:
> > > should work since there are no overlapping properties in those two.
> > 
> > 
> > Since we're adding the `dir` to `htmlAtrrs` there will be an overlapping. Maybe we can get over it by adding the `dir` to `bodyAttrs` (for now at least, until the breaking version).
> 
> Why do you think there would be an overlap? The SEO functionality doesn't set anything in `htmlAtrrs`.

It seems like it sets `lang` attribute in `htmlAttrs`

rchl:
New plan:
 - intoduce `$nuxtI18nHead()`
 - make it add the `dir` attribute and SEO attributes but only when `seo` option is disabled. When `seo` option is enabled, only add the `dir` attribute
 - in the future, the `$nuxtI18nSeo()` will be deprecated/removed and the only way to add SEO attributes will be through `$nuxtI18nHead()` (with the global `seo` option controlling if it adds the SEO attributes).

EDIT: To be fair, it probably would make no difference if the SEO attributes would be added both through the `$nuxtI18nHead()` and global SEO mixin at the same time. So we could go a simpler route and not have that check at all.

iifawzi:
This's the best plan. I was thinking about all possible edge cases, it seems like with this approach we will **ONLY** not be able then to disable/enable SEO on specific pages, i checked and it's also the current behavior when using `$nuxtI18nSeo()` ,  i think it goes under the main issue #984, right?, nothing else is missing i think (tests will prove :d), and all edge cases is covered, gonna work on it after the launch.

rchl:
Yes, disabling SEO per-page is broken already - https://github.com/nuxt-community/i18n-module/issues/932
Thanks for working on that.

iifawzi:
Thank you too, couldn't be done without your guidance.

I committed the changes, and here's some points might need to be discussed: 

- i added the `addDirAttribute` argument with default value `true` to be used when **only** the SEO meta is needed. 

```JS
export const nuxtI18nHead = function (addDirAttribute = true) {
  if (addDirAttribute) {
  // add dir attribute
  }
```
if this needs to be changed to a different approach, we might allow the `DefaultDirection` option to accept `false` value.
and then the `dir` attribute wouldn't be added, but i think it will not be meaningful, for that i went with the argument approach.  

- I didn't change the SEO page in docs, since i'm not sure how can we introduce the `$nuxtI18nHead()`. Should we add it in the more performant section instead of the current way? or keeping both until the next version? It might be much better, if you got a time to introduce it in docs. 

- I tried to add a warning that `$nuxtI18nSeo()` will be deprecated/removed, but since the global mixin is getting applied in a lot of components, it seems to be annoying, therefore  i removed it.

rchl:
I've refactored a bit so that we could remove the `seo-head.js` file.
Feel free to review my commits.

iifawzi:
Way much better, i've only fixed the links to `head-meta.js`
don't we need to add a note for the new way in `SEO.md`?

rchl:
Added something to that effect but wouldn't be surprised if I've overlooked something.

iifawzi:
Awesome, looks good.

rchl:
Thanks for your great work.

melnikovdv:
Guys, thanks for everything you're doing. It's really a great tool.

By the way, you have already updated docs, and people like me are looking for `seo: false` option or nuxtI18nHead function. And only a glimpse into the sources explained what's going on)

rchl:
Released in https://github.com/nuxt-community/i18n-module/releases/tag/v6.19.0

(Also added version annotations in documentation to indicate from which version the new properties/options were added - 90a19595cd8c5288ac5cbef3cc1ac064ad52ab7d)

Issue #1039, opened 2021-01-23
Memory leak


Machine-Maker:
### Version
nuxt-i18n: 6.18.0
nuxt: 2.14.12

#### [mode](https://nuxtjs.org/guides/configuration-glossary/configuration-mode): 
 - [x] universal
 - [ ] spa

### Nuxt-i18n configuration
no configuration (aka whatever the defaults are)

### Steps to reproduce
1. `yarn create nuxt-app <name>`
2. (settings used) 
![image](https://user-images.githubusercontent.com/15055071/105567774-68c7ec00-5ce9-11eb-8dd0-fd200624ecb0.png)

3. Added `nuxt-i18n` to `buildModules` in `nuxt.config.js`
4. `yarn dev`
5. Open whatever resource manager (task manager, system monitor, etc) and find the right node task
6. navigate to localhost:3000 and refresh a bunch and watch the memory go up.

Note: I tried it w/o vuetify and it doesn't happen, and it doesn't happen with just vuetify, only when both are present.

EDIT: Use **[this](https://github.com/Machine-Maker/i18n-vuetify-test)** repo to reproduce it.
### What is Expected?
node shouldn't run out of memory.

### What is actually happening?
[NodeJS crash](https://paste.gg/p/Machine_Maker/c0b034fe69e24d429355e10643e93687)

Machine-Maker:
Ok, so it turns out that vuetify only speeds up the process significantly. I re-did the process w/o vuetify and added another page after project creation, then I just put `location.reload()` in the mounted() hook and let it run for a while, and the leak was still there.

Br4x:
Same here

divine:
Hello,

Can't reproduce, also it's 404:
> EDIT: Use **[this](https://github.com/Machine-Maker/i18n-vuetify-test)** repo to reproduce it.

Thanks!

leonardlin:
I can reproduce this also 
nuxt: 2.15.7
@nuxtjs/i18n: 7.0.3
vuetify: 2.5.5

Added 1 page with "<h1>Hi</h1>"
With postman calling the page 50x times.
You can see the response time going up and the memory going up.
At hit ~40x the out of memory exception is thrown:

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory

If I comment out i18n in nuxt.config.js the problem does not appear.

leonardlin:
[...]

PR #1040, opened 2021-01-23
fix mixin mem leak

Machine-Maker:
* https://nuxtjs.org/docs/2.x/directory-structure/plugins/#global-mixins

I think this is the solution to #1039

### DIFF
--- a/src/templates/plugin.routing.js
+++ b/src/templates/plugin.routing.js
@@ -196,14 +196,17 @@ const NuxtContextProxy = function (context, targetFunction) {
 
 const plugin = {
   install (Vue) {
-    Vue.mixin({
-      methods: {
-        localePath: VueInstanceProxy(localePath),
-        localeRoute: VueInstanceProxy(localeRoute),
-        switchLocalePath: VueInstanceProxy(switchLocalePath),
-        getRouteBaseName: VueInstanceProxy(getRouteBaseName)
-      }
-    })
+    if (!Vue.__nuxtI18nRouting__) {
+      Vue.__nuxtI18nRouting__ = true
+      Vue.mixin({
+        methods: {
+          localePath: VueInstanceProxy(localePath),
+          localeRoute: VueInstanceProxy(localeRoute),
+          switchLocalePath: VueInstanceProxy(switchLocalePath),
+          getRouteBaseName: VueInstanceProxy(getRouteBaseName)
+        }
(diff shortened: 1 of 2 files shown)


rchl:
Very nice find and very surprising to me that we have this bug and had for quite a while (maybe from the start).

It seems like Nuxt is re-evaluating the plugin file on each page reload but it doesn't clear out the state within the imported Vue object.

I'm gonna discuss it a bit more with the Nuxt team. While your change fixes it, it really shouldn't be necessary and it's not clear why it is unless you know Nuxt internal behavior. And I can see pretty much every Nuxt module/plugin developer making the same "mistake" in their code if this continues to be the default Nuxt behavior.

At least it's a good thing that this only affects the dev mode and not production...

rchl:
BTW. The https://nuxtjs.org/docs/2.x/directory-structure/plugins/#global-mixins that you've linked indeed makes a mention of having a custom property for guarding multiple mixin registrations but with plugins, the Vue itself keeps such book-keeping so it's quite unexpected that the developer would need to repeat the same logic on top of that.

And the test failures seem to indicate that that fix is possibly wrong.

Machine-Maker:
Yeah I was unsure of how to read the test results and fix the issue. I just know that whenever I use global mixins w/nuxt, I always add the custom property to prevent duplicates. And yeah, you're totally right that it only affects developer mode. And I am surprised no one has reported this before.

rchl:
The problem with tests is due to this change. The mixin added from `plugin.routing.js` is not overridden after being added once which causes problems when `nuxt-i18n` is built again with different options but the "old" mixin that sees old options is still registered.

This is a really messy situation...

rchl:
FYI: @pi0 @danielroe 

Summary:
 - If you don't add extra guards for the Vue plugins or mixins registered from a plugin then you get a leak in dev mode.
 - If you add a guard, you can end up with "outdated" mixin being used during testing as apparently `jest` doesn't fully sandbox each `describe` block.

(The most likely solution for this is probably to change something in tests to work around that...)

rchl:
It looks like @clarkdo might have made a fix for this in Nuxt: https://github.com/nuxt/nuxt.js/pull/8697

(at least for the case of registering a plugin with `Vue.use` which this module uses)

EDIT: Well, I guess that fix might not remove the mixins added through a plugin.

PR #1043, opened 2021-01-25
fix: localeProperties undefined when <i18n> component used

iifawzi:
Fixes: [#1023 (comment)](https://github.com/nuxt-community/i18n-module/pull/1023#discussion_r563148174)

This pull request includes: 
- Assigning the `localeProperties` in `extendVueI18nInstance`
- Altering the test to cover all cases, instead of adding a separate one. 
- re-phrasing the related tests a bit.


rchl:
Thank you

PR #1046, opened 2021-01-27
docs: change to correct syntax in lazy load example

m1aw:


rchl:
Can you also update it in `docs/content/es/lazy-load-translations.md`?

rchl:
Thanks

PR #1056, opened 2021-02-09
fix(docs): missing semicolon in Locales option example

qnp:


rchl:
Thanks but can you also fix the `es` example?

qnp:
Done for `en` & `es` documentation

rchl:
Thanks!

PR #1059, opened 2021-02-10
Try Nuxt-edge

rchl:


package.json
```diff
@@ -95,7 +95,7 @@
     "@babel/core": "7.12.17",
     "@babel/preset-env": "7.12.17",
     "@babel/runtime": "7.12.18",
-    "@nuxt/types": "2.15.0",
+    "@nuxt/types-edge": "latest",
```
pi0:
```suggestion
    "@nuxt/types-edge": "npm:@nuxt/types",
```

Since modules extend `@nuxt/types` directly, this package is exception

types/vue.d.ts
```diff
@@ -46,7 +46,7 @@ declare module 'vue/types/options' {
 /**
  * Extends types in Nuxt
  */
-declare module '@nuxt/types' {
+declare module '@nuxt/types-edge' {
```
pi0: see above ;)

package.json
rchl: I think you meant the opposite?

```
"@nuxt/types": "npm:@nuxt/types-edge",
```

pi0: Yes. Sorry for typo

rchl: No problem and thanks. Didn't know about such syntax even.

rchl:
Will close as I don't think there is big if any advantage in testing against nuxt-edge.

It's not like this would mean that the module would be continuously checked against every push to nuxt-edge. It would only be checked when I'm actually making some changes/PRs in this module. And those are fairly seldom.

rchl:
Since the instructions for migrating to Nuxt 3 recommend using nuxt-edge, I will revive this one.

src/helpers/routes.js
```diff
@@ -175,7 +175,7 @@ export function makeRoutes (baseRoutes, {
   if (sortRoutes) {
     try {
       // @ts-ignore
-      const { sortRoutes: sortRoutesFn } = require('@nuxt/utils')
+      const { sortRoutes: sortRoutesFn } = require('@nuxt/utils-edge')
```
rchl, line 178: This can't use `edge` variant since the code is run by the user.

PR #1064, opened 2021-02-15
fix: compatibility issue with Nuxt 2.15.0

### LINKED ISSUE #1063: Nuxt@2.15.0 error - Cannot read property 'options' of undefined
### Version
nuxt-i18n: 6.20.0
nuxt: 2.15.0

 - [X] universal

### Nuxt-i18n configuration
```
i18n: {
        locales: [...],
        defaultLocale: "en",
        vueI18nLoader: true
    }
```

### What is Expected?
Not sure where the issue lies and what could have changed since the previous version of Nuxt, but the description of the issue should be self explanatory about what is failing and where.

Error when running npm run dev for Nuxt:
image

Result of console.log(config.module.rules) above failing line;
image

hooks.js logic which is looking for vue-loader when on the screenshot above there is an absolute path not equal to but **including** vue-loader:
image

rchl:
Resolves #1063

### DIFF
--- a/src/core/hooks.js
+++ b/src/core/hooks.js
@@ -52,13 +52,13 @@ export function buildHook (moduleContainer, options) {
       if (vueLoader && vueLoader.options && vueLoader.options.loaders) {
         // vue-loader under 15.0.0
         /* istanbul ignore next */
-        vueLoader.options.loaders.i18n = '@intlify/vue-i18n-loader'
+        vueLoader.options.loaders.i18n = require.resolve('@intlify/vue-i18n-loader')
       } else {
         // vue-loader after 15.0.0
         config.module.rules.push({
           resourceQuery: /blockType=i18n/,
           type: 'javascript/auto',
-          loader: '@intlify/vue-i18n-loader'
+          loader: require.resolve('@intlify/vue-i18n-loader')
         })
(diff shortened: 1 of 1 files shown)


src/core/hooks.js
```diff
@@ -48,11 +48,11 @@ export function buildHook (moduleContainer, options) {
...
   if (options.vueI18nLoader) {
     moduleContainer.extendBuild(config => {
-      const loaders = config.module.rules.find(el => el.loader === 'vue-loader').options.loaders
-      if (loaders) {
+      const vueLoader = config.module.rules.find(el => el.loader.includes('vue-loader'))
+      if (vueLoader && vueLoader.options && vueLoader.options.loaders) {
         // vue-loader under 15.0.0
         /* istanbul ignore next */
-        loaders.i18n = '@intlify/vue-i18n-loader'
+        vueLoader.options.loaders.i18n = '@intlify/vue-i18n-loader'
```
pi0: It would be nice if you use `require.resolve` here to resovle `@intlify/vue-i18n-loader` on behalf of module (otherwise we depend on implicit `node_modules` structure)

rchl: Done.

PR #1080, opened 2021-03-01
fix: avoid using commonjs for runtime (vite support)

pi0:
While tools like webpack allow using mixed esm/cjs syntax, it will break vite. And generally esm syntax allows having more predictable code. Dependencies like `is-https` will be treeshaked since have no side-effects and for client bundle since it is unreachable.

### DIFF
--- a/src/templates/options.js
+++ b/src/templates/options.js
@@ -11,7 +11,7 @@ function stringifyValue(value) {
 
 for (const [key, value] of Object.entries(options)) {
     if (key === 'vueI18n' && typeof value === 'string') {
-%>export const <%= key %> = require('<%= value %>').default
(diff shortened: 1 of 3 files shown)


src/templates/options.js
```diff
@@ -11,7 +11,7 @@ function stringifyValue(value) {
 
 for (const [key, value] of Object.entries(options)) {
     if (key === 'vueI18n' && typeof value === 'string') {
-%>export const <%= key %> = require('<%= value %>').default
+%>export const <%= key %> = () => import('<%= value %>').then(m => m.default)
```
pi0: @rchl I couldn't find where is exactly used. We should update it to check function type and await on lazy import.

rchl: 31532e7fa74bc479704cac21ff4960fbd676bce6

PR #1084, opened 2021-03-03
refactor: add Vite compatibility when lazy loading

farnabaz:
Using the module with Nuxt Vite does not lazy load languages on client site because of limitations of dynamic import.   
 
This PR will expand dynamic import into a switch case to remove variables inside import name and automatically detects build target and change the `srcDir` alias

@pi0  What do you think about this?

### DIFF
--- a/src/templates/options.js
+++ b/src/templates/options.js
@@ -19,3 +19,9 @@ for (const [key, value] of Object.entries(options)) {
     }
 }
 %>
+
+/* <% if (options.langDir) { %> */
+export const ASYNC_LOCALES = {
+  <%= options.locales.map(l => `'${l.code}': () => import('${options.langDir}/${l.file || l.iso}' /* webpackChunkName: "lang-${l.code}" */)`).join(',\n  ') %>
+}
+/* <% } %> */
--- a/src/templates/options.js
+++ b/src/templates/options.js
@@ -22,9 +22,8 @@ for (const [key, value] of Object.entries(options)) {
 
 <% if (options.langDir) { %>
 export const ASYNC_LOCALES = {
-  <%= options.locales
-    .map(l => `'${l.file}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`)
-    .filter((l, index, self) => self.indexOf(l) === index)
-    .join(',\n  ') %>
+  <%= Array.from(
+        new Set(options.locales.map(l => `'${l.file}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`))
+      ).join(',\n  ') %>
 }
 <% } %>
--- a/src/templates/options.js
+++ b/src/templates/options.js
@@ -22,6 +22,6 @@ for (const [key, value] of Object.entries(options)) {
 
 /* <% if (options.langDir) { %> */
 export const ASYNC_LOCALES = {
-  <%= options.locales.filter(l => l.file).map(l => `'${l.code}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`).join(',\n  ') %>
+  <%= options.locales.map(l => `'${l.code}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`).join(',\n  ') %>
 }
 /* <% } %> */
--- a/src/templates/options.js
+++ b/src/templates/options.js
@@ -22,6 +22,9 @@ for (const [key, value] of Object.entries(options)) {
 
 <% if (options.langDir) { %>
 export const ASYNC_LOCALES = {
-  <%= options.locales.map(l => `'${l.file}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`).join(',\n  ') %>
+  <%= options.locales
+    .map(l => `'${l.file}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`)
+    .filter((l, index, self) => self.indexOf(l) === index)
+    .join(',\n  ') %>
 }
 <% } %>
--- a/src/templates/options.js
+++ b/src/templates/options.js
@@ -22,6 +22,6 @@ for (const [key, value] of Object.entries(options)) {
 
 /* <% if (options.langDir) { %> */
 export const ASYNC_LOCALES = {
-  <%= options.locales.map(l => `'${l.code}': () => import('../${relativeToBuild(options.langDir, l.file || l.iso)}' /* webpackChunkName: "lang-${l.code}" */)`).join(',\n  ') %>
+  <%= options.locales.filter(l => l.file).map(l => `'${l.code}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`).join(',\n  ') %>
 }
 /* <% } %> */
--- a/src/templates/options.js
+++ b/src/templates/options.js
@@ -20,8 +20,8 @@ for (const [key, value] of Object.entries(options)) {
 }
 %>
 
-/* <% if (options.langDir) { %> */
+<% if (options.langDir) { %>
 export const ASYNC_LOCALES = {
   <%= options.locales.map(l => `'${l.code}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`).join(',\n  ') %>
 }
-/* <% } %> */
+<% } %>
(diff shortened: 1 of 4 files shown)


rchl:
I'm fine with it although I would love if there wouldn't have to be a special case checking for the presence of nuxt-vite. Do you have some links that explain that `@fs{dir}` syntax?

farnabaz:
> I would love if there wouldn't have to be a special case checking for the presence of nuxt-vite 

I asked pooya to check it, I wanted to make sure about checking `vite` environment here (He may have better idea): 
```
/* <% const _srcDir = (isDev && (nuxtOptions.buildModules || []).includes('nuxt-vite')) ? '/@fs' + srcDir : '~' %> */
```

> Do you have some links that explain that `@fs{dir}` syntax?

It is the second reason why I keep this drafted and asked Pooya. `@fs` is not documented and I found it in Vite internals.

rchl:
> > I would love if there wouldn't have to be a special case checking for the presence of nuxt-vite
> > I wanted to make sure about checking `vite` environment here:
> 
> ```
> /* <% const _srcDir = (isDev && (nuxtOptions.buildModules || []).includes('nuxt-vite')) ? '/@fs' + srcDir : '~' %> */
> ``
> ```

What do you mean exactly? This code is still there and working. Putting it in block comment doesn't affect the behavior since it doesn't print anything.

farnabaz:
My bad, your refactor is great :) I was addressing `isDev && (nuxtOptions.buildModules || []).includes('nuxt-vite'))`

rchl:
Ideally, Vite would also support the `~` alias since it's used in many places. I'm surprised if it doesn't.

(sorry, I've removed my earlier comment as I've realized that you meant something else)

pi0:
@farnabaz @rchl At least at this stage, we should not depend on any vite-specific feature like checking it's existence from buildModules or `@fs/` requests.

> Ideally, Vite would also support the ~ alias since it's used in many places

Actually it should do since we are providing resolve.alias option for `~` (otherwise it is a bug in either vite or nuxt-vite)

Let me check PR locally :)

src/templates/utils.js
```diff
@@ -1,5 +1,25 @@
...
+/* <% if (options.langDir) { %> */
+/* <% const _srcDir = (isDev && (nuxtOptions.buildModules || []).includes('nuxt-vite')) ? '/@fs' + srcDir : '~' %> */
+async function loadFileModule (lang) {
+  let langFileModule
+  switch (lang) {
+    /* <% for (let locale of options.locales) { %> */
+    case '<%= locale.code %>':
+      langFileModule = await import(
+        /* webpackChunkName: "lang-[request]" */
+        /* webpackInclude: /\.(js|ts|json|ya?ml)$/ */
```
rchl: If the file is explicitly specified now then I guess we don't need this directive anymore.

src/templates/options.js
```diff
@@ -19,3 +19,9 @@ for (const [key, value] of Object.entries(options)) {
     }
 }
 %>
+
+/* <% if (options.langDir) { %> */
+export const ASYNC_LOCALES = {
+  <%= options.locales.map(l => `'${l.code}': () => import('${options.langDir}/${l.file || l.iso}' /* webpackChunkName: "lang-${l.code}" */)`).join(',\n  ') %>
```
pi0: Notes:

Generates this:
```js
export const ASYNC_LOCALES = {
  'en': () => import('../../lang/en' /* webpackChunkName: "lang-en" */),
  'fr': () => import('../../lang/fr-FR' /* webpackChunkName: "lang-fr" */)
}
```

Adds a new feature to fallback to iso for default filename. @rchl please let me know if you prefer to move it to a feat PR.

rchl: The `webpackChunkName` could be `lang-${l.file}`. That way if multiple locales reuse the same language file, those wouldn't create separate chunks.

rchl: The `|| l.iso` part seems new and I'm not sure it makes sense. `iso` value doesn't have a file extension so importing it would likely not work.

rchl (changes requested):

src/templates/options.js
rchl: Made a separate comment about the `iso` thing.

pi0: Resovler (cjs, webpack, vite) does rersolve extension for us. It is convention over configuration. But agree have to move it to new PR later.

rchl: Also agree. And if we want that, we should also document it.

How is it resolved automatically though? Based on content? How would it know if it's a json or yaml or js?

src/templates/options.js
```diff
@@ -22,6 +22,6 @@ for (const [key, value] of Object.entries(options)) {
 
 /* <% if (options.langDir) { %> */
 export const ASYNC_LOCALES = {
-  <%= options.locales.map(l => `'${l.code}': () => import('../${relativeToBuild(options.langDir, l.file || l.iso)}' /* webpackChunkName: "lang-${l.code}" */)`).join(',\n  ') %>
+  <%= options.locales.filter(l => l.file).map(l => `'${l.code}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`).join(',\n  ') %>
```
rchl: It's better to fail at build-time than at runtime when `l.file` is missing. I'm not sure if it would fail without the `filter` though...

The documentation states that [`file` is required](https://i18n.nuxtjs.org/options-reference#locales) when using lazy loading.

pi0: Indeed we should fail earlier in module when lazy option is on (which i think is not current behavior of module)

pi0: I've reverted for this PR

src/templates/options.js
```diff
@@ -19,3 +19,9 @@ for (const [key, value] of Object.entries(options)) {
     }
 }
 %>
+
+/* <% if (options.langDir) { %> */
+export const ASYNC_LOCALES = {
+  <%= options.locales.filter(l => l.file).map(l => `'${l.code}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`).join(',\n  ') %>
```
rchl: BTW. `relativeToBuild` is a nice one. Didn't know about it. :)

rchl:
Thanks to both of you for your work on that.

src/templates/options.js
```diff
@@ -19,3 +19,9 @@ for (const [key, value] of Object.entries(options)) {
     }
 }
 %>
+
+<% if (options.langDir) { %>
+export const ASYNC_LOCALES = {
+  <%= options.locales.map(l => `'${l.file}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`).join(',\n  ') %>
```
rchl: If multiple locales reuse the same `file` then now the object will contain repeated and redundant properties.

src/templates/options.js
```diff
@@ -22,6 +22,9 @@ for (const [key, value] of Object.entries(options)) {
 
 <% if (options.langDir) { %>
 export const ASYNC_LOCALES = {
-  <%= options.locales.map(l => `'${l.file}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`).join(',\n  ') %>
+  <%= options.locales
+    .map(l => `'${l.file}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`)
+    .filter((l, index, self) => self.indexOf(l) === index)
```
pi0: This could be done in a simpler approach by mapping to file and use a Set 😃

src/templates/options.js
```diff
@@ -22,9 +22,8 @@ for (const [key, value] of Object.entries(options)) {
 
 <% if (options.langDir) { %>
 export const ASYNC_LOCALES = {
-  <%= options.locales
-    .map(l => `'${l.file}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`)
-    .filter((l, index, self) => self.indexOf(l) === index)
-    .join(',\n  ') %>
+  <%= Array.from(
+        new Set(options.locales.map(l => `'${l.file}': () => import('../${relativeToBuild(options.langDir, l.file)}' /* webpackChunkName: "lang-${l.file}" */)`))
```
pi0, line 26: Cool! I was thinking to unique on and then generate imports but this is also nice pattern might :D

pi0 (approved):
Needs to tested but looks good to me now

rchl (approved):
LGTM too. I've tested it in one of my projects that uses lazy loading and of course tests pass also.

rchl:
Haven't tested with Vite though so I can wait until someone does that.

pi0:
> Haven't tested with Vite though so I can wait until someone does that.

There is no pressure on vite. Since everything is super experimental know so i don't want to make chaos in community for vite support. These PRs generally improved runtime quality with esm support.

PR #1087, opened 2021-03-04
fix(docs): adjusted example code in documentation to call the right context

gekkedev:
https://i18n.nuxtjs.org/lang-switcher - I went to this page and tried out the given snippet; got an error; tried the obvious change inspired by the previous examples; it worked. Here's the necessary change to the documentation.


rchl:
That makes sense. Thanks.

PR #1088, opened 2021-03-04
fix: support localePath with path input and customized routes

Energon7:
```js
 nuxtI18n: {
        paths: {
          en: '/about-us', 
          fr: '/a-propos',
          es: '/sobre'
      }
  }
```

Example:
Current locale is: fr.
Calling `localePath('/about-us')` --> returns /fr/a-propos;
or calling `localePath('/en/about-us')` --> returns /fr/a-propos;

tested when strategy is  'prefix_except_default'

### DIFF
--- a/src/templates/plugin.routing.js
+++ b/src/templates/plugin.routing.js
@@ -63,19 +63,20 @@ function localeRoute (route, locale) {
       !i18n.differentDomains
     const thisRoute = this.router.resolve(route.path).route
     const routeName = this.getRouteBaseName(thisRoute)
-
     if (isPrefixed && !routeName) {
       let path = `/${locale}${route.path}`
-      path = path.replace(/\/+$/, '') + (trailingSlash ? '/' : '') || '/'
+      path = getReplacedPath(path)
       localizedRoute.path = path
     } else {
-      localizedRoute.path = route.path.replace(/\/+$/, '') + (trailingSlash ? '/' : '') || '/'
+      localizedRoute.path = getReplacedPath(route.path)
       if (routeName) {
-        localizedRoute.name = getLocaleRouteName(routeName, locale)
-        localizedRoute.params = thisRoute.params
-        localizedRoute.query = thisRoute.query
-        localizedRoute.hash = thisRoute.hash
-        return this.router.resolve(localizedRoute).route
+        return this.router.resolve({
+          name: getLocaleRouteName(routeName, locale),
+          params: thisRoute.params,
+          query: thisRoute.query,
+          hash: thisRoute.hash,
+          path: localizedRoute.path
+        }).route
       }
     }
   } else {
@@ -92,6 +93,10 @@ function localeRoute (route, locale) {
   }
 
   return this.router.resolve(localizedRoute).route
+}
+
+function getReplacedPath (path) {
+  return path.replace(/\/+$/, '') + (trailingSlash ? '/' : '') || '/'
 }
 
 function switchLocalePath (locale) {
--- a/src/templates/plugin.routing.js
+++ b/src/templates/plugin.routing.js
@@ -55,12 +55,12 @@ function localeRoute (route, locale) {
     const isDefaultLocale = locale === defaultLocale
     // if route has a path defined but no name, resolve full route using the path
     const isPrefixed =
-      // don't prefix default locale
-      !(isDefaultLocale && [STRATEGIES.PREFIX_EXCEPT_DEFAULT, STRATEGIES.PREFIX_AND_DEFAULT].includes(strategy)) &&
-      // no prefix for any language
-      !(strategy === STRATEGIES.NO_PREFIX) &&
-      // no prefix for different domains
-      !i18n.differentDomains
+        // don't prefix default locale
+        !(isDefaultLocale && [STRATEGIES.PREFIX_EXCEPT_DEFAULT, STRATEGIES.PREFIX_AND_DEFAULT].includes(strategy)) &&
+        // no prefix for any language
+        !(strategy === STRATEGIES.NO_PREFIX) &&
+        // no prefix for different domains
+        !i18n.differentDomains
     const thisRoute = this.router.resolve(route.path).route
     const routeName = this.getRouteBaseName(thisRoute)
     if (isPrefixed && !routeName) {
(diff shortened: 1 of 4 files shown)


rchl:
It doesn't look too good at the first sight to be honest.

At minimum you need to run the tests when developing.

Energon7:
You will need to update tests. since, now translated links are returned

rchl:
Sorry but you need to make sure that your change and the tests are aligned. I'm sure you will find cases where your change is incorrect if you analyze the failing tests.

Also, `localeRoute` is supposed to return a route (or nothing) while with your change it sometimes returns a plain path. This is not in line with the API and is the first issue I see from a far perspective.

Energon7:
Now check please, tests passed successfully.

rchl:
They did not pass yet.

I'll check more thoroughly when I have more time but please continue with it.

(Also you have a lot of style issues that will be reported once the tests pass. You can see those locally by running `yarn lint`)

rchl:
Also, I'd like a new test added that showcases a case that was previously failing but doesn't anymore with the fix.

src/templates/plugin.routing.js
```diff
@@ -55,16 +55,26 @@ function localeRoute (route, locale) {
...
-
-    let path = (isPrefixed ? `/${locale}${route.path}` : route.path)
-    path = path.replace(/\/+$/, '') + (trailingSlash ? '/' : '') || '/'
-    localizedRoute.path = path
+      // don't prefix default locale
+      !(isDefaultLocale && [STRATEGIES.PREFIX_EXCEPT_DEFAULT, STRATEGIES.PREFIX_AND_DEFAULT].includes(strategy)) &&
+      // no prefix for any language
+      !(strategy === STRATEGIES.NO_PREFIX) &&
+      // no prefix for different domains
+      !i18n.differentDomains
```
rchl: No need to change indentation here and make the resulting diff bigger.

Energon7:
@rchl Your browser test is wrong. The error is related to this line:  `75. localizedRoute.query = thisRoute.query`. 
If remove this line, all tests will pass successfully, but in this case, if you set the any query, it will be removed in translation.
For example: 
setLocale('/about?page=1') --> /fr/a-propos

rchl:
The test asserts that loading a `/nopage?a#h` path doesn't change it to something else. And with your fix the URL is changed to `/nopage?a#h?a` which is obviously wrong.

The test is not wrong, there is a bug in your changes.

src/templates/plugin.routing.js
rchl: Please revert this indentation change.

src/templates/plugin.routing.js
```diff
@@ -55,16 +55,29 @@ function localeRoute (route, locale) {
...
+    const thisRoute = this.router.resolve(route.path).route
+    const routeName = this.getRouteBaseName(thisRoute)
+
+    if (isPrefixed && !routeName) {
+      let path = `/${locale}${route.path}`
+      path = path.replace(/\/+$/, '') + (trailingSlash ? '/' : '') || '/'
+      localizedRoute.path = path
+    } else {
+      localizedRoute.path = route.path.replace(/\/+$/, '') + (trailingSlash ? '/' : '') || '/'
```
rchl: Those two `replace` calls are basically the same so you can re-arrange the code a bit to not repeat yourself.

Energon7:
Now tests passed successfully.

src/templates/plugin.routing.js
Energon7: no, the first path includes locale, other does not

src/templates/plugin.routing.js
```diff
@@ -55,16 +55,29 @@ function localeRoute (route, locale) {
...
+      if (routeName) {
+        localizedRoute.name = getLocaleRouteName(routeName, locale)
+        localizedRoute.params = thisRoute.params
+        localizedRoute.query = thisRoute.query
+        localizedRoute.hash = thisRoute.hash
+        return this.router.resolve(localizedRoute).route
```
rchl: You can remove this line and let the line at the end of the function handle that.

src/templates/plugin.routing.js
rchl: I'm talking about the `.replace(/\/+$/, '') + (trailingSlash ? '/' : '') || '/'` part.

You can refactor the code so that you don't repeat that part.

src/templates/plugin.routing.js
Energon7: There are no ideas what is the matter, the indents are the same.

src/templates/plugin.routing.js
Energon7: it's done.

src/templates/plugin.routing.js
Energon7: it's also done

src/templates/plugin.routing.js
Energon7: Done.

rchl:
I've also asked for a new test that tests the behavior that was fixed.

If you don't have time or don't know how to do it then this will have to wait until I have time.

src/templates/plugin.routing.js
```diff
@@ -61,10 +61,24 @@ function localeRoute (route, locale) {
...
+      localizedRoute.path = getReplacedPath(route.path)
+      if (routeName) {
+        return this.router.resolve({
+          name: getLocaleRouteName(routeName, locale),
+          params: thisRoute.params,
+          query: thisRoute.query,
+          hash: thisRoute.hash,
+          path: localizedRoute.path
+        }).route
+      }
```
rchl: I'd have to verify that with tests but this logic should be equivalent with no extra functions and less code duplication:

```suggestion
    let path = route.path
    if (isPrefixed && !routeName) {
      path = `/${locale}${path}`
    }
    localizedRoute.path = path.replace(/\/+$/, '') + (trailingSlash ? '/' : '') || '/'
    if (routeName) {
      localizedRoute = {
        name: getLocaleRouteName(routeName, locale),
        params: thisRoute.params,
        query: thisRoute.query,
        hash: thisRoute.hash,
        path: localizedRoute.path
      }
    }
```

Energon7: Tests passed successfully. Thanks.

Energon7: I have committed your code

rchl:
Refactored a bit to make the logic clearer and added some tests.

Now doing the trailing slash normalization only for the case when there is no resolved route since router should be doing that itself for resolved route.

rchl (approved):

rchl:
Thanks!

Issue #873, opened 2020-09-01
Access `$nuxtI18nSeo` through @nuxtjs/composition-api


bissolli:
Hey guys,

I am trying to access `$nuxtI18nSeo` from the setup method using `useContext` from nuxt composition API but I can't find a way to access it.

Any guesses?

**UPDATE:**
I managed to access it with the following code... will that to what it's supposed to do? Also is that the "correct" way to do that?

```javascript
const { app: { i18n } } = useContext() 
const seo = i18n.vm.$nuxtI18nSeo()
```

rchl:
If it works and returns correct information then I guess it's just fine.
What are you doing with that `seo` object though? How are you passing it to VueMeta?
I don't have much experience with composition API yet so not sure what would be the ideal way. Maybe that SEO object should be exposed as new API?

bissolli:
@rchl it's actually not returning the correct information but just an empty object.

I found another way to do that which did the job! See below:

```javascript
...
setup(_, { root }: SetupContext) {

  const i18nSeo = root.$nuxtI18nSeo()
  useMeta({
      title,
      htmlAttrs: {
        ...i18nSeo.htmlAttrs,
      }
     ...
  })

}
...
```

I am leaving the issue open just in case someone has a better solution - it doesn't seems nice to accessing `root` for that 😅

rchl:
I would need to provide Vue3-compatible API as described at https://dev.to/alvarosaburido/how-to-migrate-your-library-from-vue2-to-vue3-1h81

Similarly to how VueMeta does it already.
BTW. How did you configure VueMeta? Where did you place the VueMeta initialization code?

aaronhuisinga:
@bissolli I'm stuck with this same issue. For some reason, however, the solution you posted using `root.$nuxtI18nSeo()` doesn't work for us. We get a "Cannot read property 'req' of undefined" error from .nuxt/nuxt-i18n/plugin.routing.js on line 167.

Did you run into this same issue, or did it just work for you right away? We might just temporarily ditch the SEO tags for now as we move to composition API until we can get it all working.

TheAlexLichter:
I've tried monkeypatching the `req` assignment to access it via `context.ssrContext.req`. That worked fine in my tests when using the composition API and `target: 'static'`:

I changed https://github.com/nuxt-community/i18n-module/blob/master/src/templates/plugin.routing.js#L166 to 
[...]

PR #1089, opened 2021-03-04
test: add fixture that uses @nuxtjs/composition-api

dschreij:
If one uses the @nuxtjs/composition-api module, nuxt-i18n causes a blocking error because it cannot find req on `this.$ssrContext`. This is only the case if nuxt.config contains

```js
{
  ssr: true,
  target: 'server'
}
```
(which are the default settings)

Apparently, when using the composition api module, ssrContext is moved from the root to the context object (https://github.com/nuxt-community/i18n-module/issues/711#issuecomment-733650829)

This is a fix as suggested by @manniL in https://github.com/nuxt-community/i18n-module/issues/873#issuecomment-707584305.
With this fix the default this.$ssrContext is just used if it is defined, which should be the case if composition-api is not used. If it is not found there, an attempt is done to find ssrContext on the context object, and otherwise the value simply becomes `null` (just as in the original scenario).

I know a similar PR has been submitted once before (https://github.com/nuxt-community/i18n-module/pull/596) but there was no mention of the composition API being a cause there. Without this change nuxt-i18n does not work with the (official) @nuxt/composition-api module, so I thought that would warrant to open up a new PR. The error should also be reproducible in this case by adding both nuxt-i18n and @nuxtjs/composition-api modules.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "@babel/preset-env": "7.13.8",
     "@babel/runtime": "7.13.8",
     "@nuxt/types": "2.15.2",
-    "@nuxtjs/composition-api": "^0.21.0",
+    "@nuxtjs/composition-api": "0.21.0",
     "@nuxtjs/eslint-config-typescript": "5.0.0",
     "@nuxtjs/module-test-utils": "1.6.3",
     "@release-it/conventional-changelog": "2.0.1",
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
     "@babel/preset-env": "7.13.8",
     "@babel/runtime": "7.13.8",
     "@nuxt/types": "2.15.2",
+    "@nuxtjs/composition-api": "^0.21.0",
     "@nuxtjs/eslint-config-typescript": "5.0.0",
     "@nuxtjs/module-test-utils": "1.6.3",
     "@release-it/conventional-changelog": "2.0.1",
--- a/src/templates/plugin.routing.js
+++ b/src/templates/plugin.routing.js
@@ -150,7 +150,7 @@ const VueInstanceProxy = function (targetFunction) {
       getRouteBaseName: this.getRouteBaseName,
       i18n: this.$i18n,
       localePath: this.localePath,
-      req: process.server ? this.$ssrContext?.req ?? this.context?.ssrContext?.req : null,
+      req: process.server ? this.$ssrContext.req : null,
       route: this.$route,
       router: this.$router,
       store: this.$store
--- a/src/templates/plugin.routing.js
+++ b/src/templates/plugin.routing.js
@@ -150,7 +150,7 @@ const VueInstanceProxy = function (targetFunction) {
       getRouteBaseName: this.getRouteBaseName,
       i18n: this.$i18n,
       localePath: this.localePath,
-      req: process.server ? this.$ssrContext.req : null,
+      req: process.server ? this.$ssrContext?.req ?? this.context?.ssrContext?.req : null,
       route: this.$route,
       router: this.$router,
       store: this.$store
--- a/test/fixture/composition-api/nuxt.config.js
+++ b/test/fixture/composition-api/nuxt.config.js
@@ -5,7 +5,9 @@ import BaseConfig from '../base.config'
 const config = {
   ...BaseConfig,
   buildDir: resolve(__dirname, '.nuxt'),
-  srcDir: __dirname
+  srcDir: __dirname,
+  buildModules: [
+    '@nuxtjs/composition-api'
+  ]
 }
-config.modules?.push('@nuxtjs/composition-api')
-module.exports = config
+export default config
--- a/test/fixture/composition-api/nuxt.config.js
+++ b/test/fixture/composition-api/nuxt.config.js
@@ -10,4 +10,5 @@ const config = {
     '@nuxtjs/composition-api'
   ]
 }
-export default config
+
+module.exports = config
--- a/test/fixture/composition-api/nuxt.config.js
+++ b/test/fixture/composition-api/nuxt.config.js
@@ -1,0 +1,11 @@
+import { resolve } from 'path'
+import BaseConfig from '../base.config'
+
+/** @type {import('@nuxt/types').NuxtConfig} */
+const config = {
+  ...BaseConfig,
+  buildDir: resolve(__dirname, '.nuxt'),
+  srcDir: __dirname
+}
+config.modules?.push('@nuxtjs/composition-api')
+module.exports = config
--- a/test/fixture/composition-api/pages/index.vue
+++ b/test/fixture/composition-api/pages/index.vue
@@ -1,7 +1,19 @@
 <template>
   <div>
     <div id="current-locale">locale: {{ $i18n.locale }}</div>
-    <div id="content">{{ someText }}</div>
+    <nuxt-link
+      id="unprocessed-url"
+      :to="localePath(unprocessedUrl.to)"
+      v-text="$t(unprocessedUrl.text)"
+    />
+    <nuxt-link
+      id="processed-url"
+      :to="processedUrl.to"
+      v-text="processedUrl.text"
+    />
+    <div id="route-base-name">
+      {{ getRouteBaseName() }}
+    </div>
   </div>
 </template>
 
@@ -9,9 +21,19 @@
 import { defineComponent, useContext } from '@nuxtjs/composition-api'
 export default defineComponent({
   setup () {
-    const { app: { i18n } } = useContext()
-    const someText = i18n.t('untranslated')
-    return { someText }
+    const { app: { i18n, localePath } } = useContext()
+
+    const unprocessedUrl = {
+      text: 'home',
+      to: 'index'
+    }
+
+    const processedUrl = {
+      text: i18n.t('home'),
+      to: localePath('index')
+    }
+
+    return { unprocessedUrl, processedUrl }
   }
 })
 </script>
--- a/test/fixture/composition-api/pages/index.vue
+++ b/test/fixture/composition-api/pages/index.vue
@@ -1,0 +1,17 @@
+<template>
+  <div>
+    <div id="current-locale">locale: {{ $i18n.locale }}</div>
+    <div id="content">{{ someText }}</div>
+  </div>
+</template>
+
+<script>
+import { defineComponent, useContext } from '@nuxtjs/composition-api'
+export default defineComponent({
+  setup () {
+    const { app: { i18n } } = useContext()
+    const someText = i18n.t('untranslated')
+    return { someText }
+  }
+})
+</script>
--- a/test/module.test.js
+++ b/test/module.test.js
@@ -2077,19 +2077,28 @@ describe('Composition API', () => {
   /** @type {Nuxt} */
   let nuxt
 
+  const override = {
+    ssr: true,
+    target: 'server'
+  }
+
   beforeAll(async () => {
-    nuxt = (await setup(loadConfig(__dirname, 'composition-api'))).nuxt
+    nuxt = (await setup(loadConfig(__dirname, 'composition-api', override, { merge: true }))).nuxt
   })
 
   afterAll(async () => {
     await nuxt.close()
   })
 
-  test('should not crash if composition API module is active', async () => {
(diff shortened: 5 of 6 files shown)


TheAlexLichter:
Adding a small fixture (https://github.com/nuxt-community/i18n-module/tree/master/test/fixture) to show that it works with CAPI now would be ideal (to avoid regressions later)

dschreij:
@manniL On it! :)
Might not have time to finish it until tonight, but will get back about it soon.

src/templates/plugin.routing.js
```diff
@@ -150,7 +150,7 @@ const VueInstanceProxy = function (targetFunction) {
       getRouteBaseName: this.getRouteBaseName,
       i18n: this.$i18n,
       localePath: this.localePath,
-      req: process.server ? this.$ssrContext.req : null,
+      req: process.server ? this.$ssrContext?.req ?? this.context?.ssrContext?.req : null,
```
rchl: I wonder if you have found any reference/documentation for that `context` property somewhere. It seems public but not sure if it's official.

dschreij: No I have been looking for this as well in the official nuxt document, but have yet to find it. The closest I could find was https://nuxtjs.org/docs/2.x/internals-glossary/context but there is no mention of `ssrContext` in there.
I don't know if it is composition api specific either. Maybe @danielroe can shed some light on this?

dschreij:
I've added a small test for this change, but some notes regarding this:
- Even though the composition api docs tell to place the module in the `buildModules` I could only get this to work if I placed `@nuxtjs/composition-api` in `modules`. I didn't see buildModules being used in any other tests, so I assume this is fine, but please tell me if you think otherwise. When I did place the CAPI module in the buildModules array, the test failed with the message the the i18n module could not load `fs`, which seems totally unrelated to my additions.
- The test passes if the composition api is loaded, but I don't explicitly assess if everything still works if the module isn't loaded. Since this is the case in all the other tests, I think that is acceptable.
- Should the test explicitly check for the presence of `req`? Or do we just assume that since the test passes and the app seemingly doesn't crash that this is the case. (I know, assumption is the mother of al f..... )

package.json
```diff
@@ -96,6 +96,7 @@
     "@babel/preset-env": "7.13.8",
     "@babel/runtime": "7.13.8",
     "@nuxt/types": "2.15.2",
+    "@nuxtjs/composition-api": "^0.21.0",
```
rchl: Let's remove the `^`. Renovate would otherwise pin it later.

test/fixture/composition-api/pages/index.vue
```diff
@@ -0,0 +1,17 @@
...
+    <div id="content">{{ someText }}</div>
+  </div>
+</template>
+
+<script>
+import { defineComponent, useContext } from '@nuxtjs/composition-api'
+export default defineComponent({
+  setup () {
+    const { app: { i18n } } = useContext()
+    const someText = i18n.t('untranslated')
```
rchl: Let's use an existing string so that we can check that it translates also:
```suggestion
    const someText = i18n.t('home')
```
That should translate to `Homepage`.

rchl (changes requested):
Everything appears to pass even without the fix in `plugin.routing.js`. Do you know why is that? Are we not triggering the right paths with the test?

rchl:
> Everything appears to pass even without the fix in `plugin.routing.js`. Do you know why is that? Are we not triggering the right paths with the test?

It appears that `req` is only used with the `differentDomains` option enabled?

test/fixture/composition-api/pages/index.vue
dschreij: Good point. Changed this

dschreij:
> > Everything appears to pass even without the fix in `plugin.routing.js`. Do you know why is that? Are we not triggering the right paths with the test?
> 
> It appears that `req` is only used with the `differentDomains` option enabled?

I reverted my change and ran the test, and it indeed passes. I have no clue why the problem doesn't occur in this case and will look into this. In general, `req` doesn't have to be used for this error to occur. The problem is that $ssrContext is undefined in

```js
req: process.server ? this.$ssrContext.req : null,
```

when the composition API is used, so every call to `VueInstanceProxy` should cause this error when the CAPI is active.

In any case this is the stack trace that I receive for this error:
image
Maybe that provides some extra information. The composition API is prominently present in this trace at least. I'll see if I can alter the test in such a way that it fails in the situation that this PR attempts to fix.

rchl:
To get similar stack trace in the test you can add this to the template:
```
<div id="route-base-name">{{ getRouteBaseName() }}</div>
```

But it seems like `this.$ssrContext` **is** defined in the test at least so there must be something more to it.

danielroe:
I've not looked into how this module is registering its helpers but if `@nuxtjs/composition-api` is *changing* the object for ssrContext, that is a bug and should be addressed there.

(I think it's more likely that the `this` in the code above is pointing to a different place when invoked with CAPI and so possibly could be addressed by getting into a global context by doing something like (from memory) `this.$root.context`, `this.$nuxt.app.context` etc. Happy to look if someone shares a repro with me.)

rchl:
Once we can reproduce it will be pretty clear what's the problem. It could be that the problem is not even in composition API but in the user's project itself.

dschreij:
One key difference is that in my project, `@nuxtjs/composition-api` is registered in `buildModules` while in this test case, I am adding it to `modules`. If I try to add it to `buildModules` here it fails due to some (seemingly) unrelated error.

dschreij:
> Once we can reproduce it will be pretty clear what's the problem. It could be that the problem is not even in composition API but in the user's project itself.

that is of course always a possibility, even though I am not doing anything funky with the i18n library or the composition-api library. I'll see if I can set up a repo with a minimal setup that reproduces this issue.

rchl:
> One key difference is that in my project, `@nuxtjs/composition-api` is registered in `buildModules` while in this test case, I am adding it to `modules`. If I try to add it to `buildModules` here it fails due to some (seemingly) unrelated error.

It shouldn't really matter if it's in `modules` or `buildModules` but this works for me:

```diff
--- a/test/fixture/composition-api/nuxt.config.js
+++ b/test/fixture/composition-api/nuxt.config.js
@@ -5,7 +5,9 @@ import BaseConfig from '../base.config'
 const config = {
   ...BaseConfig,
   buildDir: resolve(__dirname, '.nuxt'),
-  srcDir: __dirname
+  srcDir: __dirname,
+  buildModules: [
+    '@nuxtjs/composition-api'
+  ]
 }
-config.modules?.push('@nuxtjs/composition-api')
 module.exports = config
```

> that is of course always a possibility, even though I am not doing anything funky with the i18n library or the composition-api library. I'll see if I can set up a repo with a minimal setup that reproduces this issue.

It would be enough that you use an arrow function somewhere which would freeze `this` and cause failure later.

dschreij:
> It would be enough that you use an arrow function somewhere which would freeze this and cause failure later.

This made me dive into my code again, and I found the culprit:

The problem occurs if I use `localePath` inside the `setup()` function, but only when I import it in a 'deprecated' way.
So:

```js
setup (_, { root: { localePath } }) {
  const url = localePath({to: 'index'})
}
```
causes the aforementioned error, but

```js
setup () {
    const { app: { localePath } } = useContext()
    const url = localePath({to: 'index'})
}
```
Works perfectly...

The question is, what does this mean for this PR and the test case. Is this test case still valid and should we guard against this situation? Or since the first way of importing localePath is deprecated anyway, just let's hope this situation goes away on its own in the (near) future?

rchl:
> The question is, what does this mean for this PR and the test case. Is this test case still valid and should we guard against this situation? Or since the first way of importing localePath is deprecated anyway, just let's hope this situation goes away on its own in the (near) future?

The newly added fixture is useful and should stay. But we should update it so that we call `localePath` both from the `setup` function and from the template so that we are checking both cases.

I would not include the change in `plugin.routing.js` though as such use is wrong and I don't want to pretend to support that use case.

dschreij:
So we should only add a use case where this is done:
```js
setup () {
    const { app: { localePath } } = useContext()
    const url = localePath({to: 'index'})
}
```
?

rchl:
> ```js
> const url = localePath({to: 'index'})
> ```

Yes.
And also this: https://github.com/nuxt-community/i18n-module/pull/1089#issuecomment-791298473

BTW. Can be just:

```js
const url = localePath('index')
```

test/module.test.js
```diff
@@ -2072,3 +2072,33 @@ describe('Locale fallback array', () => {
...
 })
+
+describe('Composition API', () => {
+  /** @type {Nuxt} */
+  let nuxt
+
+  const override = {
+    ssr: true,
+    target: 'server'
+  }
```
rchl: Aren't those the defaults? If so, I'd prefer to omit them to avoid extra noise.

test/fixture/composition-api/nuxt.config.js
```diff
@@ -0,0 +1,11 @@
+import { resolve } from 'path'
+import BaseConfig from '../base.config'
+
+/** @type {import('@nuxt/types').NuxtConfig} */
+const config = {
+  ...BaseConfig,
+  buildDir: resolve(__dirname, '.nuxt'),
+  srcDir: __dirname
+}
+config.modules?.push('@nuxtjs/composition-api')
```
rchl: can you change to buildModules?
```suggestion
  srcDir: __dirname,
  buildModules: [
    '@nuxtjs/composition-api'
  ]
}

```

test/fixture/composition-api/nuxt.config.js
```diff
@@ -0,0 +1,11 @@
...
+module.exports = config
```
rchl, line 14:
```suggestion
export default config
```

rchl (changes requested):

test/fixture/composition-api/nuxt.config.js
dschreij: Done

test/fixture/composition-api/nuxt.config.js
dschreij: Done

rchl:
Hmm, I'm now reconsidering whether we should be pragmatic and just handle the "wrong" case...

dschreij:
> Hmm, I'm now reconsidering whether we should be pragmatic and just handle the "wrong" case...

I think it is better to do not support it. If this issue pops up elsewhere, it will force / push people towards the right way to do it.
In the 1.0 version of composition API, the deprecated way (using root) will be removed altogether, so this fix will only have temporary value in any case.

rchl:
The tests are failing now. Is it due to using `buildModules`?

dschreij:
I get this error if I add @nuxtjs/composition-api to buildModules:

image

will have to investigate further later tonight.

dschreij:
> The tests are failing now. Is it due to using `buildModules`?

No it was due to replacing 
```js
module.exports = config
```
with
```js
export default config
```
I changed it back and now the test passes again. Locally at least.

rchl:
I think it was rather due to missing lock file changes. It's not a good practice to mix CJS and ES imports and I doubt that would be an issue.

dschreij:
Alright. All the other tests use CJS as well for next.config files, so that's why I did this too at the start.

rchl:
It's an old code that I didn't get to clean up. I hold the new code by higher standards. ;)

dschreij:
As you should ;) Do I need to try with EJS again then?

rchl:
I suppose you don't. I can clean that up later.

rchl (approved):

rchl:
Thanks!

aaronhuisinga:
@dschreij This doesn't seem to fix the SEO issue that #873 is about (and that you referenced in your PR). The fix that you had implemented in this PR and then removed _did_ fix that issue when I was testing manually. Is there a way that we can now properly use the SEO portion of this plugin with the Composition API after this has been merged, or is that still not functional?

dschreij:
I am looking at @rchl for this one. If this change fixes other issues as well it might be worthwile to add it after all. @aaronhuisinga I do think an extra test will need to be added for your specific scenario too in this case.

rchl:
This needs to be thought out and implemented properly. Going through the root Vue instance is likely not a correct solution in the first place so this should be exposed in a proper place.

PR #1090, opened 2021-03-06
fix: recursive clone triggered when "locales" object made reactive

### LINKED ISSUE #1075: RangeError: Maximum call stack size exceeded on any version after 6.17.0
### Version
nuxt-i18n: any version after 6.17.0
nuxt: 2.14.2 and 2.15.2

 - [x] universal
 - [] spa

### Nuxt-i18n configuration

inside a component we have:

methods: {
async switLang(lang) {
  await this.$i18n.setLocale(lang)
  }
}

I'm just using this.$i18n.setLocale(language);
and it works 100% fine only IF "nuxt-i18n": "^6.17.0",
any other version after that drops the error. Maybe not the first time but randomly after repeating some times the lang switching

image

I use no_prefix strategy
any help?
Thanks

rchl:
Ensure that we always clone the original "locales" when assigning to
"$i18n.locales" so that if the object is made reactive by the user, we
don't attempt to clone that.

Resolves #1075

### DIFF
--- a/src/templates/plugin.main.js
+++ b/src/templates/plugin.main.js
@@ -275,7 +275,7 @@ export default async (context) => {
   }
 
   const extendVueI18nInstance = i18n => {
-    i18n.locales = locales
+    i18n.locales = klona(locales)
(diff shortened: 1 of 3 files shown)


nelsonlarocca:
Thanks A LOT

PR #1097, opened 2021-03-08
fix: ensure langDir is escaped on Windows

kedrzu:
After updating `nuxt-i18n` my project stopped compiling. 
After some digging, I've found this problem:
image

I'm working on windows, and paths do use backslashes.
When creating script templates this requires for string to be properly escaped.

It's enough to pass it through `JSON.stringify`, so I think easiest way to fix that, was to remove the special case for string.
Also, checking for `boolean` and `null` should also be unnecessary - `JSON.stringify` should handle it as well.

PS: this would also break if there were `'` characters anywhere inside.

### DIFF
--- a/src/templates/options.js
+++ b/src/templates/options.js
@@ -1,6 +1,6 @@
 <%
 function stringifyValue(value) {
-  if (value === undefined || value === null || typeof value === 'boolean' || typeof value === 'function') {
+  if (value === undefined || typeof value === 'function') {
     return String(value);
   } else {
     return JSON.stringify(value)
--- a/src/templates/options.js
+++ b/src/templates/options.js
@@ -1,8 +1,6 @@
 <%
 function stringifyValue(value) {
-  if (typeof value === 'string') {
-    return `'${value}'`
-  } else if (value === undefined || value === null || typeof value === 'boolean' || typeof value === 'function') {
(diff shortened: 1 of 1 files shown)


rchl:
I think the issue is deeper than that. Do the ASYNC_LOCALES variables look OK in the `options.js`?

I think it will look something like this which doesn't seem correct:
Screenshot 2021-03-08 at 14 19 26

rchl:
Or maybe that's just on my system because I'm forcing a Windows path on MacOS?

kedrzu:
It didn't look so bad. Slashes were mixed (backslashes and standard ones), but they are at least escaped properly.

src/templates/options.js
```diff
@@ -1,8 +1,6 @@
 <%
 function stringifyValue(value) {
-  if (typeof value === 'string') {
-    return `'${value}'`
-  } else if (value === undefined || value === null || typeof value === 'boolean' || typeof value === 'function') {
+  if (value === undefined || value === null || typeof value === 'boolean' || typeof value === 'function') {
```
rchl: Feel free to also remove null and boolean

rchl (approved):

rchl:
Thanks!
Released in v6.20.6

PR #1099, opened 2021-03-08
fix: Make unique codes list in the getLocalCodes helper function

jinevo:
This fix attempts to remove the warnings that appear in the console and terminal when there are same codes in different domains options.

https://github.com/nuxt-community/i18n-module/issues/1076


rchl:
I'm not that sure if including duplicated locales is the best approach. Maybe the `domain` property should just support a list of domains instead?

BTW. In any case, I'd also like some test(s) added for the changes.

jinevo:
@rchl Yes, you're right, this is the best approach especially that it maintains the use of the local options in the same entry (file, iso...).

PR #1103, opened 2021-03-10
fix: localePath/localeRoute missing query params

kairyou:
After updating nuxt-i18n@6.20.6,`localePath/localeRoute`(with `path` input) will missing `query` parameters.
- `localePath({ path: '/', query: {test: 1} })` => `/`  (**missing query**)
- `localePath({ name: 'index', query: {test: 1} })` => `/?test=1`
- `localeRoute({ path: '/', query: {test: 1} })` => `query={}` (**missing query**)
- `localeRoute({ name: 'index', query: {test: 1} })` => `query={test: 1}`


rchl:
Thanks. Added some tests also.

PR #1113, opened 2021-03-18
Fix cookie params in DetectBrowserLanguageInterface

rrmesquita:
There is no reference to `crossOriginCookie` in the codebase. The right one is `cookieCrossOrigin`. Also, `cookieSecure` parameter was missing as well.


rchl:
I have a bigger refactoring in-progress that fixes this and a couple other issues with types: #1114
I should be finished with it in a day or two so I think I would prefer to avoid conflicts and just wait for the refactor.
Thanks though.

rrmesquita:
Thank you very much for the good work @rchl !

PR #1121, opened 2021-03-25
fix: move lodash.merge to deps

### LINKED ISSUE #1122: Cannot find module 'lodash.merge'. at v6.22.0
### Version
nuxt-i18n: 6.22.0
nuxt: 2.15.3

 - [X] universal

### Bug

`lodash.merge` is installed as `devDependencies` in [package.json](https://github.com/nuxt-community/i18n-module/blob/v6.22.0/package.json#L131), but actually imported at [https://github.com/nuxt-community/i18n-module/blob/v6.22.0/src/index.js#L3](https://github.com/nuxt-community/i18n-module/blob/v6.22.0/src/index.js#L3). It will make an error when building, like:

```
   │   ✖ Nuxt Fatal Error                                                                │
   │                                                                                     │
   │   Error: Cannot find module 'lodash.merge'                                          │
   │   Require stack:                                                                    │
   │   - /xxx/node_modules/nuxt-i18n/src/index.js                                        │

```

unless I run `$ npm i lodash.merge` manually.

It looks like `lodash.merge` should be move to `dependencies` in package.json?

mi-upto:
bugfix: Cannot find module 'lodash.merge"

スクリーンショット 2021-03-25 11 03 06

Using it in src/index.js, but it's not in dependencies.

----------------------------
resolve #1122

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,6 +95,7 @@
     "is-https": "^3.0.0",
     "js-cookie": "^2.2.1",
     "klona": "^2.0.4",
+    "lodash.merge": "4.6.2",
     "ufo": "^0.6.7",
     "vue-i18n": "^8.23.0"
   },
@@ -128,7 +129,6 @@
     "jest": "26.6.3",
     "jest-dev-server": "4.4.0",
     "jsdom": "16.4.0",
-    "lodash.merge": "4.6.2",
     "messageformat": "2.3.0",
     "nuxt": "2.15.2",
     "playwright-chromium": "1.9.1",


package.json
```diff
@@ -95,6 +95,7 @@
     "is-https": "^3.0.0",
     "js-cookie": "^2.2.1",
     "klona": "^2.0.4",
+    "lodash.merge": "4.6.2",
```
rchl, line 98: should be a semantic range since the rules are different in dependencies.

But I've done the change on master and released a fix already. Thanks for that though.

PR #1150, opened 2021-04-15
refactor: initialize some variable at runtime to allow overrides

tatarysh:
Currently I need to overwrite the settings before they can be processed by the plugin, as it dynamically adds new languages from the API. Without this change, the creation of `getLocaleFromRoute` takes place at the moment of importing the file, and for me it is necessary that it takes place only during the processing of the plugin. this is a very important change if someone wants to add languages from API or other source. I tested it locally and with this change I am able to change the default i18n settings regardless of what was set in nuxt.config.js


rchl:
This makes sense but pushed some small change to remove some duplication.

PR #1162, opened 2021-05-03
docs: fix localeLocation example

saeub:


rchl:
Can you also do the same for `es` version?

saeub:
@rchl

> Can you also do the same for `es` version?

Done!

PR #1169, opened 2021-05-11
fix: different domains seo links

### LINKED ISSUE #1167: SEO: Domains in the webpage source code starts with http but expects to start with https
### Version
nuxt-i18n: ^6.18.0
nuxt: ^2.14.12

 - [✔️] universal
 - [] spa

### Nuxt-i18n configuration

```
 {
        lazy: true,
        seo: true,
        langDir: 'lang/',
        defaultLocale: 'en',
        detectBrowserLanguage: false,
        locales: [
          {
            code: 'en',
            name: '',
            iso: 'en-US',
            file: 'en-US.ts',
            domain: 'xxx.xxx.io',
          },
          {
            code: 'cn',
            name: '中文',
            iso: 'zh-CN',
            file: 'zh-CN.ts',
            domain: 'xxx.xxx.cn',
          },
        ],
        differentDomains: true,
 },
```

### Steps to reproduce
Use `nuxt generate` to generate `dist` package...
Right-click to view the source code of the webpage

### What is Expected?
```
<link data-n-head="ssr" data-hid="i18n-alt-en" rel="alternate" href="https://xxx.xxx.io/" hreflang="en">
```

```
<link data-n-head="ssr" data-hid="i18n-alt-en" rel="alternate" href="http://xxx.xxx.io/" hreflang="en">
```

divine:
Closes: #1167

Edit: I know that I've removed the check for SSL in SSR but I think there is no reason to force that check, maybe making by default HTTPS will be better, not sure.

Another option would be adding `secure: true`, then forcing HTTPS links by default but you're against that 😄.

Thanks!

### DIFF
--- a/src/templates/utils-common.js
+++ b/src/templates/utils-common.js
@@ -1,5 +1,6 @@
 import Cookie from 'cookie'
 import JsCookie from 'js-cookie'
+import { hasProtocol } from 'ufo'
 
 /** @typedef {import('../../types/internal').ResolvedOptions} ResolvedOptions */
 
@@ -120,6 +121,9 @@ export function getDomainFromLocale (localeCode, req, { normalizedLocales }) {
 // Lookup the `differentDomain` origin associated with given locale.
   const lang = normalizedLocales.find(locale => locale.code === localeCode)
   if (lang && lang.domain) {
+    if (hasProtocol(lang.domain, true)) {
+      return lang.domain
+    }
     let protocol
     if (process.server) {
       protocol = (req && isHTTPS(req)) ? 'https' : 'http'
--- a/src/templates/utils-common.js
+++ b/src/templates/utils-common.js
@@ -121,7 +121,7 @@ export function getDomainFromLocale (localeCode, req, { normalizedLocales }) {
 // Lookup the `differentDomain` origin associated with given locale.
   const lang = normalizedLocales.find(locale => locale.code === localeCode)
   if (lang && lang.domain) {
-    if (hasProtocol(lang.domain, true)) {
+    if (hasProtocol(lang.domain)) {
       return lang.domain
     }
     let protocol
--- a/test/module.test.js
+++ b/test/module.test.js
@@ -84,6 +84,18 @@ describe('differentDomains enabled', () => {
         iso: 'fr-FR',
         name: 'Français',
         domain: 'fr.nuxt-app.localhost'
+      },
+      {
+        code: 'ru',
+        iso: 'ru-RU',
+        name: 'Русский',
+        domain: 'https://ru.nuxt-app.localhost'
+      },
+      {
+        code: 'ua',
+        iso: 'uk-UA',
+        name: 'Українська',
+        domain: 'http://ua.nuxt-app.localhost'
       }
     ]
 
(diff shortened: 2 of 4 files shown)


rchl:
> Edit: I know that I've removed the check for SSL in SSR but I think there is no reason to force that check,

We can't just make a breaking change unless we make a major version release but I don't want to do that now.

We need to preserve the existing behavior that auto-detects the protocol from headers on the server side and from the protocol on the client side.

The fix needs to extend the functionality.

We could just respect the protocol if included in the `domain` key.

Also we need to document the new functionality.

test/module.test.js
```diff
@@ -84,6 +84,18 @@ describe('differentDomains enabled', () => {
...
+        code: 'ru',
+        iso: 'ru-RU',
+        name: 'Русский',
+        domain: 'https://ru.nuxt-app.localhost'
+      },
+      {
+        code: 'ua',
+        iso: 'uk-UA',
+        name: 'Українська',
+        domain: 'http://ua.nuxt-app.localhost'
```
rchl, line 98: As far as I can see there are no new tests utilising those additions?

divine:
Got it, changed a little bit of logic, and added documentation as well.

src/templates/utils-common.js
```diff
@@ -120,6 +121,9 @@ export function getDomainFromLocale (localeCode, req, { normalizedLocales }) {
 // Lookup the `differentDomain` origin associated with given locale.
   const lang = normalizedLocales.find(locale => locale.code === localeCode)
   if (lang && lang.domain) {
+    if (hasProtocol(lang.domain, true)) {
+      return `${parseURL(lang.domain).protocol}//${parseURL(lang.domain).host}`
```
rchl: Why do we parse? Can't we just use directly what the user has specified?

And if we really want to parse then create separate variable for parsed representation instead of parsing twice.

divine: Sorry, just copy-pasted and didn't think about it.

Should be all good now.

src/templates/utils-common.js
```diff
@@ -120,6 +121,9 @@ export function getDomainFromLocale (localeCode, req, { normalizedLocales }) {
 // Lookup the `differentDomain` origin associated with given locale.
   const lang = normalizedLocales.find(locale => locale.code === localeCode)
   if (lang && lang.domain) {
+    if (hasProtocol(lang.domain, true)) {
```
rchl: The `true` argument allows for URLs starting with `//` which I'm not sure if is supported for `alternate` link URLs. And it also wouldn't work properly due to https://github.com/nuxt-community/i18n-module/blob/7d1e74fd2a8186fe00a3ff852a0a5b72a216a915/src/templates/head-meta.js#L199-L201 not using the same logic. So we either have to disallow `//`-style URLs here or use the same check in `toAbsoluteUrl`.

divine: I think that it would be better to change `toAbsoluteUrl` function to accept `//` URLs because it could be used both on development and production deployments.

However, `//` style URLs are currently marked as anti-pattern because it gives browser opportunity to select on its own instead of forcing SSL (which is better).

Basically, at the end I think that disabling  `// ` style URLs is a way to go.

Thanks!

rchl: I agree

rchl (approved):

rchl:
Thanks

PR #1172, opened 2021-05-17
feat: transpile is-https

divine:
Adding an option to transpile `is-https` instead of inlining it.

Many other nuxt libraries are using it, and even if it's small, it's better to transpile and make things simpler.

Also, it might get some updates later.

Related https://github.com/nuxt-community/i18n-module/issues/1137 , https://github.com/nuxt-community/i18n-module/pull/1138

Thanks

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -105,6 +105,7 @@ export default function (moduleOptions) {
   this.options.render.bundleRenderer.directives = this.options.render.bundleRenderer.directives || {}
   this.options.render.bundleRenderer.directives.t = i18nExtensionsDirective
 
-  // Transpile is-https for IE11 support
-  this.nuxt.options.build.transpile.push('is-https')
+  // Transpile is-https (IE11)
+  this.options.build.transpile = this.options.build.transpile || []
+  this.options.build.transpile.push('is-https')
 }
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -104,4 +104,7 @@ export default function (moduleOptions) {
   }
   this.options.render.bundleRenderer.directives = this.options.render.bundleRenderer.directives || {}
   this.options.render.bundleRenderer.directives.t = i18nExtensionsDirective
+
+  // Transpile is-https for IE11 support
+  this.nuxt.options.build.transpile.push('is-https')
```
rchl: Should use `this.options` since that has proper types associated and is shorter.
Though you will then need to also make sure that `transpile` is an array.

divine: Hello,

Which one you'd prefer?

1)
```javascript
this.options.build.transpile = this.options.build.transpile || []
```
2)
```javascript
  if (!this.options.build.transpile) {
    this.options.build.transpile = []
  }
```

Thanks!

rchl: The first one is fine. Thanks :)

rchl (approved):

PR #1183, opened 2021-06-02
fix(differentDomains): handling of runtime domains from store

enwin:
Following documentation https://i18n.nuxtjs.org/different-domains/#runtime-environment-variables. Configuring nuxt-i18n using `differentDomains: true` and defining the domains only in the store using with runtime environment variables end up with the warning `Could not find domain name for locale xx`. 

The issue is that domain matching is made using `options.normalizedLocales` which is not updated with the domains from the store: https://github.com/nuxt-community/i18n-module/blob/master/src/templates/plugin.main.js#L370-L375.

This PR aims to update `options.locales` instead of `app.i18n.locales` since it is based on it, and update `options.normalizedLocales` so `app.i18n.localeProperties` contains the `domain` property and `getLocaleDomain` method can match a provided domain.

It also update the `onNavigate` method so we don't try to get the locale from the route when using differentDomains since the route wont contain the proper locale.


rchl:
Sorry for the delay. I'll try to understand and review it soon.

For now have only one initial thought that it's likely not safe to assume that runtime configuration is supported by this module. Thus making some cases handle runtime configuration is still not likely gonna be flawless.

src/templates/plugin.main.js
```diff
@@ -351,31 +351,37 @@ export default async (context) => {
...
     // Inject in store.
     store.$i18n = app.i18n
 
     if (store.state.localeDomains) {
-      for (const locale of app.i18n.locales) {
+      let index = 0
+      for (const locale of options.normalizedLocales) {
         if (typeof (locale) === 'string') {
           continue
         }
```
rchl: This would not be needed anymore.

src/templates/plugin.main.js
```diff
@@ -351,31 +351,37 @@ export default async (context) => {
...
-        locale.domain = store.state.localeDomains[locale.code]
+        const domain = store.state.localeDomains[locale.code]
+        locale.domain = domain
```
rchl: The options defined in the options import are not really supposed to be mutated since that affects the whole app and not only the current request.

Maybe it makes sense in this case but I will have to look closer into the logic when I have more time. But possibly we should be storing the updated info somewhere on the instance or in the store instead of mutating the original options.

enwin: I agree that mutating the options is not great. But in the default flow where you only have to deal with static domains you can define them in the configuration file. You end up having the domain property available in the options directly. Since the options import is the one mainly used for all the following tasks and tests I ended up mutating it to keep the logic the same as defining the domains directly in the configuration file. Of course I'm happy to find a better way to do it 😄 This PR was mainly to raise the issue with some description of the problem.

rchl: Sorry for me forgetting about this PR.

I guess it's acceptable after all since the domains are not really meant to change per-request but rather be defined per-app so it should be fine if the options are mutated.

rchl:
Thanks. Sorry for a very long wait.

enwin:
No worries at all! Happy to take the discussion further if you want to improve the solution 👍

PR #1190, opened 2021-06-08
fix(types): fix "lazy" type in deprecated types

fabis94:
The current typing is wrong, the `lazy` option can be a bool value or an object
https://i18n.nuxtjs.org/lazy-load-translations#lazy-configuration-options

### DIFF
--- a/types/nuxt-i18n.d.ts
+++ b/types/nuxt-i18n.d.ts
@@ -68,7 +68,7 @@ declare namespace NuxtVueI18n {
       baseUrl?: string | ((context: NuxtContext) => string)
       detectBrowserLanguage?: DetectBrowserLanguageInterface | false
       langDir?: string | null
-      lazy?: boolean
+      lazy?: boolean | {skipNuxtState: boolean}
       // see https://goo.gl/NbzX3f
       pages?: {
         [key: string]: boolean | {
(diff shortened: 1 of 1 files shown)


types/nuxt-i18n.d.ts
```diff
@@ -68,7 +68,7 @@ declare namespace NuxtVueI18n {
       baseUrl?: string | ((context: NuxtContext) => string)
       detectBrowserLanguage?: DetectBrowserLanguageInterface | false
       langDir?: string | null
-      lazy?: boolean
+      lazy?: boolean | {skipNuxtState: boolean}
```
rchl:
```suggestion
      lazy?: boolean | {skipNuxtState?: boolean}
```

The `NuxtVueI18n` namespace is deprecated though and I would suggest switching to `Options` type imported from `nuxt-i18n`. The types are also correct there.

rchl (changes requested):

types/nuxt-i18n.d.ts
fabis94: @rchl `import {Options} from 'nuxt-i18n'` doesn't find anything. Besides, I'm relying on the augmented type of the Nuxt config which types the `i18n` to `AllOptionsInterface`, so if we're supposed to move to a different type, that augmented type should be updated as well.

fabis94:
Apologies, I was on an outdated version of this lib, turns out that a newer version does indeed have these types fixed!

rchl:
Well, it does have it correct for the "new" types but those deprecated ones have this issue so I wouldn't mind fixing it.

fabis94:
@rchl It's not that important to me anymore since I'm using the latest version now, but if you like you can merge this in to fix the deprecated types

rchl:
thanks

PR #1200, opened 2021-06-18
refactor!: remove deprecated beforeLanguageSwitch

divine:
Remove deprecated beforeLanguageSwitch

### DIFF
--- a/docs/content/en/api.md
+++ b/docs/content/en/api.md
@@ -197,12 +197,6 @@ Instance of [VueI18n class](http://kazupon.github.io/vue-i18n/api/#vuei18n-class
 
   Whether `differentDomains` option is enabled.
 
-#### beforeLanguageSwitch
-
-  - **Type**: `Function`
-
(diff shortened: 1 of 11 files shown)


rchl:
Note: While doing those changes, it would be nice to at the same time start adding info about breaking changes to `migrating.md`. It will be harder to miss something if we do it right away.

divine:
> Note: While doing those changes, it would be nice to at the same time start adding info about breaking changes to `migrating.md`. It will be harder to miss something if we do it right away.

I'm not the best at writing English migration guides, however, I will try my best.

Thanks!

rchl:
We can reword later. It would be enough to just mention the change and suggest alternative.

divine:
@rchl anything is left here?

Should I continue pushing PRs while this one is open?

Thanks!

rchl (approved):

rchl:
Thanks!

PR #1210, opened 2021-06-23
refactor!: replace onlyOnNoPrefix and onlyOnRoot with redirectOn

divine:
Trying to make this PR sensible because enabling `detectBrowserLanguage.onlyOnRoot` here by default creates a lot of failing tests thus making this PR huge.

### DIFF
--- a/docs/content/en/migrating.md
+++ b/docs/content/en/migrating.md
@@ -17,6 +17,8 @@ Follow this guide to upgrade from one major version to the other.
 
 ### `seo` option has been removed. Use `$nuxtI18nHead({ addDirAttribute: true, addSeoAttributes: true })` instead.
 
+### `onlyOnRoot` and `onlyOnNoPrefix` has been removed. It's features are combined into `redirectOn` instead. Use `redirectOn: 'all|root|no prefix'` instead.
+
 ## Upgrading from 5.x to 6.x
 
 ### Global SEO features are now disabled by default
--- a/docs/content/en/options-reference.md
+++ b/docs/content/en/options-reference.md
@@ -149,7 +149,7 @@ Directory that contains translation files to load. Can be used with or without l
 ## `detectBrowserLanguage`
 
 - type: `object`
-- default: `{ alwaysRedirect: false, fallbackLocale: '', onlyOnRoot: false, useCookie: true, cookieCrossOrigin: false, cookieDomain: null, cookieKey: 'i18n_redirected', cookieSecure: false }`
+- default: `{ alwaysRedirect: false, fallbackLocale: '', redirectOn: 'all', useCookie: true, cookieCrossOrigin: false, cookieDomain: null, cookieKey: 'i18n_redirected', cookieSecure: false }`
 
 Enables browser language detection to automatically redirect visitors to their preferred locale as they visit your site for the first time.
 
@@ -157,15 +157,17 @@ See also [Browser language detection](/browser-language-detection) for a guide.
 
 <alert type="info">
 
-Note that for better SEO it's recommended to set `onlyOnRoot` to true.
+Note that for better SEO it's recommended to set `redirectOn` to `root`.
 
 </alert>
 
 Supported properties:
 - `alwaysRedirect` (default: `false`) - Set to always redirect to the value stored in the cookie, not just on first visit.
 - `fallbackLocale` (default: `null`) - If none of the locales match the browser's locale, use this one as a fallback.
-- `onlyOnRoot` (default: `false`) - Set to `true` (recommended for improved SEO) to only attempt to detect the browser locale on the root path (`/`) of the site. Only effective when using strategy other than `'no_prefix'`.
-- `onlyOnNoPrefix` (default: `false`) - This is a more permissive variant of `onlyOnRoot` that will allow attempt to detect the browser locale on the root path (`/`) and also on paths that have no locale prefix (like `/foo`). Only effective when `onlyOnRoot` is not enabled and using strategy other than `'no_prefix'`.
+- `redirectOn` (default: `all`) - Supported options:
+  - `all` - detect browser locale on all paths.
+  - `root` (recommended for improved SEO) - only detect the browser locale on the root path (`/`) of the site. Only effective when using strategy other than `'no_prefix'`.
+  - `no prefix` - a more permissive variant of `root` that will detect the browser locale on the root path (`/`) and also on paths that have no locale prefix (like `/foo`). Only effective when using strategy other than `'no_prefix'`.
 - `useCookie` (default: `true`) - If enabled, a cookie is set once the user has been redirected to browser's preferred locale, to prevent subsequent redirections. Set to `false` to redirect every time.
 - `cookieKey` (default: `'i18n_redirected'`) - Cookie name.
 - `cookieDomain` (default: `null`) - Set to override the default domain of the cookie. Defaults to the **host** of the site.
--- a/src/helpers/constants.js
+++ b/src/helpers/constants.js
@@ -17,6 +17,15 @@ export const STRATEGIES = {
   NO_PREFIX: STRATEGY_NO_PREFIX
 }
 
+const REDIRECT_ON_ALL = 'all'
+const REDIRECT_ON_ROOT = 'root'
+const REDIRECT_ON_NO_PREFIX = 'no prefix'
+export const REDIRECT_ON_OPTIONS = {
+  ALL: REDIRECT_ON_ALL,
+  ROOT: REDIRECT_ON_ROOT,
+  NO_PREFIX: REDIRECT_ON_NO_PREFIX
+}
+
 export const COMPONENT_OPTIONS_KEY = 'nuxtI18n'
 
 /** @type {Options} */
@@ -39,8 +48,7 @@ export const DEFAULT_OPTIONS = {
     cookieKey: 'i18n_redirected',
     cookieSecure: false,
     fallbackLocale: '',
-    onlyOnNoPrefix: false,
-    onlyOnRoot: false,
+    redirectOn: 'all',
     useCookie: true
   },
   differentDomains: false,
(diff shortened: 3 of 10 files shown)


rchl:
Aren't we losing support for a certain behavior now?

Previously there were three options (if we assume `alwaysRedirect` is false):
 - default behavior - redirect on all URLs
 - `onlyOnRoot` - redirect only from root
 - `onlyOnNoPrefix` - redirect only when no prefix

Now the `onlyOnRoot` options combines the two last cases.

Even if I said to "combine" those two options, I didn't mean to combine the behavior but combine those two separate options into one option that the user can use to selectively enable each behavior. So something like `redirectOn: 'root' | 'no-prefix'` (not entirely convinced on the naming here but you get the idea).

divine:
> So something like `redirectOn: 'root' | 'no-prefix'`

Hello,

I still think that we should merge both options ( `'root' | 'no-prefix'`) into one because it's indeed a bug fix or add another `lax` option which will combine two of them into one as well.

@dword-design you've worked on adding that feature, what do you think?

Thanks!

rchl:
> I still think that we should merge both options ( `'root' | 'no-prefix'`) into one because it's indeed a bug fix or add another `lax` option which will combine two of them into one as well.

I was under impression that `no-prefix` (old name `onlyOnNoPrefix`) is worse for SEO since it might result in crawler being redirected from path like `/foo` to `en/foo` while with `root` (`onlyOnRoot`) the issue would only exist for the root path.

Here is current description of the options:

- `onlyOnRoot` (default: `false`) - Set to `true` (recommended for improved SEO) to only attempt to detect the browser locale on the root path (`/`) of the site. Only effective when using strategy other than `'no_prefix'`.
- `onlyOnNoPrefix` (default: `false`) - This is a more permissive variant of `onlyOnRoot` that will attempt to detect the browser locale on the root path (`/`) and also on paths that have no locale prefix (like `/foo`). Only effective when `onlyOnRoot` is not enabled and using strategy other than `'no_prefix'`.

dword-design:
@divine Hey, not sure where I am helpful here, sounds more like an architectural issue 😄. But I will help if needed.

Merging currently does not make sense to me because `root` means it should only redirect on root and nothing else, while `no-prefix` redirects on all routes without a prefix (possibly including root). Which is distinctive behavior.

Which of both you think is a bugfix?

docs/content/en/options-reference.md
```diff
@@ -186,19 +185,6 @@ Set to a path to which you want to redirect users accessing the root URL (`/`).
...
-<alert type="warning">
-
-This option is deprecated from v6.19.0. The recommended way is to set up SEO as described in [Improving performance](/seo#improving-performance).
-
-</alert>
-
-- type: `boolean`
-- default: `false`
-
 If `true`, a SEO metadata will be generated for the routes. Note that performance can suffer with this option enabled and there might be compatibility issues with some plugins.
```
rchl: This part of `seo` description wasn't removed.

But I've removed all that on `next` branch now since it shouldn't be mixed into this PR.

divine: Sorry, my mistake, I'll remove that now.

Thanks!

divine:
Sorry guys, I misunderstood that issue!

@rchl let me know if anything else is needed.

Thanks!

docs/content/en/migrating.md
```diff
@@ -17,6 +17,8 @@ Follow this guide to upgrade from one major version to the other.
 
 ### `seo` option has been removed. Use `$nuxtI18nHead({ addDirAttribute: true, addSeoAttributes: true })` instead.
 
+### `onlyOnRoot` and `onlyOnNoPrefix` has been removed. It's features are combined into `redirectOn` instead. Use `redirectOn: 'root|no_prefix'` instead.
```
rchl: I think I would prefer the default value to be `'all'` rather than `null`.

Also would use `no prefix` instead of `no_prefix`. I guess you wanted to match the value for the `no_prefix` strategy but I'm not sure that's a good idea. Can be potentially more confusing than anything.

docs/content/en/options-reference.md
```diff
@@ -147,23 +147,22 @@ Directory that contains translation files to load. Can be used with or without l
...
 - `fallbackLocale` (default: `null`) - If none of the locales match the browser's locale, use this one as a fallback.
-- `onlyOnRoot` (default: `false`) - Set to `true` (recommended for improved SEO) to only attempt to detect the browser locale on the root path (`/`) of the site. Only effective when using strategy other than `'no_prefix'`.
-- `onlyOnNoPrefix` (default: `false`) - This is a more permissive variant of `onlyOnRoot` that will allow attempt to detect the browser locale on the root path (`/`) and also on paths that have no locale prefix (like `/foo`). Only effective when `onlyOnRoot` is not enabled and using strategy other than `'no_prefix'`.
+- `redirectOn` (default: `null`) - Set to `root` (recommended for improved SEO) to only attempt to detect the browser locale on the root path (`/`) of the site. Only effective when using strategy other than `'no_prefix'`. Set to `no_prefix` as this is a more permissive variant of `root` that will allow attempt to detect the browser locale on the root path (`/`) and also on paths that have no locale prefix (like `/foo`). Only effective when `onlyOnRoot` is not enabled and using strategy other than `'no_prefix'`.
```
rchl: The text flow is not quite right here. It says to set to `root` and then it says to set to `no_prefix`.

And still mentions `onlyOnRoot`.

```suggestion
- `redirectOn` (default: `all`) - Supported options:
  - `all` - detect browser locale on all paths.
  - `root` (recommended for improved SEO) - only detect the browser locale on the root path (`/`) of the site. Only effective when using strategy other than `'no_prefix'`.
  - `no prefix` - a more permissive variant of `root` that will detect the browser locale on the root path (`/`) and also on paths that have no locale prefix (like `/foo`). Only effective when using strategy other than `'no_prefix'`.
```

src/helpers/constants.js
```diff
@@ -17,6 +17,13 @@ export const STRATEGIES = {
   NO_PREFIX: STRATEGY_NO_PREFIX
 }
 
+const REDIRECT_ROOT = 'root'
+const REDIRECT_NO_PREFIX = 'no_prefix'
+export const REDIRECTIONS = {
+  ROOT: REDIRECT_ROOT,
+  NO_PREFIX: REDIRECT_NO_PREFIX
+}
```
rchl: Would like the `redirectOn` to be reflected in the constant name also. So something like:

```suggestion
const REDIRECT_ON_ROOT = 'root'
const REDIRECT_ON_NO_PREFIX = 'no prefix'
export const REDIRECT_ON_OPTIONS = {
  ROOT: REDIRECT_ON_ROOT,
  NO_PREFIX: REDIRECT_ON_NO_PREFIX
}
```

types/index.d.ts
```diff
@@ -5,6 +5,7 @@ import './vue'
 export { Locale }
 export type Strategies = 'no_prefix' | 'prefix_except_default' | 'prefix' | 'prefix_and_default'
 export type Directions = 'ltr' | 'rtl' | 'auto'
+export type RedirectOptions = 'root' | 'no_prefix'
```
rchl:
```suggestion
export type RedirectOnOptions = 'all' | 'root' | 'no prefix'
```

rchl (changes requested):

types/index.d.ts
```diff
@@ -21,8 +22,7 @@ export interface DetectBrowserLanguageOptions {
   cookieKey?: string
   cookieSecure?: boolean
   fallbackLocale?: Locale | null
-  onlyOnNoPrefix?: boolean
-  onlyOnRoot?: boolean
+  redirectOn?: RedirectOnOptions | null
```
rchl:
```suggestion
  redirectOn?: RedirectOnOptions
```

test/module.test.js
```diff
@@ -1797,7 +1797,7 @@ describe('prefix + detectBrowserLanguage + onlyOnNoPrefix', () => {
         strategy: 'prefix',
         detectBrowserLanguage: {
           useCookie: true,
-          onlyOnNoPrefix: true
+          redirectOn: 'no_prefix'
```
rchl:
```suggestion
          redirectOn: 'no prefix'
```

src/templates/plugin.main.js
```diff
@@ -181,9 +180,9 @@ export default async (context) => {
     }
 
     if (getLocaleFromRoute(route) === locale) {
-      // If "onlyOnRoot" or "onlyOnNoPrefix" is set and strategy is "prefix_and_default", prefer unprefixed route for
+      // If "redirectOn" is set and strategy is "prefix_and_default", prefer unprefixed route for
```
rchl:
```suggestion
      // If "redirectOn" is "all" and strategy is "prefix_and_default", prefer unprefixed route for
```

src/templates/plugin.main.js
```diff
@@ -181,9 +180,9 @@ export default async (context) => {
...
       // default locale.
-      if (!(onlyOnRoot || onlyOnNoPrefix) || locale !== options.defaultLocale || options.strategy !== Constants.STRATEGIES.PREFIX_AND_DEFAULT) {
+      if (!(redirectOn) || locale !== options.defaultLocale || options.strategy !== Constants.STRATEGIES.PREFIX_AND_DEFAULT) {
```
rchl: Would need to add constant for "all" but then should be:

```suggestion
      if (redirectOn === Constants.REDIRECT_ON_OPTIONS.ALL || locale !== options.defaultLocale || options.strategy !== Constants.STRATEGIES.PREFIX_AND_DEFAULT) {
```

rchl (changes requested):

rchl:
There are still 2 references to `onlyOnRoot` in documentation.

rchl:
> There are still 2 references to `onlyOnRoot` in documentation.

Actually that's in the SEO page so should merge `next` before updating those.

rchl:
> > There are still 2 references to `onlyOnRoot` in documentation.
> 
> Actually that's in the SEO page so should merge `next` before updating those.

No, sorry, it's actually the `browser-language-detection.md` file that still references `onlyOnRoot`.

rchl (approved):

rchl:
Thanks!

PR #1214, opened 2021-06-26
fix: redirect without appending extra chars

### LINKED ISSUE #1204: Unwanted %2F appended to url when typed in the url bar
### Version
nuxt-i18n: `^6.22.3`
nuxt: `^2.9.2`

 - [x] universal
 - [ ] spa

### Nuxt-i18n configuration
```javascript
import en from "./locales/en.json"
import fr from "./locales/fr.json"

// ...
export default {
  vueI18n: {
    locale: "en",
    fallbackLocale: "en",
    messages: { en, fr },
    locales: [
      {
        code: "en",
        name: "English",
        iso: "en-US",
      },
      {
        code: "fr",
        name: "Français",
        iso: "fr-FR",
      },
    ],
    seo: true,
    defaultLocale: "en",
    strategy: "prefix",
    lazy: false,
    langDir: "~/locales/",
    vuex: {
      moduleName: "i18n",
      syncLocale: true,
    },
    detectBrowserLanguage: {
      useCookie: true,
      alwaysRedirect: true,
    },
  },
}
```

### Steps to reproduce
We'll consider that in this example that the nuxt app supports `en` as it's language
For this given url : `/foo/bar/?q=value`
We would expect the app to resolve this into `/en/foo/bar/?q=value` when typed in the url bar
Sadly in my case the module just resolves this to `/en/foo/bar/?q=value%2F`

### What is Expected?
The url to remain without any `%2F` occurences

The url should stay as is when first typed in

divine:
Closes #1204

We might simplify it when https://github.com/unjs/ufo/pull/24 will be merged.

@rchl what do you think?


src/templates/plugin.routing.js
```diff
@@ -86,7 +87,8 @@ function resolveRoute (route, locale) {
       if (isPrefixed) {
         localizedRoute.path = `/${locale}${localizedRoute.path}`
       }
-      localizedRoute.path = localizedRoute.path.replace(/\/+$/, '') + (nuxtOptions.trailingSlash ? '/' : '') || '/'
+      const parsedPath = parsePath(localizedRoute.path)
+      localizedRoute.path = (nuxtOptions.trailingSlash ? withTrailingSlash(parsedPath.pathname) : withoutTrailingSlash(parsedPath.pathname)) + parsedPath.search + parsedPath.hash
```
rchl: Looks good but I would rather delegate joining back the path to ufo's `stringifyParsedURL`.

divine: Do you mean this?

```javascript
const parsedPath = parsePath(localizedRoute.path)
parsedPath.pathname = nuxtOptions.trailingSlash ? withTrailingSlash(parsedPath.pathname) : withoutTrailingSlash(parsedPath.pathname)
localizedRoute.path = stringifyParsedURL(parsedPath)
```

It's adding extra `?` and basically breaks, looks like a bug https://github.com/unjs/ufo/blob/2a21bc6b821eafd2284cacdf1f99a4bd133c1fd7/src/parse.ts#L68

Thanks!

rchl: I don't have much confidence in ufo at this point...

divine: I have created a new issue https://github.com/unjs/ufo/issues/28

Maybe I should take a look at another library that might do better.

Thanks!

divine: We don't have much choice as there is no alternative.

I have created a fix, let's see when this will be merged, I'll update this PR  with your suggested changes.

Thanks!

divine:
https://github.com/unjs/ufo/pull/24 has been merged.

Should be now better.

Thanks!

src/helpers/routes.js
```diff
@@ -149,7 +150,7 @@ export function makeRoutes (baseRoutes, {
       // - If "router.trailingSlash" is not specified then default to no trailing slash (like Nuxt)
       // - Children with relative paths must not start with slash so don't append if path is empty.
       if (path.length) { // Don't replace empty (child) path with a slash!
-        path = path.replace(/\/+$/, '') + (trailingSlash ? '/' : '') || (isChildWithRelativePath ? '' : '/')
+        path = trailingSlash ? withTrailingSlash(path, true) : withoutTrailingSlash(path, true) || (isChildWithRelativePath ? withoutTrailingSlash(path, true) : withTrailingSlash(path, true))
```
rchl: The `(isChildWithRelativePath ? '' : '/')` could  stay as before, I think. Even if the new logic is equivalent (hard to say for sure), the old one seems easier to read.

Also shouldn't `trailingSlash ? withTrailingSlash(path, true) : withoutTrailingSlash(path, true)` be wrapped in parenthesis? Otherwise the logic might not work exactly as intended.

src/templates/plugin.routing.js
```diff
@@ -86,7 +87,7 @@ function resolveRoute (route, locale) {
       if (isPrefixed) {
         localizedRoute.path = `/${locale}${localizedRoute.path}`
       }
-      localizedRoute.path = localizedRoute.path.replace(/\/+$/, '') + (nuxtOptions.trailingSlash ? '/' : '') || '/'
+      localizedRoute.path = nuxtOptions.trailingSlash ? withTrailingSlash(localizedRoute.path, true) : withoutTrailingSlash(localizedRoute.path, true)
```
rchl, line 91: What about the case of `/` path and `trailingSlash = false`? Does this end up with empty string or `/` (as it should)?

test/browser.test.js
```diff
@@ -178,6 +178,51 @@ describe(`${browserString}, target: static, trailingSlash: true`, () => {
...
+    expect(await (await page.$('body'))?.textContent()).toContain('locale: en')
+    expect(await getRouteFullPath(page)).toBe('/en/?key=value')
+  })
+
+  test('redirects from root (404) path to default locale while preserving query and hash', async () => {
+    const page = await browser.newPage()
+    await page.goto(url('/?key=value#hash'))
+    expect(await (await page.$('body'))?.textContent()).toContain('locale: en')
+    expect(await getRouteFullPath(page)).toBe('/en/?key=value#hash')
+  })
```
rchl: Those tests pass even without the fix so probably not testing the right thing?

rchl (changes requested):

src/templates/plugin.routing.js
divine: It's ending up as `/`

src/helpers/routes.js
divine: Good catch, I will add a parenthesis!

test/browser.test.js
divine: Sorry, there is no problem with the static version, only on the server version.

I will update the tests ASAP.

Thanks!

divine:
Good to go, let me know if anything is left @rchl 

Thanks!

yarn.lock
```diff
@@ -12210,11 +12210,16 @@ ua-parser-js@^0.7.28:
...
-ufo@^0.7.4, ufo@^0.7.5:
+ufo@^0.7.4:
   version "0.7.5"
   resolved "https://registry.yarnpkg.com/ufo/-/ufo-0.7.5.tgz#5d5b2174747c0072edd30501994c8ac4ad2e77ea"
   integrity sha512-FGG+EgguC1oz5dTE1JptPWCyj6Z9mYpwvZY8PTu9Vh/Aoy+Mj9cpeQ3gg4kyEMDbMrH+lTYiw7bomG58B8X7Kg==
 
+ufo@^0.7.5:
+  version "0.7.7"
+  resolved "https://registry.yarnpkg.com/ufo/-/ufo-0.7.7.tgz#0062f9e5e790819b0fb23ca24d7c63a4011c036a"
+  integrity sha512-N25aY3HBkJBnahm+2l4JRBBrX5I+JPakF/tDHYDTjd3wUR7iFLdyiPhj8mBwBz21v728BKwM9L9tgBfCntgdlw==
```
rchl, line 12221: Interesting...
This might end up being an issue because importing `ufo` from plugin files might pick up the version used by Nuxt, which doesn't have the fix. Unless we do a trick like we do for `klona` where we create an alternative alias...

divine: I hope it's what you've asked for.

Thanks!

divine: Aliasing fails, I don't have much knowledge on how to fix those 🤷

Any help @rchl?

Thanks!

rchl: The problem is that `utils-common.js` is used both in the webpack context and in the pure module context where alias is not available.

divine: What can be done?

Thanks!

rchl: I'll try to move stuff around.

rchl (approved):

src/helpers/routes.js
```diff
@@ -1,3 +1,4 @@
+import { withoutTrailingSlash, withTrailingSlash } from 'ufo'
```
divine, line 1: This should `'~i18n-ufo'` @rchl?

Thanks!

test/module.test.js
```diff
@@ -2,6 +2,7 @@ import { resolve } from 'path'
 import { readFileSync } from 'fs'
 import { generate, setup, loadConfig, get, url } from '@nuxtjs/module-test-utils'
 import { JSDOM } from 'jsdom'
+import { withoutTrailingSlash, withTrailingSlash } from 'ufo'
```
divine, line 5: And this as well?

divine (commented):
Looks like we've again `'ufo'` instead of `'~i18n-ufo'`

test/module.test.js
rchl: This is OK in the tests and in the context of the module since then the normal node importing mechanism will ensure that the project-specific version of ufo is picked up.

It's only a problem for webpack context - the files that are added through `addTemplate` or `addPlugin`.

divine: @rchl thank you very much for making it clear!

PR #1237, opened 2021-07-19
refactor(docs): rewrite and structure SEO docs

rchl:


divine:
Just a minor question not related to documentation that I wanted to ask.

Why `addDirAttribute` is by default: `true`?

Basically, it's adding an extra dir attribute, I don't think that more than %98 of nuxt sites need that `rtl` support?

Thanks!

PR #1238, opened 2021-07-19
fix: Upgrade vue-i18n@^8.25.0

pperzyna:
Upgrade vue-i18n to 8.25.0

# Related to:
https://github.com/kazupon/vue-i18n/pull/1321
https://github.com/kazupon/vue-i18n/issues/1108

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "klona": "^2.0.4",
     "lodash.merge": "^4.6.2",
     "ufo": "^0.7.7",
-    "vue-i18n": "^8.24.5"
+    "vue-i18n": "^8.25.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


pperzyna:
@rchl :point_up:

divine:
Hello,

If you don't know `renovate` is configured in this repo and there is a PR https://github.com/nuxt-community/i18n-module/pull/1236 👈  for that.

Thanks!

rchl:
I've merged the renovate PR.

Note that the `vue-i18n` dependency is not pinned so removing and adding back the `nuxt-i18n` package will get you the latest version of `vue-i18n`. That's probably why I'm a bit lazy with updating that dependency in the module itself. But of course this might matter more in case the dependency has fixed some regression.

PR #1239, opened 2021-07-20
refactor!: disable addDirAttribute by default

divine:


rchl:
Please also note it in the migration docs.

rchl:
Also found this place that needs updating: https://github.com/nuxt-community/i18n-module/blob/ffeaa7e41d025329a41f6c387010d9c3806db762/types/vue.d.ts#L17-L21

rchl:
BTW. We can use the jsdoc `@default` tag in the comment.

```ts
  /**
   * Adds a `dir` attribute to the HTML element.
   * @default false
   */
  addDirAttribute?: boolean
```

divine:
> BTW. We can use the jsdoc `@default` tag in the comment.
> 
> ```ts
>   /**
>    * Adds a `dir` attribute to the HTML element.
>    * @default false
>    */
>   addDirAttribute?: boolean
> ```

Update the other one as well?

https://github.com/nuxt-community/i18n-module/blob/ffeaa7e41d025329a41f6c387010d9c3806db762/types/vue.d.ts#L22-L26

rchl:
Yes please.

rchl (approved):

rchl:
Thanks!

PR #1240, opened 2021-07-21
refactor!: remove vuex syncLocale & vuex syncMessages

divine:
Remove vuex making this library less opinionated as there are alternatives and it's up to the user what to integrate.


rchl:
I'm sorry if I confused you at some point making you think that I want to remove the whole vuex option. In https://github.com/nuxt-community/i18n-module/issues/1168#issuecomment-883927367 I was only talking about removing the `syncLocale` and `syncMessages` parts.

Especially the `syncRouteParams` has to stay since there is no other way to achieve what it supports.

divine:
Yup, I misunderstood that, I've removed what you meant now.

Thanks!

PR #1241, opened 2021-07-22
feat: allow disabling route sorting

### LINKED ISSUE #1001: Avoid sorting routes by default
I need to localize my routes and keep the routes order.
More specifically I use 2 regex expressions, and I need one regex path to be evaluated before the other.
``` 
'_country/products/filters/index': {
    it: '/:country/prodotti/:category(.*)?/filtri',
    en: '/:country/products/:category(.*)?/filters',
  },
  '_country/products/index': {
    it: '/:country/prodotti/:category(.*)?',
    en: '/:country/products/:category(.*)?',
  },
```
(parsePages: false)

May be we can avoid ordering routes, or at least not by default.

https://github.com/nuxt-community/i18n-module/blob/master/src/helpers/routes.js#L153
```
try {
    const { sortRoutes } = require('@nuxt/utils')
    localizedRoutes = sortRoutes(localizedRoutes)
  } catch (error) {
    // Ignore
  }
```
Add something like
`if(pageOptions.sortRoutes) {....`

In this way I can define the routes order in `extendRoutes(...)` method of nuxt.config.js file.

Without the possibility to avoid routes sorting, I need to write a custom plugin to reorder routes after your sorting.

divine:
not sure if this should go to the next version or be in master.

Let me know @rchl what do you think.

Closes #1001 

Thanks!

### DIFF
--- a/docs/content/en/options-reference.md
+++ b/docs/content/en/options-reference.md
@@ -106,7 +106,7 @@ When using `prefix_except_default` strategy, URLs for locale specified here won'
 
 ## `sortRoutes`
 
-<badge>v6.27.4+</badge>
+<badge>v6.28.0+</badge>
 
 - type: `boolean`
 - default: `true`
--- a/docs/content/en/options-reference.md
+++ b/docs/content/en/options-reference.md
@@ -111,7 +111,9 @@ When using `prefix_except_default` strategy, URLs for locale specified here won'
 - type: `boolean`
 - default: `true`
 
-Whether to sort routes by using `sortRoutes` from `@nuxt/utils`.
+Whether to sort routes by using the `sortRoutes` function from the `@nuxt/utils` package.
+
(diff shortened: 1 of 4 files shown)


rchl:
It makes sense to make this against master.

divine:
> It makes sense to make this against master.

Master or next?
> rchl changed the base branch from next to master 15 minutes ago
> rchl changed the base branch from master to next 15 minutes ago

Thanks!

rchl:
> Master or next?

master. I've tried to change that but then it includes all changes from `next` so this has to be done manually.

docs/content/en/options-reference.md
```diff
@@ -106,6 +106,13 @@ The app's default locale. Should match code of one of defined `locales`.
 
 When using `prefix_except_default` strategy, URLs for locale specified here won't have a prefix. **It's recommended to set this to some locale** regardless of chosen strategy, as it will be used as a fallback locale when navigating to a non-existent route.
 
+## `sortRoutes` <badge>v6.27.4+</badge>
```
rchl: The badge should be below the header. Otherwise it's included in the link and in the side menu which looks bad.

rchl: Also. This would end up in 6.28.0 since it's a new feature.

divine: Done!

rchl (approved):

rchl:
Thanks!

PR #1243, opened 2021-07-22
fix: custom routes with optional params adjusted incorrectly

### LINKED ISSUE #1242: Routes ending with an optional slug are not working anymore
### Version
nuxt-i18n: 6.27.3
nuxt: 2.15.7

 - [x] universal

### Nuxt-i18n configuration
```
i18n: {
    strategy: 'prefix',
    lazy: true,
    detectBrowserLanguage: false
}
```

### Steps to reproduce

`nuxt.config.js`
```js
i18n: {
    strategy: 'prefix',
    lazy: true,
    detectBrowserLanguage: false
},
router: {
    trailingSlash: true
}
modules: [
    '~/modules/routes'
]
```

`modules/routes.js`
```js
export default async function asyncModule() {
  this.extendRoutes(function extendRoutes(routes, resolve) {
    routes.push({
      name: 'product-page',
      path: `/:product(game|software)/:uid/:slug?`, // <-- notice the optional slug used with "?"
      component: resolve(__dirname, 'pages/product.vue')
    });
  }
}
```

### What is Expected?
1. A route with a path of `/es-ES/:product(game|software)/:uid/:slug?/` should be created and listed in `$nuxt.$router.options.routes`
2. Browsing to `/game/98f76a/cod/` should return the correct page as worked in `nuxt-i18n@6.27.2`

1. A route with a path of `/es-ES/:product(game|software)/:uid/:slug/?` gets created (Notice the `?` got moved to the end of the path) and listed in `$nuxt.$router.options.routes`.
2. Browsing to `/game/98f76a/cod/` returns a `404` error page.

rchl:
Don't use ufo on Vue Router route definitions as those are not
a normal paths, or not always at least.

Fixes #1242

### DIFF
--- a/src/helpers/routes.js
+++ b/src/helpers/routes.js
@@ -1,6 +1,6 @@
 import { STRATEGIES } from './constants'
 import { extractComponentOptions } from './components'
-import { adjustRouteForTrailingSlash, getPageOptions } from './utils'
+import { adjustRouteDefinitionForTrailingSlash, getPageOptions } from './utils'
 
 /**
  * @typedef {import('@nuxt/types/config/router').NuxtRouteConfig} NuxtRouteConfig
@@ -149,7 +149,7 @@ export function makeRoutes (baseRoutes, {
       // - If "router.trailingSlash" is not specified then default to no trailing slash (like Nuxt)
       // - Children with relative paths must not start with slash so don't append if path is empty.
       if (path.length) { // Don't replace empty (child) path with a slash!
-        path = adjustRouteForTrailingSlash(path, trailingSlash, isChildWithRelativePath)
+        path = adjustRouteDefinitionForTrailingSlash(path, trailingSlash, isChildWithRelativePath)
       }
 
       if (shouldAddPrefix && isDefaultLocale && strategy === STRATEGIES.PREFIX && includeUprefixedFallback) {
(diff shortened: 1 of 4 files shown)


rchl:
@divine I guess it wasn't correct or necessary to use ufo in `makeRoutes`. I don't think it was a fix for anything.

divine:
> @divine I guess it wasn't correct or necessary to use ufo in `makeRoutes`. I don't think it was a fix for anything.

Well, I just thought it would be better to handle via ufo instead of manually appending trailing slashes.

Sorry about that!

rchl:
> > @divine I guess it wasn't correct or necessary to use ufo in `makeRoutes`. I don't think it was a fix for anything.
> 
> Well, I just thought it would be better to handle via ufo instead of manually appending trailing slashes.
> 
> Sorry about that!

Not blaming, just discussing to ensure that I'm not reverting actual fix for something.

divine:
> Not blaming, just discussing to ensure that I'm not reverting actual fix for something.

Got it, the actual fix was in https://github.com/nuxt-community/i18n-module/blob/0527d63b99cf30fbe71bd62ded731de3a86798fc/src/templates/plugin.routing.js#L91

You were right that there are too many options and not enough tests for each option which created a regression without knowing it.

Thanks!

PR #1247, opened 2021-07-26
create migration guide from nuxt-i18n and rename to @nuxtjs/i18n

rchl:


rchl:
Will have to restore the badges as the latest plan is to continue the versioning.

rchl:
I guess we don't have to restore the badges as that documentation will only apply to >=v7. <=v6 will have its own documentation.

rchl:
(separated the rename into its own commit)

PR #1274, opened 2021-08-31
feat: add support for query params in canonical url

RobbieTheWagner:
This is a first attempt at adding support for including certain query params in the canonical url. Currently, it uses global options, but I am open to making it work on a page by page basis, I just wasn't sure how to hook that up.

Related to #912

### DIFF
--- a/docs/content/en/api.md
+++ b/docs/content/en/api.md
@@ -73,9 +73,10 @@ All [Vue I18n properties and methods](http://kazupon.github.io/vue-i18n/api/#vue
   - **Returns**: [`MetaInfo`](https://github.com/nuxt/vue-meta/blob/74182e388ad1b1977cb7217b0ade729321761403/types/vue-meta.d.ts#L173)
 
   The `options` object accepts these optional properties:
-  - `addDirAttribute` - Adds a `dir` attribute to the HTML element. Default: `false`
-  - `addSeoAttributes` - Adds various SEO attributes. Default: `false`
-  - `canonicalQueries` - An array of strings corresponding to query params you would like to include in your canonical URL. Default: `[]`
+  - `addDirAttribute` (type: `boolean`) - Adds a `dir` attribute to the HTML element. Default: `false`
+  - `addSeoAttributes` (type: `boolean | SeoAttributesOptions`) - Adds various SEO attributes. Default: `false`
+
+  See also [SEO](../seo).
 
 ## Extension of VueI18n
 
--- a/docs/content/en/api.md
+++ b/docs/content/en/api.md
@@ -75,7 +75,7 @@ All [Vue I18n properties and methods](http://kazupon.github.io/vue-i18n/api/#vue
   The `options` object accepts these optional properties:
   - `addDirAttribute` - Adds a `dir` attribute to the HTML element. Default: `false`
   - `addSeoAttributes` - Adds various SEO attributes. Default: `false`
-  - `canonicalQueries` - An array of strings corresponding to query params you would like to include in your canonical ref url. Default: `[]`
+  - `canonicalQueries` - An array of strings corresponding to query params you would like to include in your canonical URL. Default: `[]`
 
 ## Extension of VueI18n
 
--- a/docs/content/en/api.md
+++ b/docs/content/en/api.md
@@ -75,6 +75,7 @@ All [Vue I18n properties and methods](http://kazupon.github.io/vue-i18n/api/#vue
   The `options` object accepts these optional properties:
   - `addDirAttribute` - Adds a `dir` attribute to the HTML element. Default: `false`
   - `addSeoAttributes` - Adds various SEO attributes. Default: `false`
+  - `canonicalQueries` - An array of strings corresponding to query params you would like to include in your canonical ref url. Default: `[]`
 
 ## Extension of VueI18n
 
--- a/src/templates/head-meta.js
+++ b/src/templates/head-meta.js
@@ -6,7 +6,7 @@ import { formatMessage } from './utils-common'
  * @this {import('vue/types/vue').Vue}
  * @return {import('vue-meta').MetaInfo}
  */
-export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = false, canonicalQueries = [] } = {}) {
+export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = false, canonicalQueries = false } = {}) {
   // Can happen when using from a global mixin.
   if (!this.$i18n) {
     return {}
@@ -129,13 +129,13 @@ export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = fals
     if (canonicalPath) {
       let href = toAbsoluteUrl(canonicalPath, baseUrl)
 
-      if (currentRoute && canonicalQueries?.length) {
+      if (currentRoute && Array.isArray(canonicalQueries) && canonicalQueries.length) {
         const currentRouteQueryParams = currentRoute.query
         const params = new URLSearchParams()
         for (const queryParamName of canonicalQueries) {
           const queryParamValue = currentRouteQueryParams[queryParamName]
           if (queryParamValue && Array.isArray(queryParamValue)) {
-            queryParamValue.forEach(v => params.append(queryParamName, v))
+            queryParamValue.forEach(v => params.append(queryParamName, v || ''))
           }
           if (queryParamValue && typeof queryParamValue === 'string') {
             params.append(queryParamName, queryParamValue)
--- a/src/templates/head-meta.js
+++ b/src/templates/head-meta.js
@@ -131,8 +131,7 @@ export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = fals
 
       if (currentRoute && canonicalQueries?.length) {
         const currentRouteQueryParams = currentRoute.query
-        const url = new URL(href)
-        const params = new URLSearchParams(url.search.slice(1))
+        const params = new URLSearchParams(href.split('?')[1])
         for (const queryParamName of canonicalQueries) {
           const queryParamValue = currentRouteQueryParams[queryParamName]
           if (typeof queryParamValue === 'string') {
--- a/src/templates/head-meta.js
+++ b/src/templates/head-meta.js
@@ -133,15 +133,23 @@ export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = fals
         const currentRouteQueryParams = currentRoute.query
         const params = new URLSearchParams()
         for (const queryParamName of canonicalQueries) {
-          const queryParamValue = currentRouteQueryParams[queryParamName]
-          if (Array.isArray(queryParamValue)) {
-            queryParamValue.forEach(v => params.append(queryParamName, v || ''))
-          } else {
-            params.append(queryParamName, queryParamValue)
+          if (queryParamName in currentRouteQueryParams) {
+            const queryParamValue = currentRouteQueryParams[queryParamName]
+
+            if (Array.isArray(queryParamValue)) {
+              queryParamValue.forEach(v => params.append(queryParamName, v || ''))
+            } else {
+              params.append(queryParamName, queryParamValue)
+            }
           }
         }
-        const queryString = params.toString()
+        let queryString = params.toString()
+
         if (queryString) {
+          // URLSearchParams.append won't let us append just a key, so remove null and undefined values
+          queryString = queryString.replace('=undefined', '')
+          queryString = queryString.replace('=null', '')
+
           href = `${href}?${queryString}`
         }
       }
--- a/src/templates/head-meta.js
+++ b/src/templates/head-meta.js
@@ -131,14 +131,20 @@ export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = fals
 
       if (currentRoute && canonicalQueries?.length) {
         const currentRouteQueryParams = currentRoute.query
-        const params = new URLSearchParams(href.split('?')[1])
+        const params = new URLSearchParams()
         for (const queryParamName of canonicalQueries) {
           const queryParamValue = currentRouteQueryParams[queryParamName]
-          if (typeof queryParamValue === 'string') {
+          if (queryParamValue && Array.isArray(queryParamValue)) {
+            queryParamValue.forEach(v => params.append(queryParamName, v))
+          }
+          if (queryParamValue && typeof queryParamValue === 'string') {
             params.append(queryParamName, queryParamValue)
           }
         }
-        href = `${href}?${params.toString()}`
+        const queryString = params.toString()
+        if (queryString) {
+          href = `${href}?${params.toString()}`
+        }
       }
 
       link.push({
--- a/src/templates/head-meta.js
+++ b/src/templates/head-meta.js
@@ -7,7 +7,7 @@ import { formatMessage } from './utils-common'
  * @param {import('../../types/vue').NuxtI18nHeadOptions} options
  * @return {import('vue-meta').MetaInfo}
  */
-export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = false, canonicalQueries = [] } = {}) {
+export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = false } = {}) {
   // Can happen when using from a global mixin.
   if (!this.$i18n) {
     return {}
@@ -48,7 +48,7 @@ export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = fals
     const locales = /** @type {import('../../types').LocaleObject[]} */(this.$i18n.locales)
 
     addHreflangLinks.bind(this)(locales, this.$i18n.__baseUrl, metaObject.link)
-    addCanonicalLinks.bind(this)(this.$i18n.__baseUrl, metaObject.link)
+    addCanonicalLinks.bind(this)(this.$i18n.__baseUrl, metaObject.link, addSeoAttributes)
     addCurrentOgLocale.bind(this)(currentLocale, currentLocaleIso, metaObject.meta)
     addAlternateOgLocales.bind(this)(locales, currentLocaleIso, metaObject.meta)
   }
@@ -118,8 +118,9 @@ export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = fals
    *
    * @param {string} baseUrl
    * @param {import('../../types/vue').NuxtI18nMeta['link']} link
+   * @param {NonNullable<import('../../types/vue').NuxtI18nHeadOptions['addSeoAttributes']>} seoAttributesOptions
    */
-  function addCanonicalLinks (baseUrl, link) {
+  function addCanonicalLinks (baseUrl, link, seoAttributesOptions) {
     const currentRoute = this.localeRoute({
       ...this.$route,
       name: this.getRouteBaseName()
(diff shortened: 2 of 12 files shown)


divine:
Hello,

We're already using `ufo`, so I'd suggest instead of manually adjusting the URL, use `ufo`'s function https://github.com/unjs/ufo#getquery

Trailing slashes should be covered?

Thanks!

rchl:
> We're already using `ufo`, so I'd suggest instead of manually adjusting the URL, use `ufo`'s function [unjs/ufo#getquery](https://github.com/unjs/ufo#getquery)

Or even use URL but simplify a lot since URL object already contains parsed params (`searchParams`) that can be modified and then the URL can be stringified back.

But before focusing on details, we have to look at the big picture. My initial reaction was that we shouldn't introduce global option like that because we'll end up with a mess later, if we need to also introduce it per-page. And we can't just remove it later since that would be a breaking change. Though maybe global option like that would also be useful to support but then we should think of the best way to do that all together.

One idea I can think of is passing those options as arguments to `$nuxtI18nHead`. Seems like the most natural place to me. Unless there are some issues with that that I can't think of. That said, it would still have to support defining those options per-route since `$nuxtI18nHead` can be used from global places like `nuxt.config.js` or a layout file.

divine:
Well, `$nuxtI18nHead` would be the best choice to add that option and it definitely makes sense!

Thanks!

RobbieTheWagner:
@rchl @divine I defer to whatever you all think is best, I just wanted to get this started as a proof of concept.

divine:
Hello @rwwagner90,

Have you successfully implemented this feature? If not, I could try to take care.

Thanks!

RobbieTheWagner:
@divine I have not had the time to dedicate to this. If you have time, and would like to implement it, please feel free, I would love the help! 😄

RobbieTheWagner:
@divine I updated the PR. Can you please approve the workflow so the tests can run?

divine:
> @divine I updated the PR. Can you please approve the workflow so the tests can run?

Hello, 

I don't have permission to do that as I'm a just a contributor like you :).

I have done the similar work as you but didn't have a chance to complete it as well!

Can you also make canonical attibutes optional (true by default?).

If you want I can publish my branch to take a look.


Thanks!

RobbieTheWagner:
@divine it is optional and defaults to an empty array.

RobbieTheWagner:
@rchl can you approve the workflow please so we can see what tests do?

rchl:
> @rchl can you approve the workflow please so we can see what tests do?

Done.
BTW. You can run tests locally. For example this should run only your new tests:
```sh
yarn test:e2e-ssr -t 'canonical SEO link'
```

divine:
> @divine it is optional and defaults to an empty array.

Hello,

What I mean is to add an extra parameter to make canonical optional (default true):

```javascript
export function nuxtI18nHead ({ addCanonicalAttributes = true ...
```

Later:

```javascript
function addCanonicalLinks (baseUrl, link) {
  if (!addCanonicalAttributes) {
    return
  }
```

And this logic might be better:

```javascript
let href = toAbsoluteUrl(canonicalPath, baseUrl)

if (currentRoute && addCanonicalQueries.length) {
  const params = new URLSearchParams()
  for (const name in addCanonicalQueries) {
    const value = currentRoute.query[name]
    if (Array.isArray(value)) {
      value.forEach(v => params.append(name, v || ''))
    } else {
      params.append(name, value || '')
    }
  }

  href = `${href}?${params.toString()}`
}
```

What do you think @rchl?

Thank!

RobbieTheWagner:
@divine making canonical optional entirely seems out of scope for this PR. This PR is to add the ability to specify which query params should be included in the canonical link.

RobbieTheWagner:
@rchl looks like it needs to be approved again, but I ran it locally and it seemed to be working. Please let me know if I need to make any further changes!

RobbieTheWagner:
I keep getting `implicitly has an 'any[]' type` errors for canonicalQueries, even though I correctly added the types. To appease the linter I will make it false by default, instead of an empty array, but it would be great to not have to make it both a boolean and an array.

RobbieTheWagner:
@rchl I have fixed the lint issues now, if you want to approve again.

docs/content/en/api.md
```diff
@@ -75,6 +75,7 @@ All [Vue I18n properties and methods](http://kazupon.github.io/vue-i18n/api/#vue
   The `options` object accepts these optional properties:
   - `addDirAttribute` - Adds a `dir` attribute to the HTML element. Default: `false`
   - `addSeoAttributes` - Adds various SEO attributes. Default: `false`
+  - `canonicalQueries` - An array of strings corresponding to query params you would like to include in your canonical ref url. Default: `[]`
```
rchl:
```suggestion
  - `canonicalQueries` - An array of strings corresponding to query params you would like to include in your canonical URL. Default: `[]`
```

src/templates/head-meta.js
```diff
@@ -6,7 +6,7 @@ import { formatMessage } from './utils-common'
  * @this {import('vue/types/vue').Vue}
  * @return {import('vue-meta').MetaInfo}
  */
-export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = false } = {}) {
+export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = false, canonicalQueries = false } = {}) {
```
rchl: This can default to `undefined` to avoid the type issue and avoid including `boolean` in the type.

src/templates/head-meta.js
```diff
@@ -127,10 +127,30 @@ export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = fals
...
+        const currentRouteQueryParams = currentRoute.query
+        const params = new URLSearchParams()
+        for (const queryParamName of canonicalQueries) {
+          const queryParamValue = currentRouteQueryParams[queryParamName]
+          if (queryParamValue && Array.isArray(queryParamValue)) {
+            queryParamValue.forEach(v => params.append(queryParamName, v || ''))
+          }
+          if (queryParamValue && typeof queryParamValue === 'string') {
+            params.append(queryParamName, queryParamValue)
+          }
```
rchl: Avoiding extra cycles.

```suggestion
          if (Array.isArray(queryParamValue)) {
            queryParamValue.forEach(v => params.append(queryParamName, v || ''))
          } else {
            params.append(queryParamName, queryParamValue)
          }
```

test/module.test.js
```diff
@@ -1173,6 +1173,22 @@ describe('prefix_and_default strategy', () => {
...
+    expect(links[0].getAttribute('href')).toBe('nuxt-app.localhost/?foo=%22bar%22')
+  })
+
+  test('canonical SEO link does not include query params not in canonicalQueries', async () => {
+    const html = await get('/?bar="baz"')
+    const dom = getDom(html)
+    const links = dom.querySelectorAll('head link[rel="canonical"]')
+    expect(links.length).toBe(1)
+    expect(links[0].getAttribute('href')).toBe('nuxt-app.localhost/')
+  })
```
rchl: Could you also test this on a page-level? For example:

```js
  test('canonical SEO link include query params in canonicalQueries on page level', async () => {
    const html = await get('/about-us?foo=baz&page=1')
    const dom = getDom(html)
    const links = dom.querySelectorAll('head link[rel="canonical"]')
    expect(links.length).toBe(1)
    expect(links[0].getAttribute('href')).toBe('nuxt-app.localhost/about-us?page=1')
  })
```

and then in `about.vue` return:
```js
return this.$nuxtI18nHead({ addSeoAttributes: true, canonicalQueries: ['page'] })
```

types/vue.d.ts
```diff
@@ -24,6 +24,11 @@ interface NuxtI18nHeadOptions {
    * @default false
    */
   addSeoAttributes?: boolean
+  /**
+   * An array of strings corresponding to query params you would like to include in your canonical ref url.
```
rchl:
```suggestion
   * An array of strings corresponding to query params you would like to include in your canonical URL.
```

rchl (changes requested):
The paragraph at the bottom of the https://i18n.nuxtjs.org/seo documentation page should also be updated as it got outdated with those changes

src/templates/head-meta.js
RobbieTheWagner: @rchl I don't think this is equivalent. Doing an `if/else` means if there is no `queryParamValue` or the type is not a string, it's going to append it still.

src/templates/head-meta.js
RobbieTheWagner: @rchl I tried defaulting to undefined and it has type issues too. I've never had to deal with a vanilla JS file complain about types before, so I am not sure how to resolve it.

test/module.test.js
RobbieTheWagner: Added a page level test 👍

src/templates/head-meta.js
divine: Why do you even check for type string?

divine:
> @divine making canonical optional entirely seems out of scope for this PR. This PR is to add the ability to specify which query params should be included in the canonical link.

Why not? It won't take you a minute, anyways, I'll do on my own then.

Thanks!

src/templates/head-meta.js
rchl: The type can only be a string or an array so it's redundant to check for the string type if already checked for array.

As for empty queryParamValue, maybe add a test for it? A query without the value (like `/?foo` should still be preserved.

RobbieTheWagner:
> Why not? It won't take you a minute, anyways, I'll do on my own then.

@divine just because it is not related to this PR at all. This is to add support for adding query params to canonical urls. Once it is in, I would be glad to help with any additional options you think need to be added.

src/templates/head-meta.js
RobbieTheWagner: @rchl It is impossible to use `append` with URLSearchParams and pass no value, so we end up with a bunch of `?foo=undefined`

RobbieTheWagner:
@rchl I tried to support setting `?foo` with no value, but URLSearchParams does not support that, so I added some hacks to remove the null and undefined ones later:

```js
   queryString = queryString.replace('=undefined', '')
   queryString = queryString.replace('=null', '')
 ```

Let me know if you have any better ideas. I think checking that the value exists and is a string is maybe good, to avoid these hacks. I would assume people don't use query params with no value often?

File: src/templates/head-meta.js [hunk 6]
```diff
@@ -127,10 +127,37 @@ export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = fals
...
+              params.append(queryParamName, queryParamValue)
+            }
+          }
+        }
+        let queryString = params.toString()
+
+        if (queryString) {
+          // URLSearchParams.append won't let us append just a key, so remove null and undefined values
+          queryString = queryString.replace('=undefined', '')
+          queryString = queryString.replace('=null', '')
```
rchl: I have said to remove the string type check because according to VueRouter types, the queryParamValue can only be a string or an array of string or null. But if that's not the case and the types are wrong then I'm fine with instead changing the `params.append(queryParamName, queryParamValue)` line to `params.append(queryParamName, queryParamValue || '')` for example.

divine: @rwwagner90 you could have just used suggested logic... not sure why you didn't use it at first instead changed it multiple times and are now back to the initial one.

RobbieTheWagner: @divine I have explained my reasons, and I do not appreciate the hostility here. Again, the suggestions of having a plain else block does not work. A url of `?foo` will end up being `?foo=null`. Your code did not handle this case either.

RobbieTheWagner: > I have said to remove the string type check because according to VueRouter types, the queryParamValue can only be a string or an array of string or null. But if that's not the case and the types are wrong then I'm fine with instead changing the `params.append(queryParamName, queryParamValue)` line to `params.append(queryParamName, queryParamValue || '')` for example.

@rchl Thank you for clarifying! I believe the issue is when it's undefined like `?foo` as you say here undefined is not a possible value, so it converts it to null, then we end up with `?foo=null` because URLSearchParams does not support setting keys with no values.

RobbieTheWagner:
@rchl is there anything else I need to update here to get this PR in?

src/templates/head-meta.js
rchl: But then `|| ''` should help, no?

src/templates/head-meta.js
rchl: I've tested this when making the comment and `canonicalQueries = undefined` here and changing the type to `canonicalQueries?: string[]` worked for me.

I want to avoid the boolean because value `true` makes no sense in that option.

src/templates/head-meta.js
RobbieTheWagner: @rchl no unfortunately that does not help. `params.append(queryParamName, queryParamValue || '')` just means you will end up with `nuxt-app.localhost/?foo=` so you have an extra equals there and no value and you requested that when the value is `undefined` we instead have `?foo` not `?foo=`. Personally, I wouldn't support the `?foo` case where there is a key and no value, but I added the hacks to support this:

```js
       if (queryString) {
          // URLSearchParams.append won't let us append just a key, so remove null and undefined values
          queryString = queryString.replace('=undefined', '')
          queryString = queryString.replace('=null', '')

          href = `${href}?${queryString}`
        }
```

rchl:
The part where I asked to change the types of `canoncialQueries` is not addressed.

src/templates/head-meta.js
rchl: I'm fine with including the `=` because it won't make any difference from the perspective of parsing the query params.

rchl:
Also this is not addressed: https://github.com/nuxt-community/i18n-module/pull/1274#pullrequestreview-765853041

src/templates/head-meta.js
rchl: I've pushed a fix for the types.

src/templates/head-meta.js
rchl: I've pushed a fix for this.

rchl:
I'm also thinking about folding the new `canonicalQueries` option into the `addSeoAttributes` one. Would do that by accepting an object for `addSeoAttributes` with more detailed configuration.

It feels a bit wrong now because `canonicalQueries` is a separate option but it really depends on `addSeoAttributes` being `true`.

I can do that change myself if you are feeling like I'm pushing too much. :)

src/templates/head-meta.js
```diff
@@ -139,17 +138,14 @@ export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = fals
             if (Array.isArray(queryParamValue)) {
               queryParamValue.forEach(v => params.append(queryParamName, v || ''))
             } else {
-              params.append(queryParamName, queryParamValue)
+              params.append(queryParamName, queryParamValue || '')
```
RobbieTheWagner, line 144: @rchl as mentioned before, this is incorrect and gives the wrong output.

RobbieTheWagner: An undefined value will give you `website.com?foo=` which is not the `website.com?foo` you had requested and I think is an invalid case. If there is an equals, I think you need a value?

RobbieTheWagner:
> The part where I asked to change the types of `canoncialQueries` is not addressed.

@rchl I see you pushed a change to fix the types, thanks! I could not figure it out 😬 

> Also this is not addressed: [#1274 (review)](https://github.com/nuxt-community/i18n-module/pull/1274#pullrequestreview-765853041)

Could you elaborate on what is needed here please?

rchl:
> Could you elaborate on what is needed here please?

It says:

> When the page contains the query parameters, the canonical link will not include query params. ... Note that there is currently no way to override that in case that including a specific query param would be desired.

So it's obviously outdated now. Needs to be changed to mention how to use the new functionality to enable that.

src/templates/head-meta.js
RobbieTheWagner: @rchl if you would like `?foo` to generate a canonical of `?foo` instead of `?foo=` we need to put back my changes. I would expect the canonical to match the given query param exactly, not to add an `=`.

RobbieTheWagner:
@rchl I updated those docs. We still need to address this issue https://github.com/nuxt-community/i18n-module/pull/1274#discussion_r725038739

src/templates/head-meta.js
```diff
@@ -124,13 +125,35 @@ export function nuxtI18nHead ({ addDirAttribute = false, addSeoAttributes = fals
...
 
-    const canonicalPath = currentRoute ? currentRoute.path : null
+    if (currentRoute) {
+      let href = toAbsoluteUrl(currentRoute.path, baseUrl)
+
+      if (canonicalQueries.length) {
+        const currentRouteQueryParams = currentRoute.query
+        const params = new URLSearchParams()
+        for (const queryParamName of canonicalQueries) {
+          if (queryParamName in currentRouteQueryParams) {
```
divine, line 138: Why do you need this double-check over here?

RobbieTheWagner: Because we are iterating through the query params we have defined that should be included, so for example `canonicalQueries: ['foo']` would give us a `queryParamName` of `'foo'` here. If we do not check that the `currentRouteQueryParams` actually has a `foo`, then every route will get a `?foo=null` appended to it. This `if` is necessary.

divine: Not necessarily as null value is covered here:

`params.append(queryParamName, queryParamValue || '')`

divine:
Query strings which behavior is kinda weird and my suggestion would be to use something like https://github.com/sindresorhus/query-string which covers all of that problems.

That means depending on another library as well... but at least it will make things stable.

Forgot to add, that this should be configurable `skipEmptyString` `skipNull`.

src/templates/head-meta.js
RobbieTheWagner: Trust me, we need it and I disagree with adding the `|| ''` in the first place, for all the reasons I have stated many times.

RobbieTheWagner:
> Query strings which behavior is kinda weird and my suggestion would be to use something like https://github.com/sindresorhus/query-string which covers all of that problems.
> 
> That means depending on another library as well... but at least it will make things stable.
> 
> Forgot to add, that this should be configurable `skipEmptyString` `skipNull`.

I feel we should just use the code as I wrote it, which strips null and undefined, but I don't feel super strongly. Technically, query params are *always* undefined if you don't define them, so I consider that an invalid edge case.

RobbieTheWagner:
@rchl @divine you seem to have strong opinions about including `|| ''` and changing the way I had things to support `?foo`. I really don't care too much, and just want to get this feature in. Please let me know what needs to be done to get this merged. I will never use an invalid query param with no value like `?foo` anyway, so this PR is ready to go for our use cases 😄

src/templates/head-meta.js
rchl: I'm fine with having `=` since there is no difference in parsed value whether the `=` is included or not.

```
const params = new URLSearchParams('?a&b=')
params.get('a') => ""
params.get('b') => ""
```

RobbieTheWagner: @rchl but if you were at `bar.com?foo` and canonical were set to `bar.com?foo=` your canonical would be different than the url you are on then, which we don't want.

rchl: I have no evidence to prove anything but I would think that this should make no difference. Google or other search engines probably have to parse the URL (and especially queries). It can't do just a plain string comparison as the query params could be in a different order, for example. So if it's looking at parsed representation then I would think that it shouldn't matter.

For now I would go with a simpler solution and if there is evidence that it causes issues then we can fix that later.

RobbieTheWagner: Okay, please feel free to merge whenever you'd like. This edge case doesn't matter to our app anyway, as we will only have actually valid query params, and `?foo` with no value is not a valid query param.

rchl:
Sorry for the delay.
I will now also attempt to merge the new option into `addSeoAttributes` and push the changes here.

rchl:
@rwwagner90 please have a look and let me know if you are fine with my changes.

RobbieTheWagner:
@rchl looks good to me! 🚢

RobbieTheWagner:
Thanks for merging @rchl! When do you think a new release will be out?

rchl:
Released https://github.com/nuxt-community/i18n-module/releases/tag/v7.1.0

Issue #1100, opened 2021-03-09
Support multiple domains in local options


jinevo:
### Problem 
Currently the only way to support multiple domains for one code is to declare each domain separately. 
At the first sight, it could be managed by env variables but it is useful when we want to progressively migrate a website to nuxt.

```
{
  locales: [
    {
      code: 'fr',
      domain: 'domain.fr',
      file: 'fr.js'
    },
    {
      code: 'fr',
      domain: 'fr.nuxtprod-domain.com',
      file: 'fr.js'
    },
    {
      code: 'fr',
      domain: 'fr.nuxtuat-domain.com',
      file: 'fr.js'
    },
    {
      code: 'en',
      domain: 'domain.com',
      file: 'en.js'
    },
    {
      code: 'en',
      domain: 'en.nuxtprod-domain.com',
      file: 'en.js'
    },
    {
      code: 'en',
      domain: 'en.nuxtuat-domain.com',
      file: 'en.js'
    },
   // and so on
  ],
  differentDomains: true
}
```

Besides, this implementation generates duplicated routes definition warnings in the console / terminal from vue-router
`WARN  [vue-router] Duplicate named routes definition: { name: "index___en", path: "" }`

### Solution
The solution would be to support an array of domains in the domain value

```
{
  locales: [
    {
      code: 'fr',
      domain: ['domain.fr', 'fr.nuxtprod-domain.com', 'fr.nuxtuat-domain.com'],
      file: 'fr.js'
    },
    {
      code: 'en',
      domain: ['domain.com', 'en.nuxtprod-domain.com', 'en.nuxtuat-domain.com'],
      file: 'en.js'
    },
   // and so on
  ],
  differentDomains: true
}
```

larbijirari:
hello, 
i'm facing the same situation, any news about this feature ?

BobbieGoede:
Closing this as v7 is not being actively worked on (critical hotfixes only), if this still applies to v8 and later please open a new issue!

PR #1278, opened 2021-09-07
feat: Multi domain support proposal per locale

themaxsandelin:
# Support for multiple domains per locale
This is a feature we needed in our use of Nuxt i18n. We are hosting a Nuxt app on multiple domains, that then shows specific content depending on the domain, and we of course needed localization support.

So what this allows you to do is pass in a "new" parameter in each locale object called `domains`, as an alternative to the already existing `domain` parameter. In this case you provide an Array of strings, being each domain you want to match to the locale. We then match that locale to the host the app is served on.

One thing to note, while building this, was that I had to make a choice in how we determine the domain for the complementing locales on the app. Meaning, if you are hosting your app on the domains `nuxtprod-domain.com` and `nuxttest-domain.com`, you are visiting the app on `nuxtprod-domain.com`, and your locales look like this:
```javascript
locales: [
  {
    {
      code: 'en',
      domains: ['nuxtprod-domain.com', 'nuxttest-domain.com'],
      file: 'en.js'
    },
    {
      code: 'fr',
      domains: ['fr.nuxtprod-domain.com', 'fr.nuxttest-domain.com'],
      file: 'fr.js'
    }
  }
]
```
How do we find the correct url for the complementing locale? Well, I though about using simple comparison of the sub-domain and top domain without the extension. However, in some cases, this might not be the desired outcome. The best solution I found was to do a string comparison between the current host and the different domains provided, and use the one that matches best. But I am open to suggestions into how we can do this differently. Especially since this introduced a new dependency ([string-similarity](https://www.npmjs.com/package/string-similarity)).

## Updating the docs
I didn't take the time to update the docs yet, since I'm not sure this will be accepted or if any changes will be proposed. But I will of course take the time to update them if this proposal is accepted.

Solves issue #1100.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "js-cookie": "^3.0.0",
     "klona": "^2.0.4",
     "lodash.merge": "^4.6.2",
+    "string-similarity": "^4.0.4",
     "ufo": "^0.7.9",
     "vue-i18n": "^8.25.0"
   },
--- a/src/templates/plugin.main.js
+++ b/src/templates/plugin.main.js
@@ -357,10 +357,20 @@ export default async (context) => {
       for (const [index, locale] of options.normalizedLocales.entries()) {
         const domain = store.state.localeDomains[locale.code]
         if (domain) {
-          locale.domain = domain
+          if (Array.isArray(domain)) {
+            // Map over the Array to avoid issue with Klona using a reactive Array.
+            locale.domains = domain.map(dom => dom)
+          } else {
+            locale.domain = domain
+          }
           const optionsLocale = options.locales[index]
           if (typeof (optionsLocale) !== 'string') {
-            optionsLocale.domain = domain
+            if (Array.isArray(domain)) {
+              // Map over the Array to avoid issue with Klona using a reactive Array.
+              optionsLocale.domains = domain.map(dom => dom)
+            } else {
+              optionsLocale.domain = domain
+            }
           }
         }
       }
--- a/src/templates/plugin.utils.js
+++ b/src/templates/plugin.utils.js
@@ -1,4 +1,7 @@
+// @ts-ignore
 import isHTTPS from 'is-https'
+// @ts-ignore
+import { findBestMatch } from 'string-similarity'
 import { localeMessages, options } from './options'
 import { formatMessage, getHost } from './utils-common'
 // @ts-ignore
@@ -105,6 +108,16 @@ export function getDomainFromLocale (localeCode, req, { normalizedLocales }) {
       const host = getHost(req)
       if (host) {
         domain = lang.domains.find(domain => domain === host)
+        if (!domain) {
+          if (lang.domains.length === 1) {
+            domain = lang.domains[0]
+          } else {
+            const stringMatch = findBestMatch(host, lang.domains)
+            if (stringMatch && stringMatch.bestMatch && stringMatch.bestMatch.target) {
+              domain = stringMatch.bestMatch.target
+            }
+          }
+        }
       }
     } else {
       domain = lang.domain
--- a/src/templates/plugin.utils.js
+++ b/src/templates/plugin.utils.js
@@ -1,6 +1,6 @@
 import isHTTPS from 'is-https'
 import { localeMessages, options } from './options'
-import { formatMessage } from './utils-common'
+import { formatMessage, getHost } from './utils-common'
 // @ts-ignore
 import { hasProtocol } from '~i18n-ufo'
 
@@ -99,17 +99,29 @@ export function resolveBaseUrl (baseUrl, context, localeCode, { differentDomains
 export function getDomainFromLocale (localeCode, req, { normalizedLocales }) {
 // Lookup the `differentDomain` origin associated with given locale.
   const lang = normalizedLocales.find(locale => locale.code === localeCode)
-  if (lang && lang.domain) {
-    if (hasProtocol(lang.domain)) {
-      return lang.domain
-    }
-    let protocol
-    if (process.server) {
-      protocol = (req && isHTTPS(req)) ? 'https' : 'http'
+  if (lang && (lang.domain || lang.domains)) {
+    let domain
+    if (lang.domains) {
+      const host = getHost(req)
+      if (host) {
+        domain = lang.domains.find(domain => domain === host)
+      }
     } else {
-      protocol = window.location.protocol.split(':')[0]
+      domain = lang.domain
     }
-    return `${protocol}://${lang.domain}`
+
+    if (domain) {
+      if (hasProtocol(domain)) {
+        return domain
+      }
+      let protocol
+      if (process.server) {
+        protocol = (req && isHTTPS(req)) ? 'https' : 'http'
+      } else {
+        protocol = window.location.protocol.split(':')[0]
+      }
+      return `${protocol}://${domain}`
+    }
   }
 
   // eslint-disable-next-line no-console
--- a/src/templates/utils-common.js
+++ b/src/templates/utils-common.js
@@ -84,6 +84,24 @@ export function matchBrowserLocale (appLocales, browserLocales) {
 }
 
 /**
+ * Get the current host based on either window.location on client side or request header on server side.
+ *
+ * @param  {import('http').IncomingMessage | undefined} req
+ * @return {string | undefined} Host if found any
+ */
+export function getHost (req) {
+  /** @type {string | undefined} */
+  let host
+  if (process.client) {
+    host = window.location.host
+  } else if (req) {
+    const detectedHost = req.headers['x-forwarded-host'] || req.headers.host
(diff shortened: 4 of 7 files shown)


rchl:
I would rather overload the existing `domain` property with support for an array of items rather than creating a new property.

> How do we find the correct url for the complementing locale? Well, I though about using simple comparison of the sub-domain and top domain without the extension.

I feel like this can only work robustly if there is an explicit mechanism for correlating domains. For example by using some arbitrary `key` attached to each domain. So for example

```
locales: [
  {
    {
      code: 'en',
      domains: [
        { domain: 'nuxtprod-domain.com', key: 'd1' },
        { domain: 'nuxttest-domain.com', key: 'd2' }
      ],
      file: 'en.js'
    },
    {
      code: 'fr',
      domains: [
        { domain: 'fr.nuxtprod-domain.com', key: 'd1' },
        { domain: 'fr.nuxttest-domain.com', key: 'd2' },
      ],
      file: 'fr.js'
    }
  }
]
```

Don't like the verbosity of that too much but if someone really needs that functionality, they will probably don't mind setting it up once.

We could also support just a plain array of strings like in your code example in which case the first domain would always be used for the complementary locale.

themaxsandelin:
@rchl Thank you for the quick feedback!

> I would rather overload the existing domain property with support for an array of items rather than creating a new property.

Roger that! 👍

> Don't like the verbosity of that too much but if someone really needs that functionality, they will probably don't mind setting it up once.

I like the idea of being able to "link" domains together. If for some reason one would use very different top-level domains, it would be good to be able to link them together in order to mainly support the seo tags.

One question. If we have one or more locales that have fewer domain alternatives than another locale, how would we handle the seo tags / alternative locales? Should we simply delete them from the i18n object in that instance? So that the only locales that are available as alternatives, are the ones that has a corresponding (or linked) domain? Ex:

```javascript
locales: [
  {
    {
      code: 'en',
      domains: [
        { domain: 'nuxtprod-domain.com', key: 'd1' },
        { domain: 'nuxttest-domain.com', key: 'd2' }
      ],
      file: 'en.js'
    },
    {
      code: 'fr',
      domains: [
        { domain: 'fr.nuxtprod-domain.com', key: 'd1' },
        { domain: 'fr.nuxttest-domain.com', key: 'd2' },
      ],
      file: 'fr.js'
    },
    {
      code: 'sp',
      domains: [
        { domain: 'sp.nuxtprod-domain.com', key: 'd1' },
      ],
      file: 'sp.js'
    }
  }
]
```
In the case of loading the app on `nuxttest-domain.com`, logically `sp` should not be an alternative locale since it doesn't have a locale domain for nuxttest?

divine:
>  How do we find the correct url for the complementing locale? Well, I though about using simple comparison of the sub-domain and top domain without the extension. However, in some cases, this might not be the desired outcome. The best solution I found was to do a string comparison between the current host and the different domains provided, and use the one that matches best. But I am open to suggestions into how we can do this differently. Especially since this introduced a new dependency (string-similarity).

I don't like the usage of finding the best match since it's an array that could be looked up?

What are some cases that might not be the desired outcome? Can you give an example?

Another question, how you'd integrate this feature in a static rendering? How that will work?

Thanks!

themaxsandelin:
@divine Not sure if I understand you correctly in your dislike of best match due to it being an Array. What I refer to when saying the "best match" is that we take all the domains defined in a locale, and match them to the current host, to see which one matches the best. However, I find @rchl's proposal of either using the Array index or perhaps linking the domains using a key property to be more reliable.

I have not tested this on a static site yet. Do you see any immediate implications to the change? I could set one up locally and test it out to see if it works as expected.

rchl:
> In the case of loading the app on `nuxttest-domain.com`, logically `sp` should not be an alternative locale since it doesn't have a locale domain for nuxttest?

Maybe in that case it should pick the first key-less domain.

And if there is none then completely ignore that locale in terms of SEO. Although... that sounds like a new logic that we didn't have before which might require substantial changes.

themaxsandelin:
>  Although... that sounds like a new logic that we didn't have before which might require substantial changes.

What would we use as SEO tags for a locale that has no domain (string or array) set in this case? More specifically the `<link rel="alternative" />` tags. Shouldn't they be ignored / excluded in the case of a locale without a domain? We could still have the `<meta property="og:locale:alternate" />` tags.

rchl:
> > Although... that sounds like a new logic that we didn't have before which might require substantial changes.
> 
> What would we use as SEO tags for a locale that has no domain (string or array) set in this case? More specifically the `<link rel="alternative" />` tags. Shouldn't they be ignored / excluded in the case of a locale without a domain? We could still have the `<meta property="og:locale:alternate" />` tags.

Yes, I think should be ignored.

divine:
Hello,

> @divine Not sure if I understand you correctly in your dislike of best match due to it being an Array. What I refer to when saying the "best match" is that we take all the domains defined in a locale, and match them to the current host, to see which one matches the best. However, I find @rchl's proposal of either using the Array index or perhaps linking the domains using a key property to be more reliable.

I dislike best match as it could get different results since it isn't looking exactly at what it should. Using what @rchl  suggested is a much better choice.

> I have not tested this on a static site yet. Do you see any immediate implications to the change? I could set one up locally and test it out to see if it works as expected.

Static site would generate only one version (default locale) AFAIK.

---

I still see no point of it all. Who on the world would use the same language on different domains?

So basically adding this complex unneeded feature to use only for "staging/testing" domains?

Shouldn't you set that via eniroment configuration when deploying?

Thanks!

themaxsandelin:
> I dislike best match as it could get different results since it isn't looking exactly at what it should. Using what @rchl suggested is a much better choice.

Agreed! 👍

> Static site would generate only one version (default locale) AFAIK.

All right. Will implement the feature in a statically generated site to see how it works.

> I still see no point of it all. Who on the world would use the same language on different domains?

The examples I gave were not "realistic" in this sense. But for our use case, we are building a frontend for a new checkout for our different storefronts. This checkout will be deployed using one deployment, and then have multiple domains pointing to it. This is so that we can more easily scale up and handle the addition of more brands/storefronts as well as load on the app.

Meaning we will need to first off know which brand to cater to based on the host header, as well as which language.

This might be a special case, but I'm sure there are similar situations. And to me, this doesn't feel like an unnecessarily complex feature.

divine:
Hello,

> The examples I gave were not "realistic" in this sense. But for our use case, we are building a frontend for a new checkout for our different storefronts. This checkout will be deployed using one deployment, and then have multiple domains pointing to it. This is so that we can more easily scale up and handle the addition of more brands/storefronts as well as load on the app.

This shouldn't be the scope of this module as you are trying to add complexity for only your use case. You'd better waiting multi domain deployment support in nuxt itself.

> feat: Multi domain support proposal for one locale

This should be the title instead ^^


> This might be a special case, but I'm sure there are similar situations. And to me, this doesn't feel like an unnecessarily complex feature.

There are issues with the multi domain support and you creating another layer of complexity for only one use case and this won't work with static version at all.

Just my 2 cents.

It's up to @rchl decide but I don't feel it make sense at all to add this feature.

Thanks!

themaxsandelin:
> This shouldn't be the scope of this module as you are trying to add complexity for only your use case. You'd better waiting multi domain deployment support in nuxt itself.

I did look to see if there were someone else in need of this feature, and there was. The issue was linked in the PR. Just to ensure that I'm not doing this simply to add the feature for our use case. That was just an example. This simply made sense to me, as I can see the use cases of having an app hosting multiple domains.

> feat: Multi domain support proposal for one locale

LGTM, will change!

> There are issues with the multi domain support and you creating another layer of complexity for only one use case and this won't work with static version at all.

I understand your concern. I'll try it out for myself like I said, but I'm guessing by your statement that you did try it yourself. Regardless, will await @rchl's verdict. :)

coremyslo:
Would this workaround fit instead?
```
{
    locales: [
        {
            code: 'en',
            file: 'en.js'
        },
        {
            code: 'fr',
            file: 'fr.js'
        },
        {
            code: 'en-prod',
            domain: 'nuxtprod-domain.com',
            file: 'en-prod.js'
        },
        {
            code: 'en-test',
            domain: 'nuxttest-domain.com',
            file: 'en-test.js'
        },
        {
            code: 'fr-prod',
            domain: 'fr.nuxtprod-domain.com',
            file: 'fr-prod.js'
        },
        {
            code: 'fr-test',
            domain: 'fr.nuxttest-domain.com',
            file: 'fr-test.js'
        },
    ],
    vueI18n: {
        fallbackLocale: {
            'en-prod': ['en'],
            'en-test': ['en'],
            'fr-prod': ['fr', 'en'],
            'fr-test': ['fr', 'en'],
        }
    }
}
```

themaxsandelin:
@coremyslo Does that imply that you have to have the duplicate .js files for it to work? Or does it default to the fallback locale files?

larbijirari:
Hi @themaxsandelin ,
Any news about this feature ?

PR #1281, opened 2021-09-10
Fix adding duplicated routes when users were using `router.extendRoutes`

### LINKED ISSUE #1280: Not playing nice with `router.extendRoutes`
Some projects will fetch data from an external source and use `router.extendRoutes` to create pages based on the source's response contents. In my case I fetch the data from Sanity and, given a `_type` and `slug`, I'll generate a page using one of my [ignored page components](https://nuxtjs.org/docs/2.x/directory-structure/pages#ignoring-pages).

Before adding `@nuxtjs/i18n` to my project, I was able to pass an array of route objects and Nuxt was using to create the pages I wanted. With this module enabled, Nuxt will try to render all of those pages for each of the available locales.

This is what doesn't work as expected (plus I couldn't find any examples of using the module like this). I have a PR ready for review (#1281) that is solving this issue on my local env. Since I'm not familiar with the codebase and all the ins and outs of the module I'm opening this issue first.

### Version
@nuxtjs/i18n: v7.0.3
nuxt: v2.15.8

### Nuxt configuration

 - [ ] Applies to a site deployed to a server with a Node backend

```js
// nuxt.config.js

const mapPages = ([key, { component, result }]) => ([key, { component: join(__dirname, component), result }])
const pages = { // on my codebase this is a groq query to Sanity
  "legal": {
    "component": "./pages/-legal.vue",
    "result": [
      {
        "_id": "unique-id-value-for-the-document",
        "lang": "en",
        "slug": "cookies-policy"
      },
      {
        "_id": "i18n.unique-id-value-for-the-document.es",
        "lang": "es",
        "slug": "politica-cookies"
      }
    ]
  }
}

const dynamicPages = Object.fromEntries(Object.entries(pages).map(mapPages))

export default {
  buildModules: [
    ['@nuxtjs/i18n', {
      locales: ['en', 'es'],
      defaultLocale: 'en',
      detectBrowserLanguage: false,
      vueI18n: {
        fallbackLocale: 'en',
        messages: {
          en: { welcome: 'Welcome' },
          es: { welcome: 'Bienvenido' },
        },
      },
    }],
  ],

  router: {
    extendRoutes (routes) {
      const buildRoutes = ([key, { result, component }]) => result.map(({ lang, slug, _id }) => {
        const name = _id.includes('i18n.') ? result.find((i) => i._id === _id.split('.')[1]).slug : slug
        const route = _id.includes('i18n.') ? `/${lang}/${slug}` : `/${slug}`
        return { name: `${name}___${lang}`, path: route, component, chunkName: `pages/${key}`, url: route }
      })

      return sortRoutes([
        ...routes,
        ...Object.entries(dynamicPages).map(buildRoutes).flat(),
      ])
    },
  },
}
```

* Dont' mind the unnecessary `Object.entries(Object.fromEntries())`, my config is split in two parts and one of them is shared accross different projects and I just copy/pasted the relevant parts for the issue.

### Reproduction Link
https://codesandbox.io/s/test-nuxt-i18n-forked-2us3p

### Steps to reproduce
1. Open the reproduction link
2. Open the browser URL from CodeSandbox on a new tab
3. Open the routes section inside `vue-devtools`, you should se something like this:

Captura de pantalla 2021-09-10 a las 15 52 19

### What is Expected?
I expect `@nuxtjs/i18n` to either:
- allow me to pass an array of valid routes without messing with them
- handle everything automagically for me (this might be hard to implement and would require users to provide those routes with everything the module could need to make it work)

The module is assuming each of those routes needs their localized counter parts and pushing them to the router. This breaks the theme-switcher functionality and pushes non-existent pages to the router, resulting in one build error for each of those.

It also prefixes the paths with each locale, resulting in some cases in a duplicated locale. Looks like this only happens because the module doesn't expect someone to provide custom routes.

---

I tried with different name routes, including/excluding the locale on the name route, and a bunch of other things. The approach shown on the config section of this issue is what made the most sense to me since it consist of manually writing the same route Nuxt and the module would generate for me.

hacknug:
This shouldn't be merged without review for obvious reasons. An explanation for the solution provided in this PR can be found in issue #1280.

The idea here is to check if the route already contains `routesNameSeparator`. In that case it probably means the user already anticipated the lack of support for this and maybe other features and we can (hopefully and) safely skip patching the route object with redundant data.

This was adding the same route a couple times so I added a conditional to check if a route with the same name already exists. If that's the case, we only push the new route when it has a `children` property because from what I saw, the module only localizes dynamic children routes in those cases.

Works fine on my local enviroment but I'm pretty sure this isn't taking into account all of the cases this module supports. Let me know if you  want me to make any changes or add any extra details.

Closes #1280.


hacknug:
not stale

rchl:
This change is preventing prefixes from being added in some (all?) cases. See the unittesting results.

Also I'd like to see at least one new test added to test the case that you are fixing.

Nakroma:
Not stale, we have the same issue. Any news on this PR or something others could help with?

hacknug:
@Nakroma I'd say adding a failing test to demonstrate how this is broken in the current version of the module would be extremely helpful and get us one step closer to implementing a proper solution for it. Once we have that in place we can re-implement some of the changes in this PR in a way it doesn't affect any of the existing features of the module (this PR currently is making some of the tests fail). An explanation about your use case and how the current implementation is not working for you would also be helpful in case we need to consider a different scenario from the one I tried to explain in #1280.

I've had this in production since I opened the PR and it works fine for my use case. There's some issues with some routes on the sitemap but I haven't been able to put in the hours to research what's really going on and who's the culprit.

Nakroma:
When trying to replicate our issue for the failtest I took another look at our problem and it turns out it was an issue with nuxt itself that was causing the route duplication, i18n was just obfuscating it, so our problems seems to be a different one from the one mentioned in #1280

hacknug:
Not stale

BobbieGoede:

PR #1319, opened 2021-10-14
Feat: expose hook extend locales

### LINKED ISSUE #1308: Expose a nuxt hook when translation file is ready to be injected
I am developing an internal ui library and I need to have some translations inside it and for now I rely on slots
But, as the project grows, I'm copy pasting `$t('actions.submit')` (and many others) over and over.
I want my components to have a **default slot with translated content** that could be overridden by theirs parents.

As I already do for merging my components into the nuxt app or with tailwindcss when I merge the configs together, I'm wondering if something similar is doable with the i18n module?
```javascript
  // auto import components into the nuxt app
  nuxt.hook('components:dirs', (dirs) => {
    // Add ./components dir to the list
    dirs.push({
      path: join(__dirname, '../src/components')
    })
  })
```
Wanted solution
```javascript
// my-library/setup/index.js
import defu from 'defu'
import libraryLocales from 'path-to-library-locales.json'

export default function () {
  const { nuxt } = this

  nuxt.hook('i18n:locales-ready', (locales) => {
    Object.assign(locales, defu(locales, libraryLocales))
  })
}
```
```javascript
// my-nuxt-app/nuxt.config.js
export default {
  ...,
  buildModules: [
    '@my-org/my-library/setup'
  ]
}
```

I can just rely on the app to include the good translations keys but it's risky and not futur proof. Otherwise I can keep copy paste the main translation in the nuxt app.

existe-deja:
The purpose of this feature is to open to other modules the ability to add extra translated content and merge it to the main i18n messages.
Closes https://github.com/nuxt-community/i18n-module/issues/1308


test/module.test.js
```diff
@@ -2479,3 +2479,158 @@ describe('Store', () => {
     expect(dom.querySelector('#store-path-fr')?.textContent).toBe('/fr/a-propos')
   })
 })
+
+describe.only('Extend Locale with additionalMessages', () => {
```
rchl: Remove `.only`

types/index.d.ts
```diff
@@ -48,6 +48,7 @@ export interface BaseOptions {
   defaultLocaleRouteNameSuffix?: string
   differentDomains?: boolean
   locales?: Locale[] | LocaleObject[]
+  additionalMessages?: Array<LocaleMessages>
```
rchl: Do we really want to allow this to be provided through module options? I thought only through the hook.

On the other hand, I suppose it shouldn't be an issue but I think I'm not a big fan since it would be an third way to provide messages through module config and it can make things more confusing.

(Also, I prefer the `foo[]` syntax for arrays)

test/fixture/extend-locales/pages/index.vue
```diff
@@ -0,0 +1,7 @@
+<template>
+  <div>
+    <div id="current-page">page: {{ $t('home') }}</div>
+    <div id="additional-message">{{ $t('my-module.hello') }}</div>
+    <div id="debug">{{ $i18n.messages }}</div>
```
rchl: Feel free to remove this once you don't need it.

test/fixture/extend-locales/modules/externalModule/index.js
```diff
@@ -0,0 +1,9 @@
+import lang from './lang'
```
rchl: Maybe avoid extra files and just include locales in this file. I don't think extra files add anything to testing.

src/templates/plugin.utils.js
```diff
@@ -195,6 +196,26 @@ export function validateRouteParams (routeParams, localeCodes) {
...
+export function mergeAdditionalMessages (context, { additionalMessages, localeCodes }, onlyLocales = null) {
+  const locales = onlyLocales ? (Array.isArray(onlyLocales) ? onlyLocales : [onlyLocales]) : localeCodes
+  const {
+    i18n
+  } = context.app
+  additionalMessages.forEach(/** @type {Array<import('vue-i18n').LocaleMessages>} */ additionalEntry => {
+    locales.forEach(locale => {
+      i18n.mergeLocaleMessage(locale, additionalEntry[locale])
+    })
+  })
```
rchl: The type annotation is redundant here.

Also I prefer `for of` loop since that avoids extra function calls and looks cleaner.

test/fixture/extend-locales/modules/externalModule/index.js
```diff
@@ -0,0 +1,9 @@
...
+
+export default function () {
```
rchl, line 2:
```suggestion
/** @type {import('@nuxt/types').Module} */
export default function () {
```

src/templates/plugin.main.js
```diff
@@ -114,6 +119,7 @@ export default async (context) => {
             }
           } else if (newLocale !== i18nFallbackLocale) {
             localesToLoadPromises.push(loadLanguageAsync(context, i18nFallbackLocale))
+            mergeAdditionalMessages(context, options, i18nFallbackLocale)
```
rchl: This shouldn't be needed since `loadLanguageAsync` handles merging right?

src/templates/plugin.utils.js
```diff
@@ -195,6 +196,26 @@ export function validateRouteParams (routeParams, localeCodes) {
...
 
+/**
+ * Merge external additional messages
+ *
+ * @param {import('@nuxt/types').Context} context
+ * @param {Pick<ResolvedOptions, 'additionalMessages' | 'localeCodes'>} options
+ * @param {Array<string> | string | null} onlyLocales?
+ * @return {void}
+ */
+export function mergeAdditionalMessages (context, { additionalMessages, localeCodes }, onlyLocales = null) {
```
rchl: I would suggest just passing what you need.
- just `i18n` instead of `context`
- just the needed options instead of all options

rchl (changes requested):
This should also be documented.

Maybe through an additional page in guides named `Extending locales hook` or something.

And if we really want to extend default options with `additionalMessages` (which I'm not sure about) then it should also be added to https://i18n.nuxtjs.org/options-reference.

src/templates/plugin.main.js
```diff
@@ -326,6 +334,7 @@ export default async (context) => {
     i18n.localeProperties = Vue.observable(klona(options.normalizedLocales.find(l => l.code === i18n.locale) || { code: i18n.locale }))
     i18n.defaultLocale = options.defaultLocale
     i18n.differentDomains = options.differentDomains
+    i18n.additionalMessages = options.additionalMessages
```
rchl: I don't think that we should be exposing this on the Vue instance.
Should also update the types - move from `BaseOptions` to `Options`.

existe-deja: fixed by removing it.

src/templates/plugin.main.js
existe-deja: fixed

test/fixture/extend-locales/modules/externalModule/index.js
existe-deja: fixed

test/fixture/extend-locales/pages/index.vue
existe-deja: fixed

src/templates/plugin.utils.js
existe-deja: fixed

test/fixture/extend-locales/modules/externalModule/index.js
existe-deja: fixed

types/index.d.ts
existe-deja: fixed by removing it

test/module.test.js
existe-deja: fixed

existe-deja:
I added docs and an extra test to ensure translations are always overridable.

src/templates/plugin.utils.js
existe-deja: I followed the way loadLanguageAsync passes arguments. Do you want me to avoid destructuring?

test/fixture/extend-locales/lang/fr.js
```diff
@@ -0,0 +1,3 @@
+export default {
```
rchl, line 1: This file can be removed now.

test/fixture/extend-locales/lang/en.js
```diff
@@ -0,0 +1,3 @@
+export default {
```
rchl, line 1: This file can be removed now.

test/fixture/extend-locales/nuxt.config.js
```diff
@@ -0,0 +1,11 @@
...
+import BaseConfig from '../base.config'
+
+/** @type {import('@nuxt/types').NuxtConfig} */
+const config = {
+  ...BaseConfig,
+  buildDir: resolve(__dirname, '.nuxt'),
+  srcDir: __dirname
+}
+
+module.exports = config
```
rchl, line 11:
```suggestion
export default config
```

types/internal.d.ts
```diff
@@ -11,7 +11,8 @@ export type onNavigateInternal = (route: Route) => Promise<[number | null, strin
 export interface ResolvedOptions extends Omit<Required<Options>, 'detectBrowserLanguage' | 'vueI18n' | 'vuex'> {
   detectBrowserLanguage: Required<DetectBrowserLanguageOptions> | false
   localeCodes: readonly Locale[]
-  normalizedLocales: readonly LocaleObject[]
+  normalizedLocales: readonly LocaleObject[],
```
rchl:
```suggestion
  normalizedLocales: readonly LocaleObject[]
```

types/internal.d.ts
```diff
@@ -11,7 +11,8 @@ export type onNavigateInternal = (route: Route) => Promise<[number | null, strin
...
+  additionalMessages?: Array<LocaleMessages>
```
rchl: I think it would make sense to have this property default to `[]` since it would simplify types and null checking

Also, for consistency, please use the `[]` syntax.

```suggestion
  additionalMessages: LocaleMessages[]
```

src/core/hooks.js
```diff
@@ -51,6 +51,17 @@ export function buildHook (options) {
...
+  /** @type {Array<import('vue-i18n').LocaleMessages>} */
+  const additionalMessages = []
+  await this.nuxt.callHook('i18n:extend-locales', additionalMessages)
+  if (Array.isArray(additionalMessages) && additionalMessages.length) {
+    if (!Array.isArray(options.additionalMessages)) {
+      options.additionalMessages = additionalMessages
+    } else {
+      options.additionalMessages.push(...additionalMessages)
+    }
+  }
```
rchl: Can this just be:

```suggestion
  options.additionalMessages = []
  await this.nuxt.callHook('i18n:extend-locales', options.additionalMessages)
```

?

Can `buildHook` run more than once? Only then it would make some sense but then that would likely result in duplicate translations so we would want to avoid that anyway.

src/templates/plugin.utils.js
```diff
@@ -195,6 +196,29 @@ export function validateRouteParams (routeParams, localeCodes) {
   }
 }
 
+/**
+ * Merge external additional messages
+ *
+ * @param {import('@nuxt/types').Context} app
+ * @param {Pick<ResolvedOptions, 'additionalMessages' | 'localeCodes'>} options
+ * @param {Array<string> | string | null} onlyLocales?
```
rchl: Would prefer more specific parameters. For example:

```suggestion
 * @param {import('../../types').NuxtI18nInstance} i18n
 * @param {ResolvedOptions['additionalMessages']>} additionalMessages
 * @param {ResolvedOptions['localeCodes']>}  localeCodes
 * @param {string[] | null} [onlyLocales=null]
```

Note that I've changed `onlyLocales` to `string[]` since you could at call site ensure that it's an array.

src/templates/plugin.utils.js
```diff
@@ -195,6 +196,29 @@ export function validateRouteParams (routeParams, localeCodes) {
...
+ * @return {void}
+ */
+export function mergeAdditionalMessages ({ app }, { additionalMessages, localeCodes }, onlyLocales = null) {
+  if (!additionalMessages) {
+    return
+  }
```
rchl: This won't be needed if we ensure that it's always an array.

src/templates/plugin.utils.js
```diff
@@ -195,6 +196,29 @@ export function validateRouteParams (routeParams, localeCodes) {
...
+
+  const locales = onlyLocales ? (Array.isArray(onlyLocales) ? onlyLocales : [onlyLocales]) : localeCodes
```
rchl: The caller can ensure that an array is passed to make this simper.

rchl (changes requested):

docs/content/en/extend-locales.md
```diff
@@ -0,0 +1,41 @@
...
+category: Guide
+---
+
+If you're a **module author** you can merge extra translations into the project where your module is installed by using the `i18n:extend-locales` hook.
+
+To do this, in your module's setup file listen to the nuxt hook and push your translations, `@nuxtjs/i18n` will do the rest.
+
+This is particulary useful if your module use translated content and you want to offer to users nice default translations.
+
+Exemple:
```
rchl: (I'm splitting this suggestion into multiple comments since otherwise it breaks due to triple backticks)

The terminology is a bit random since we sometimes say messages and sometimes translations. Here are some suggestions.

```suggestion
---
title: Extending messages hook
description: "Nuxt hook to extend app's messages"
position: 14
category: Guide
---

If you're a **module author** and want that module to provide extra messages for your project, you can merge them into the normally loaded messages by using the `i18n:extend-messages` hook.

To do this, in your module's setup file listen to the Nuxt hook and push your messages. `@nuxtjs/i18n` will do the rest.

This is particularly useful if your module use translated content and you want to offer to users nice default translations.

Example:
```

docs/content/en/extend-locales.md
```diff
@@ -0,0 +1,41 @@
...
+```js{}[my-module-exemple/setup.js]
+export default function () {
+  const { nuxt } = this
+
+  nuxt.hook('i18n:extend-locales', function (additionalMessages) {
```
rchl: I've realized that we are not extending locales here but messages.

```suggestion
  nuxt.hook('i18n:extend-messages', function (additionalMessages) {
```

docs/content/en/extend-locales.md
```diff
@@ -0,0 +1,41 @@
...
+  })
+}
+```
+Now the project has access to new messages and can use them through `$t('my-module-exemple.hello')`.
+
+<alert>
+Because module's translation are merged with the project's ones, it's safer to prefix them.
+
+**Main project translations will always override** the module's ones.
+</alert>
```
rchl: The terminology is a bit random since we sometimes say messages and sometimes translations. Here are some suggestions.

```suggestion
Now the project has access to new messages and can use them through `$t('my-module-exemple.hello')`.

<alert>
Because module's messages are merged with the project's ones, it's safer to prefix them.

Main project messages **will always override** the module's ones.
</alert>
```

rchl (changes requested):

test/fixture/extend-locales/lang/fr.js
existe-deja: It's still required for the test with langDir option

test/fixture/extend-locales/lang/en.js
existe-deja: It's still required for the test with langDir option

test/fixture/extend-locales/nuxt.config.js
existe-deja: not sure on this one

types/internal.d.ts
existe-deja: fixed

src/core/hooks.js
existe-deja: If you override additionalMessages like this, you loose de data you may have entered in nuxt.config.js. I fixed my code with the main idea of mutation options instead of keeping it "pure".

src/templates/plugin.utils.js
existe-deja: fixed

src/templates/plugin.utils.js
existe-deja: fixed

docs/content/en/extend-locales.md
existe-deja: fixed

docs/content/en/extend-locales.md
existe-deja: fixxed

docs/content/en/extend-locales.md
existe-deja: fixed

src/core/hooks.js
rchl: You loose which data? I thought we agreed that `additionalMessages` is not supposed to be settable from nuxt.config.

test/fixture/extend-locales/nuxt.config.js
rchl: Why not? Mixing CommonJS and ES imports syntax is not ideal and Nuxt team recommends ES imports syntax.

Unless this has problems with testing framework which I guess it might if other fixtures do the same.

src/core/hooks.js
existe-deja: fixed with an extra test as additionalMessages could have been set by another module

rchl: How so?

The code looks like this right now:

```js
  if (!options.additionalMessages) {
    options.additionalMessages = []
  }
  await this.nuxt.callHook('i18n:extend-messages', options.additionalMessages)
```

and I really can't see how `options.additionalMessages` can be anything else than undefined at the point of the `if` check.

existe-deja: If module A use i18n:extend-messages hook and module B either, at some point additionalMessages have something

rchl: I'm talking about

```js
  if (!options.additionalMessages) {
    options.additionalMessages = []
  }
```

This code should execute only once within the i18n module itself.

existe-deja: You are totally right, I mixed test names so it was called twice. Fixed now.

rchl:
Thanks!

Issue #297, opened 2019-05-09
Implicit dependency on `vue-template-compiler`


pi0:
### Version

[v5.3.0](https://github.com/nuxt-i18n/releases/tag/v5.3.0)

### Reproduction link

[https://github.com/nuxt/now-builder/issues/42](https://github.com/nuxt/now-builder/issues/42)

### Steps to reproduce

-

### What is expected ?

-

### What is actually happening?

-

### Additional comments?

This module has an implicit dependency on `vue-template-compiler` [here](https://github.com/nuxt-community/nuxt-i18n/blob/master/src/helpers/components.js#L6). All direct dependencies like this should be specified in `dependencies` section of package.json.  When using `nuxt-start` (a serverless) build of nuxt, this dependency is not available.

It is also better to only require build related deps on `build:before` nuxt hook to prevent additional overhead when running in production mode.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c237">#c237</a>)</em></sub></div>

rchl:
This code is running when generating extra locale routes.

Nuxt generates routes in this order:
 - generate default routes
 - call `build:extendRoutes` hook
 - process routes returned from `options.router.extendRoutes` function (if defined)

This module uses `ModuleContainer.extendRoutes(fn)` to hook into processing of `options.router.extendRoutes`.

So even when no actually using a hook (because `build:extendRoutes` would happen at wrong time for this module anyway), I think it's OK to keep using `ModuleContainer.extendRoutes(fn)` as then the function shouldn't be called on start of built app, right?

rchl:
@pi0 you think #298 is a good solution to both problems?

pi0:
@rchl yes, that totally makes sense. Thanks for the PR.

rchl:
@pi0 @paulgv I have just noticed a problem with this change.

With nuxt@2.4.5 version at least, I see this error when trying to `build`:

```
Vue packages version mismatch:

- vue@2.6.7 (/Users/rchlodnicki/workspace/tidal-com/node_modules/vue/dist/vue.runtime.common.js)
- vue-template-compiler@2.6.10 (/Users/rchlodnicki/workspace/tidal-com/node_modules/nuxt-i18n/node_modules/vue-template-compiler/package.json)

This may cause things to work incorrectly. Make sure to use the same version for both.
[...]

PR #1327, opened 2021-10-23
fix: load vue-template-compiler with nuxt.resolver.requireModule

kazupon:
#297 has occurred in my environment too
https://github.com/kazupon/nuxt-i18n-next

There is [comment](https://github.com/nuxt-community/i18n-module/issues/297#issuecomment-491755323) at #297, 
I've fixed  with using `this.nuxt.resolver.requireModule`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,7 +135,6 @@
     "playwright-chromium": "1.15.2",
     "release-it": "14.11.6",
     "serve-static": "1.14.1",
-    "typescript": "^4.4.4",
-    "vue-template-compiler": "^2.6.14"
+    "typescript": "^4.4.4"
   }
 }
--- a/package.json
+++ b/package.json
@@ -135,6 +135,7 @@
     "playwright-chromium": "1.15.2",
     "release-it": "14.11.6",
     "serve-static": "1.14.1",
-    "typescript": "^4.4.4"
+    "typescript": "^4.4.4",
+    "vue-template-compiler": "^2.6.14"
   }
 }
--- a/src/core/hooks.js
+++ b/src/core/hooks.js
@@ -1,5 +1,6 @@
 import { STRATEGIES } from '../helpers/constants'
 import { formatMessage } from '../templates/utils-common'
+import { makeRoutes } from '../helpers/routes'
 
 /**
  * @this {import('@nuxt/types/config/module').ModuleThis}
@@ -22,7 +23,6 @@ export function createExtendRoutesHook (options) {
     // Load 'vue-template-compiler' module with Resolver dynamically
     // See https://github.com/nuxt-community/i18n-module/issues/297#issuecomment-491755323
     const { parseComponent } = this.nuxt.resolver.requireModule('vue-template-compiler')
-    const { makeRoutes } = require('../helpers/routes')
 
     const localizedRoutes = makeRoutes(routes, {
       ...options,
--- a/src/core/hooks.js
+++ b/src/core/hooks.js
@@ -19,16 +19,17 @@ export function createExtendRoutesHook (options) {
   const { trailingSlash } = nuxtOptions.router
 
   return routes => {
-    // This import (or more specifically 'vue-template-compiler' in helpers/components.js) needs to
-    // be required only at build time to avoid problems when 'vue-template-compiler' dependency is
-    // not available (at runtime, when using nuxt-start).
+    // Load 'vue-template-compiler' module with Resolver dynamically
+    // See https://github.com/nuxt-community/i18n-module/issues/297#issuecomment-491755323
+    const { parseComponent } = this.nuxt.resolver.requireModule('vue-template-compiler')
     const { makeRoutes } = require('../helpers/routes')
 
     const localizedRoutes = makeRoutes(routes, {
       ...options,
       pagesDir,
       includeUprefixedFallback,
-      trailingSlash
+      trailingSlash,
+      parseComponent
     })
     routes.splice(0, routes.length)
     routes.unshift(...localizedRoutes)
--- a/src/helpers/components.js
+++ b/src/helpers/components.js
@@ -1,9 +1,6 @@
 import { readFileSync } from 'fs'
 import { parse } from '@babel/parser'
 import traverse from '@babel/traverse'
-// Must not be an explicit dependency to avoid version mismatch issue.
-// See https://github.com/nuxt-community/i18n-module/issues/297
-import { parseComponent } from 'vue-template-compiler'
 import { formatMessage } from '../templates/utils-common'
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -135,6 +135,7 @@
     "playwright-chromium": "1.15.2",
     "release-it": "14.11.6",
     "serve-static": "1.14.1",
-    "typescript": "^4.4.4"
+    "typescript": "^4.4.4",
+    "vue-template-compiler": "^2.6.14"
```
rchl: Is it necessary to add this dependency? Since `nuxt` is already in dev-dependencies, this seems redundant and potentially problematic due to version conflicts?

src/core/hooks.js
```diff
@@ -19,16 +19,17 @@ export function createExtendRoutesHook (options) {
   const { trailingSlash } = nuxtOptions.router
 
   return routes => {
-    // This import (or more specifically 'vue-template-compiler' in helpers/components.js) needs to
-    // be required only at build time to avoid problems when 'vue-template-compiler' dependency is
-    // not available (at runtime, when using nuxt-start).
+    // Load 'vue-template-compiler' module with Resolver dynamically
+    // See https://github.com/nuxt-community/i18n-module/issues/297#issuecomment-491755323
+    const { parseComponent } = this.nuxt.resolver.requireModule('vue-template-compiler')
     const { makeRoutes } = require('../helpers/routes')
```
rchl: Since this import doesn't import from `vue-template-compiler` anymore, I think we could make it a normal top-level import now since mixing ES imports and CommonJS is not ideal.

rchl (changes requested):

kazupon:
I have just fixed the part pointed out in your review comments
https://github.com/nuxt-community/i18n-module/pull/1327/commits/5f63ca6d3754f84dee3a1589ad0690344d5b07c3

src/core/hooks.js
```diff
@@ -19,16 +20,16 @@ export function createExtendRoutesHook (options) {
   const { trailingSlash } = nuxtOptions.router
 
   return routes => {
-    // This import (or more specifically 'vue-template-compiler' in helpers/components.js) needs to
-    // be required only at build time to avoid problems when 'vue-template-compiler' dependency is
-    // not available (at runtime, when using nuxt-start).
-    const { makeRoutes } = require('../helpers/routes')
+    // Load 'vue-template-compiler' module with Resolver dynamically
+    // See https://github.com/nuxt-community/i18n-module/issues/297#issuecomment-491755323
```
rchl, line 24: I can't really think of why this change is needed in the first place. The linked comment also doesn't seem to help in understanding because the conditions that caused the original issue are no longer present (this code only runs at build time).

Care to explain why is this code problematic  in your case?

kazupon:
I'm very sorry for the lack of explanation.

Because `Cannot find module 'vue-template-compiler’ error occurs in cases other than build time, like the below issue.
https://github.com/nuxt/vercel-builder/issues/42

I am trying the verify & PoC of Nuxt Bridge / Nuxt3 support using `@nuxtjs/i18n` now in the repo here:
https://github.com/kazupon/nuxt-i18n-next

`@nuxt/kit` has an API to install existing modules made with Nuxt2.
Here is the code that uses the API:

https://github.com/kazupon/nuxt-i18n-next/blob/3cc2b3673b413b42eda4de477177755b0e9f696f/packages/nuxt-i18n/src/module.ts#L31

When I try to run the new module, which uses this existing `@nuxtjs/i18n` and is made with `nuxt/kit`, in the nuxt3 project, I get the following error:

```
 ERROR  Cannot find module 'vue-template-compiler'                                                                                                                         01:37:16
Require stack:
- /path/to/nuxt-i18n-next/node_modules/@nuxtjs/i18n/src/helpers/components.js

  Require stack:
  - /path/to/nuxt-i18n-next/node_modules/@nuxtjs/i18n/src/helpers/components.js
  at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
  at Function.resolve (node:internal/modules/cjs/helpers:108:19)
  at p (/path/to/nuxt-i18n-next/node_modules/jiti/dist/jiti.js:1:52925)
  at g (/path/to/nuxt-i18n-next/node_modules/jiti/dist/jiti.js:1:54200)
  at /path/to/nuxt-i18n-next/node_modules/@nuxtjs/i18n/src/helpers/components.js:6:28
  at g (/path/to/nuxt-i18n-next/node_modules/jiti/dist/jiti.js:1:55111)
  at /path/to/nuxt-i18n-next/node_modules/@nuxtjs/i18n/src/helpers/routes.js:2:19
  at g (/path/to/nuxt-i18n-next/node_modules/jiti/dist/jiti.js:1:55111)
  at Object.<anonymous> (/path/to/nuxt-i18n-next/node_modules/@nuxtjs/i18n/src/core/hooks.js:25:28)
  at Object.extendRoutes (/path/to/nuxt-i18n-next/node_modules/@nuxt/kit/dist/index.mjs:814:17)
  at Builder.resolveRoutes (node_modules/@nuxt/builder/dist/builder.js:546:56)
  at async Promise.all (index 1)
  at async Builder.generateRoutesAndFiles (node_modules/@nuxt/builder/dist/builder.js:375:5)
  at async Builder.build (node_modules/@nuxt/builder/dist/builder.js:319:5)
  at async Object._buildDev (node_modules/@nuxt/cli/dist/cli-dev.js:107:5)
  at async Object.startDev (node_modules/@nuxt/cli/dist/cli-dev.js:65:7)
```

From this stack trace, if you install `nuxt/kit` with API `installModule`, it will execute the build process of `@nuxtjs/i18n`.

This is outside the topic of this issue.

To be honest, I'm not sure if it's best practice to use `installModule` in `nuxt/kit` to load an existing Nuxt module.
In fact, from reading [this code](https://github.com/nuxt/framework/blob/a896c64b1232dba1d01874159a9953d5f79bccc4/packages/kit/src/module/container.ts) in `nuxt/kit`, it seems to support compatibility with existing Nuxt modules.

In fact, installModule is used [here](https://github.com/nuxt/framework/blob/a896c64b1232dba1d01874159a9953d5f79bccc4/packages/kit/src/module/container.ts#L110)

I think @pi0  would know if I'm right about it.

In any case, from [previous issues](https://github.com/nuxt/vercel-builder/issues/42), I think the same problem may occur in the future.
So, I think this issue should be fixed.

pi0:
Hi! I have no knowledge of `nuxt-i18n-next` usage but seems a valid fix as this module has no explicit dependency on `vue-template-compoiler`, `resolveModule` uses a better search path that includes `@nuxt` packages that actually require it.  

Using installModule` should be fine btw :)

rchl:
Thanks

PR #1377, opened 2022-01-12
refactor(style): corrected a typo in prefix var name

gekkedev:
also added a comment for readability

### DIFF
--- a/src/helpers/routes.js
+++ b/src/helpers/routes.js
@@ -6,7 +6,7 @@ import { adjustRouteDefinitionForTrailingSlash, getPageOptions } from './utils'
  * @typedef {import('@nuxt/types/config/router').NuxtRouteConfig} NuxtRouteConfig
  * @typedef {import('../../types/internal').ResolvedOptions & {
  *   pagesDir: string
- *   includeUprefixedFallback: boolean
+ *   includeUnprefixedFallback: boolean
  *   trailingSlash: import('@nuxt/types/config/router').NuxtOptionsRouter['trailingSlash']
  *   parseComponent: import('vue-template-compiler')['parseComponent']
  * }} MakeRouteOptions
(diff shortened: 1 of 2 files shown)


src/helpers/routes.js
```diff
@@ -155,7 +155,8 @@ export function makeRoutes (baseRoutes, {
         path = adjustRouteDefinitionForTrailingSlash(path, trailingSlash, isChildWithRelativePath)
       }
 
-      if (shouldAddPrefix && isDefaultLocale && strategy === STRATEGIES.PREFIX && includeUprefixedFallback) {
+      //add a route without prefix as fallback
```
rchl: Apart from introducing a linting issue, is this even needed? The variable names are self-descriptive so this feels redundant.

gekkedev: Not aware of your linting; I edited it in the browser but assume this is what you prefer?
```suggestion
      // add a route without prefix as fallback
```
Regarding obsoleteness of the comment, I just thought it's helpful to write down what will happen in the following lines (yes, plural) instead of expecting people to evaluate it in their head just to find out that this is not the closure they were looking for. There are many places in this file where a route gets pushed so I thought why not help with differentiation. :slightly_smiling_face:

rchl: When one of the variables checked is named "includeUnprefixedFallback" then I would argue that a comment that says essentially the same, only using a plain sentence, is redundant since the code is self-documenting already. The `includeUnprefixedFallback` check could be done first, if you think that that will help readability.

But I won't make it an issue if you want to keep it in. Just fix the linter issue.

gekkedev: Either way should be fine, you can choose one
```suggestion
      // add a route without prefix as fallback
```
```suggestion
```
not sure if rearranging conditionals will do much since they still need to be seen in conjunction

rchl: Since you are asking then just remove it.

PR #1392, opened 2022-02-09
fix(docs): correct links to [Custom paths] and [Ignore routes]

hydRAnger:
It seems these links are not correct (with anchor, but they are not on this page.)


rchl:
Thanks

PR #1396, opened 2022-02-15
chore(ci): add github action for creating releases

rchl:
@kazupon I'm setting up an action for releasing to make the process easy and transparent. I'm not sure this will work due to it being a monorepo but it should work eventually, even if it will require some tweaking.

One thing of note is that this requires a proper use of conventional commit convention for the changelog and versions to be correct so this would need to be followed like it already is on the `main` branch.

Of course we can still allow non-convential commits, especially at this stage of development but then those won't be included in the changelog.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,37 @@
+name: Release New Version
+on:
+  push:
+    branches:
+      - next
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: google-github-actions/release-please-action@v3
+        id: release
+        with:
+          release-type: node
+          package-name: nuxt-neo
+          monorepo-tags: true
+      - uses: actions/checkout@v2
+        if: ${{ steps.release.outputs.releases_created }}
+      - uses: actions/setup-node@v2
+        if: ${{ steps.release.outputs.releases_created }}
+        with:
+          node-version: 14
+          registry-url: https://registry.npmjs.org
+          cache: yarn
+      - name: Lint and Test
+        if: ${{ steps.release.outputs.releases_created }}
+        run: yarn --immutable && yarn lint && yarn test
+      - name: Publish to NPM
+        if: ${{ steps.release.outputs.releases_created }}
+        run: |
+          cd packages/nuxt-i18n
+          npm publish --tag edge
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      # - name: Deploys Github Pages Docs
+      #   if: ${{ steps.release.outputs.releases_created }}
+      #   run: yarn deploy:docs


rchl:
Actually maybe I've jumped to conclusions to fast.
Do you want to instead release every single commit as an `edge` tag?
Or do you want to decide when to release and bump the version for each release?

kazupon:
@rchl 

> One thing of note is that this requires a proper use of conventional commit convention for the changelog and versions to be correct so this would need to be followed like it already is on the `main` branch.
> 
> Of course we can still allow non-convential commits, especially at this stage of development but then those won't be included in the changelog.

I haven't committed according to the conventional commit 😅 , but I believe I will need to move to that commit convention in near the future when we will release the alpha version 😃 



> Actually maybe I've jumped to conclusions to fast.
> Do you want to instead release every single commit as an `edge` tag?

Yeah, I hope it
In the nuxt/framework repo, it's already based on the commit, and versioning is also released using the commit hash.
https://github.com/nuxt/framework/tree/main/scripts

The advantage of versioning releases with a commit hash is that you can roll back to a version that works correctly. The disadvantage is that all edge versions are recorded in npm DB for each commit, which means that versions other than the edge version will be buried because they are all listed in the version on the npm site.

e.g.:
https://www.npmjs.com/package/nuxt-edge

rchl:
And for Nuxt the downsides are not that big because they use a separate package name for nightly releases. I guess we could do the same? It's still possible to alias that to an original name with some trickery in package.json.

kazupon:
I think so too :)

If you would be possible to setup for edge releasing, I would like to do as a favor 🙏

rchl:
So what's the conclusion? Create `@nuxtjs/i18n-edge`?

kazupon:
>So what's the conclusion? Create @nuxtjs/i18n-edge?

I think it should be named, but I think it might be named by `@nuxt/i18n-edge`. because the scope name of nuxt official packages move to `@nuxt` from `@nuxtjs` recently.

I’m not sure that `nuxt/i18n` is nuxt official package, so I recognize it’s maintained nuxt module by nuxt community :) /cc @pi0 @Atinux

atinux:
I think for today until we work all together with Pooya on an official module, `@nuxtjs/i18n-edge` is best name.

kazupon:
Thanks!
Okay, so We'll use the package name 🙂

kazupon:
@rchl 
Could you release `@nuxtjs/i18n-edge`, please?
I heard that someone already want to use it :)

rchl:
Sorry, was sick lately.

We need to do the rename first (https://github.com/nuxt-community/i18n-module/pull/1402) and then adapt the script from nuxt framework I guess

dveldhoen:
Yes. Please release, we are happy to test and provide feedback!

rchl:
It was released. See https://www.npmjs.com/package/@nuxtjs/i18n-edge

You can install it locally by adding `"@nuxtjs/i18n": "npm:@nuxtjs/i18n-edge"` into package json dependencies.

dveldhoen:
> It was released. See https://www.npmjs.com/package/@nuxtjs/i18n-edge
> 
> You can install it locally by adding `"@nuxtjs/i18n": "npm:@nuxtjs/i18n-edge"` into package json dependencies.

Awesome! Will definitely try it out. Thanks!

safejace:
@dveldhoen have you been able to get this working at all?

@rchl is there anyway to know what aspects of the module should* be working with nuxt3 in relation to the docs? (ie if i follow the nuxt2 docs, how confused will i be lol :) )

rchl:
I believe this is more or less up-to-date list https://github.com/nuxt-community/i18n-module/blob/next/TODO.md

TomMiller-mas:
New to nuxt, do you have a basic config for nuxt.config.ts? No need to setup a config?
Install went fine. Recompiled the project with no problem after installation.

Is this documentation still accurate for this module? 
https://github.com/intlify/nuxt3#-what-is-defference-from-nuxtjsi18n-

PR #1402, opened 2022-02-18
refactor: rename back to `@nuxtjs/i18n`

rchl:


rchl:
Maybe should consider renaming `packages/nuxt-i18n` to `packages/nuxtjs-i18n` for consistency. Or if we don't need consistency here then we could even rename to `packages/i18n`.

kazupon (approved):

kazupon:
Thank you!

PR #1403, opened 2022-02-18
chore: add scripts for releasing -edge

rchl:
NOTE: This doesn't work currently.

The main issue with getting it to work is the fact that there is:
```
  "resolutions": {
    "@nuxtjs/i18n": "workspace:./packages/nuxt-i18n"
  },
```
in `package-json`.

For the script to work, I think this would need to be also renamed to `@nuxtjs/i18n-edge` and possibility other references would also have to be updated...

It might be a bit of work and I don't have time right now to work on it so if you have time, you could try to make it work @kazupon .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,3 +35,31 @@ jobs:
         run: yarn workspace @nuxtjs/i18n build
       - name: Testing
         run: yarn test
+  edge-release:
+    needs:
+      - build
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        node: [14]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup node
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node }}
+          cache: 'yarn'
+      - name: yarn install
+        run: yarn --immutable
+      - name: Building
+        run: yarn workspace @nuxtjs/i18n build
+      - name: Release Edge
+        if: |
+          github.event_name == 'push' &&
+          !contains(github.event.head_commit.message, '[skip-release]') &&
+          !contains(github.event.head_commit.message, 'chore') &&
+          !contains(github.event.head_commit.message, 'docs')
+        run: ./scripts/release-edge.sh
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NODE_AUTH_TOKEN}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - next
-      - chore/release-edge
   pull_request:
     branches:
       - next
@@ -39,17 +38,13 @@ jobs:
   edge-release:
     needs:
       - build
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        node: [14]
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - name: Setup node
         uses: actions/setup-node@v2
         with:
-          node-version: ${{ matrix.node }}
+          node-version: 14
           cache: 'yarn'
       - name: yarn install
         run: yarn --immutable
@@ -63,4 +58,4 @@ jobs:
           !contains(github.event.head_commit.message, 'docs')
         run: ./scripts/release-edge.sh
         env:
-          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+          NPM_TOKEN: ${{secrets.NPM_TOKEN}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,4 +63,4 @@ jobs:
           !contains(github.event.head_commit.message, 'docs')
         run: ./scripts/release-edge.sh
         env:
-          NODE_AUTH_TOKEN: ${{secrets.NODE_AUTH_TOKEN}}
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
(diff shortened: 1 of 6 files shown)


rchl:
I have published it manually at https://www.npmjs.com/package/@nuxtjs/i18n-edge but we should still fix this script and release through CI. I think the only change really needed is to rename the `resolutions`.

kazupon:
@rchl 
I've tweaked the release script and GitHub Actions CI workflow.
So we publish edge release on CI, we need the npm token, and need to set the `NODE_AUTH_TOKEN` env.

package.json
```diff
@@ -44,7 +44,7 @@
   },
   "packageManager": "yarn@3.1.1",
   "resolutions": {
-    "@nuxtjs/i18n": "workspace:./packages/nuxt-i18n"
+    "@nuxtjs/i18n": "npm:@nuxtjs/i18n-edge@latest"
```
rchl: Is this change really supposed to be committed?
I'm guessing that it would make local testing not use the local version of the package which could be quite inconvenient, if not wrong.

.github/workflows/ci.yml
```diff
@@ -35,3 +35,31 @@ jobs:
...
+        run: yarn workspace @nuxtjs/i18n build
+      - name: Release Edge
+        if: |
+          github.event_name == 'push' &&
+          !contains(github.event.head_commit.message, '[skip-release]') &&
+          !contains(github.event.head_commit.message, 'chore') &&
+          !contains(github.event.head_commit.message, 'docs')
+        run: ./scripts/release-edge.sh
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NODE_AUTH_TOKEN}}
```
rchl: The token already exists as a `NPM_TOKEN` secret.

scripts/release-edge.sh
```diff
@@ -0,0 +1,28 @@
...
+# Bump versions to edge
+yarn jiti ./scripts/bump-edge
+
+# Resolve yarn
+YARN_ENABLE_IMMUTABLE_INSTALLS=false yarn install
+
+# Update token
+if [[ ! -z ${NODE_AUTH_TOKEN} ]] ; then
+  echo "//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}" >> ~/.npmrc
+  echo "registry=https://registry.npmjs.org/" >> ~/.npmrc
```
rchl: Can we store the `.npmrc` file in the root of the repo instead? I wouldn't want this to be saved in my home directory when accidentally (or even intentionally) running this command from the terminal.

.github/workflows/ci.yml
```diff
@@ -35,3 +36,31 @@ jobs:
         run: yarn workspace @nuxtjs/i18n build
```
rchl: Just `yarn build`

.github/workflows/ci.yml
```diff
@@ -35,3 +36,31 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - name: Setup node
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node }}
+          cache: 'yarn'
+      - name: yarn install
+        run: yarn --immutable
+      - name: Building
+        run: yarn workspace @nuxtjs/i18n build
```
rchl: Just `yarn build`

.npmrc
```diff
@@ -0,0 +1 @@
+# This npmrc is used by CI edge releasing
```
rchl: We don't need this file committed, right?

scripts/release-edge.sh
```diff
@@ -0,0 +1,29 @@
...
+  echo "always-auth=true" >> ~/.npmrc
+  echo "npmAuthToken: ${NODE_AUTH_TOKEN}" >> ~/.yarnrc.yml
```
rchl: Why is that needed? It looks like `npm` is used for publishing.

scripts/release-edge.sh
rchl: Planning to address this? I see that you went back and forth.

kazupon:
I've been trying some things to do an edge release, and since GitHub Actions is hard to debug, there are some commits that look strange for you.

The issue I'm running into is that for some reason `packages/nuxt-i18n` is not recognized by workspace and yarn.
https://github.com/nuxt-community/i18n-module/runs/5265893030?check_suite_focus=true

```
YN0001: │ Error: @nuxtjs/i18n@workspace:packages/nuxt-i18n: Workspace not found (@nuxtjs/i18n@workspace:packages/nuxt-i18n)
      at Ke.getWorkspaceByDescriptor (/home/runner/work/i18n-module/i18n-module/.yarn/releases/yarn-3.1.1.cjs:441:3390)
      at zp.getCandidates (/home/runner/work/i18n-module/i18n-module/.yarn/releases/yarn-3.1.1.cjs:407:28[17](https://github.com/nuxt-community/i18n-module/runs/5265893030?check_suite_focus=true#step:6:17))
      at pd.getCandidates (/home/runner/work/i[18](https://github.com/nuxt-community/i18n-module/runs/5265893030?check_suite_focus=true#step:6:18)n-module/i18n-module/.yarn/releases/yarn-3.1.1.cjs:432:5275)
      at pd.getCandidates (/home/runner/work/i18n-module/i18n-module/.yarn/releases/yarn-3.1.1.cjs:432:5275)
      at /home/runner/work/i18n-module/i18n-module/.yarn/releases/yarn-3.1.1.cjs:441:7674
      at du (/home/runner/work/i18n-module/i18n-module/.yarn/releases/yarn-3.1.1.cjs:382:4003)
      at R (/home/runner/work/i18n-module/i18n-module/.yarn/releases/yarn-3.1.1.cjs:441:7654)
```

If you have the time, I would like you to give it a try too, so to solve this issue 🙏

rchl:
That's the issue I've mentioned in the initial comment. It should be enough to rename the key.

rchl:
It works.

kazupon:
good job! thanks!

kazupon (approved):

PR #1429, opened 2022-04-24
fix: fixed locale changing issue in server side middleware

### LINKED ISSUE #1310: Change locale in the middleware does not set locale in the browser
### Version
@nuxtjs/i18n: 7.0.3
nuxt: 2.15.7

### Nuxt configuration

 - [ ] Applies to a site deployed to a static server (site generated with `nuxt generate`)
 - [x] Applies to a site deployed to a server with a Node backend

```js
import { Options } from '@nuxtjs/i18n';

export default {
  lazy: false,
  langDir: 'i18n/locales/',
  defaultLocale: 'en_US',
  strategy: 'no_prefix',
  vuex: {
    moduleName: 'i18n',
    syncLocale: true,
  },
  vueI18n: {
    fallbackLocale: 'en_US',
    fallbackRoot: true,
    silentFallbackWarn: true,
  },
  locales: [
    {
      code: 'en_US',
      iso: 'en-US',
      file: 'en_US.json',
      name: 'English',
    },
    {
      code: 'pt_BR',
      iso: 'pt-BR',
      file: 'pt_BR.json',
      name: 'Português Brasil',
    },
  ],
} as Options;
```

### Steps to reproduce
I'm trying to redirect the user based on a queryString using middleware. I know that the middleware runs in the server but I couldn't find a way to sync with the browser so even if I use `i18n.setLocale('pt_BR')` in the middleware the application sets it back to `en_US` in the browser. Code:

```js
export default async function languageMiddleware({ route, i18n }) {
  const { lang } = route.query;

  if (lang) {
    // the lang is pt_BR for example
    await i18n.setLocale(lang);
  }
}
```

### What is Expected?
To the browser load the page with the set language

Browser is defaulting to the default language.

I found an issue saying to use the `vuex.syncLocale` but apparently it does not work anymore.
Am I using a bad strategy here or it is just a bug?

suleymangezsat:
Close: https://github.com/nuxt-community/i18n-module/issues/1310

When I tried to change the language in middleware running on the server side, I called setLocale, but the language did not change. While investigating the situation, I noticed that the language I added was added to the end of the array of previously added cookie headers. This means that there are more than one redirect cookie in the cookie headers and this prevents language change.

So I made a fix like this:
If there is a redirect cookie in the cookies in the incoming request, then override that cookie, otherwise add the new locale cookie as the value of the 'Set-Cookie' header.

### DIFF
--- a/src/templates/utils-common.js
+++ b/src/templates/utils-common.js
@@ -214,8 +214,21 @@ export function setLocaleCookie (locale, res, { useCookie, cookieDomain, cookieK
     }
 
     const redirectCookie = cookieSerialize(cookieKey, locale, cookieOptions)
-    headers.push(redirectCookie)
+    let foundRedirectCookie = false
+    const cookies = headers.map(header => {
+      const cookie = cookieParse(header)
+      if (cookie[cookieKey]) {
+        cookie[cookieKey] = locale
+        foundRedirectCookie = true
+        return redirectCookie
+      }
+      return header
+    })
 
-    res.setHeader('Set-Cookie', headers)
+    if (!foundRedirectCookie) {
+      cookies.push(redirectCookie)
+    }
+
+    res.setHeader('Set-Cookie', cookies)
   }
 }
--- a/src/templates/utils-common.js
+++ b/src/templates/utils-common.js
@@ -214,21 +214,12 @@ export function setLocaleCookie (locale, res, { useCookie, cookieDomain, cookieK
     }
 
     const redirectCookie = cookieSerialize(cookieKey, locale, cookieOptions)
-    let foundRedirectCookie = false
-    const cookies = headers.map(header => {
+    headers = headers.filter(header => {
       const cookie = cookieParse(header)
-      if (cookie[cookieKey]) {
-        cookie[cookieKey] = locale
-        foundRedirectCookie = true
-        return redirectCookie
-      }
-      return header
+      return !(cookieKey in cookie)
     })
+    headers.push(redirectCookie)
 
-    if (!foundRedirectCookie) {
-      cookies.push(redirectCookie)
-    }
-
-    res.setHeader('Set-Cookie', cookies)
+    res.setHeader('Set-Cookie', headers)
   }
 }


src/templates/utils-common.js
```diff
@@ -214,8 +214,21 @@ export function setLocaleCookie (locale, res, { useCookie, cookieDomain, cookieK
     }
 
     const redirectCookie = cookieSerialize(cookieKey, locale, cookieOptions)
-    headers.push(redirectCookie)
+    let foundRedirectCookie = false
+    const cookies = headers.map(header => {
+      const cookie = cookieParse(header)
+      if (cookie[cookieKey]) {
+        cookie[cookieKey] = locale
```
rchl: There is no need to set the cookie value if it's being dropped anyway.

Though I would prefer to have a filter that just filters out existing cookie and then always adds the new cookie after filtering is done.

rchl (changes requested):

src/templates/utils-common.js
suleymangezsat: I just forgot to remove `cookie[cookieKey] = locale`

The points are addressed.

src/templates/utils-common.js
```diff
@@ -214,6 +214,10 @@ export function setLocaleCookie (locale, res, { useCookie, cookieDomain, cookieK
     }
 
     const redirectCookie = cookieSerialize(cookieKey, locale, cookieOptions)
+    headers = headers.filter(header => {
+      const cookie = cookieParse(header)
+      return !cookie[cookieKey]
```
rchl: I'd rather check whether the cookie name matches rather than checking if the cookie value is truthy:
```suggestion
      return cookieKey in cookie
```

rchl: Also, parsed header can include more than one cookie so in theory this could unintentionally drop some irrelevant cookies.

Though if we assume that only this module sets the locale cookie then I guess it should always be set on its own (unless node does some merging on its own but I doubt)...

suleymangezsat: Node does this merging staff, so I think it's not possible to have the same key in different cookies, don't worry about it.

src/templates/utils-common.js
```diff
@@ -214,6 +214,10 @@ export function setLocaleCookie (locale, res, { useCookie, cookieDomain, cookieK
     }
 
     const redirectCookie = cookieSerialize(cookieKey, locale, cookieOptions)
+    headers = headers.filter(header => {
+      const cookie = cookieParse(header)
+      return cookieKey in cookie
```
rchl: Sorry, my bad. It was supposed to be.
```suggestion
      return !(cookieKey in cookie)
```

rchl (approved):

Issue #1401, opened 2022-02-17
Selectively disable redirects to prefixless for default locale routes with prefix_and_default strategy


eatyrghost:
**Problem**
We are using the `prefix_and_default` strategy with a default locale set to `en-us`.

If we visit a route with the default locale as the prefix, e.g. `/en-us/about-us`, we are redirected to the prefixless route, `/about-us`. This is acceptable for most of our pages.

In a handful of cases, we don't want to redirect - we want to specifically use the `/en-us` prefix.

**Desired Solution**
There might be two ways we could configure this.

1. A new property on the page component:

```
export default {
    disableDefaultRedirect: true
};
```

2. A configuration that accepts a boolean or an array of prefixless paths:

```
// Applies to all routes
i18n: {
    disableDefaultRedirect: true
}

// Applies to /about-us and /products only
i18n: {
    disableDefaultRedirect: [
        '/about-us',
        '/products'
    ]
}
```

jeiksegovia:
That seems very useful

jeiksegovia:
This feature is still relevant....

jeiksegovia:
nice!!

DanielHefti:
I'm looking for this feature as well!

julioglobal100:
we need this! thanks

BobbieGoede:
Closing this as v7 is not being actively worked on (critical hotfixes only), if this still applies to v8 and later please open a new issue!

PR #1437, opened 2022-05-04
feat: add possibility to disable default redirect for prefix_and_default strategy

Renhor:
Regarding to issue #1401 

Added property to config: 
```
i18n: {
  disableDefaultRedirect: boolean | string[] // false by default
}
```
It works only for strategy **prefix_and_default**.
It disables redirect for routes are prefixed with default language, or we can choose custom path pattern for disabling redirect behaviour:
```
i18n: {
  disableDefaultRedirect: [
    '/about',
    '/blog/*'
  ]
}
```

**UPD:**

Added property to config: 
``` 
type PrefixAndDefaultRule = 'default' | 'prefix'

i18n: {
  prefixAndDefaultRule?: PrefixAndDefaultRule
}
```

By setting the value **prefixAndDefaultRule = 'prefix'**  we are no longer redirected from prefixed routes and if there is no prefix  then the other routers will also have no prefix.

For now, I decided not to implement an array with a list of routes for which the default behavior is disabled

### DIFF
--- a/docs/content/en/api.md
+++ b/docs/content/en/api.md
@@ -28,11 +28,17 @@ All [Vue I18n properties and methods](http://kazupon.github.io/vue-i18n/api/#vue
 
   - **Arguments**:
     - locale: (type: `string`)
+    - forcePrefix: (type: `boolean`)
   - **Returns**: `string`
 
-  Returns path of the current route for specified `locale`.
+  Returns path of the current route for specified `locale`. 
+
+  If `forcePrefix` is set to `true`, it returns the path with the prefix. Useful for `prefix_and_default` strategy. 
+  You can also set global behavior through option prefixAndDefaultRules.
 
   See also [Basic usage - nuxt-link](../basic-usage#nuxt-link).
+
+  See more details about [prefixAndDefaultRules](../options-reference#prefixanddefaultrules).
 
   See type definition for [Location](https://github.com/vuejs/vue-router/blob/f40139c27a9736efcbda69ec136cb00d8e00fa97/types/router.d.ts#L125).
 
--- a/docs/content/en/api.md
+++ b/docs/content/en/api.md
@@ -28,17 +28,11 @@ All [Vue I18n properties and methods](http://kazupon.github.io/vue-i18n/api/#vue
 
   - **Arguments**:
     - locale: (type: `string`)
-    - forcePrefix: (type: `boolean`)
   - **Returns**: `string`
 
   Returns path of the current route for specified `locale`. 
 
-  If `forcePrefix` is set to `true`, it returns the path with the prefix. Useful for `prefix_and_default` strategy. 
-  You can also set global behavior through option prefixAndDefaultRules.
-
   See also [Basic usage - nuxt-link](../basic-usage#nuxt-link).
-
-  See more details about [prefixAndDefaultRules](../options-reference#prefixanddefaultrules).
 
   See type definition for [Location](https://github.com/vuejs/vue-router/blob/f40139c27a9736efcbda69ec136cb00d8e00fa97/types/router.d.ts#L125).
 
--- a/docs/content/en/options-reference.md
+++ b/docs/content/en/options-reference.md
@@ -186,17 +186,15 @@ Set this to `true` when using different domains for each locale. If enabled, no 
 
 Whether [custom paths](/routing#custom-paths) are extracted from page files using babel parser.
 
-## `prefixAndDefaultRules`
+## `prefixAndDefaultRule`
 
 - type: `object`
-- default: `{ routing: 'default', switchLocale: 'default' }`
+- default: `'default'`
 
 Modification of the standard behavior of the `prefix_and_default` strategy.
 
-By setting the value `routing = 'prefix'` we are no longer being redirected from prefixed path.
+By setting the value `prefixAndDefaultRule = 'prefix'` we are no longer being redirected from prefixed path.
 If there is no prefix then the other routers will also have no prefix.
-
-By setting the value `switchLocale = 'prefix'` the `switchLocalePath()` method adds a prefix to the current route.
 
 ## `pages`
 
--- a/docs/content/en/options-reference.md
+++ b/docs/content/en/options-reference.md
@@ -186,6 +186,18 @@ Set this to `true` when using different domains for each locale. If enabled, no 
 
 Whether [custom paths](/routing#custom-paths) are extracted from page files using babel parser.
 
+## `prefixAndDefaultRules`
+
+- type: `object`
+- default: `{ routing: 'default', switchLocale: 'default' }`
+
+Modification of the standard behavior of the `prefix_and_default` strategy.
+
+By setting the value `routing = 'prefix'` we are no longer being redirected from prefixed path.
+If there is no prefix then the other routers will also have no prefix.
+
+By setting the value `switchLocale = 'prefix'` the `switchLocalePath()` method adds a prefix to the current route.
+
 ## `pages`
 
 - type: `object`
--- a/src/templates/plugin.main.js
+++ b/src/templates/plugin.main.js
@@ -15,7 +15,7 @@ import {
   loadLanguageAsync,
   resolveBaseUrl,
   registerStore,
-  mergeAdditionalMessages
+  mergeAdditionalMessages, getHelpers
 } from './plugin.utils'
 // @ts-ignore
 import { joinURL } from '~i18n-ufo'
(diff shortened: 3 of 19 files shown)


src/templates/plugin.utils.js
```diff
@@ -216,6 +216,40 @@ export function mergeAdditionalMessages (i18n, additionalMessages, localeCodes,
...
+  if (typeof disableDefaultRedirect === 'boolean') {
+    return false
+  }
+
+  const clearPath = clearPathFromLocale(pathString, localeCodes)
+  const clearPathWithSlash = addSlashToPath(clearPath)
+
+  return !disableDefaultRedirect
+    .map((patternString) => {
+      const isPattern = patternString.includes('*')
```
rchl: Looks like `*` is only supported at the end so you could use `endsWith` here.

src/templates/plugin.main.js
```diff
@@ -153,14 +153,20 @@ export default async (context) => {
     let redirectPath = ''
 
     const isStaticGenerate = process.static && process.server
+    const isSameLocale = getLocaleFromRoute(route) !== newLocale
```
rchl: The variable name is wrong. Should be `isDifferentLocale`.

src/templates/utils-common.js
```diff
@@ -223,3 +223,22 @@ export function setLocaleCookie (locale, res, { useCookie, cookieDomain, cookieK
     res.setHeader('Set-Cookie', headers)
   }
 }
+
+/**
+ * @param  {string} pathString
+ * @param  {readonly string[]} localeCodes
+ * @return {string}
+ */
+export function clearPathFromLocale (pathString, localeCodes) {
```
rchl: Naming feels off. Maybe
```suggestion
export function removeLocaleFromPath (pathString, localeCodes) {
```

src/templates/utils-common.js
```diff
@@ -223,3 +223,22 @@ export function setLocaleCookie (locale, res, { useCookie, cookieDomain, cookieK
...
+
+  return pathString.replace(regexp, '') || '/'
+}
+
+/**
+ * @param  {string} pathString
+ * @return {string}
+ */
+export function addSlashToPath (pathString) {
+  return pathString.endsWith('/') ? pathString : `${pathString}/`
```
rchl: This seems a bit too naive and will likely fail if there are query params in the path (`route.fullPath`). This module uses `ufo` which should have more robust implementation of that (`withTrailingSlash` maybe).

types/index.d.ts
```diff
@@ -52,9 +52,12 @@ export interface BaseOptions {
   onLanguageSwitched?: (oldLocale: string, newLocale: string) => void
 }
 
+export type DisableDefaultRedirect = boolean | string[]
+
 export interface Options extends BaseOptions {
   baseUrl?: string | ((context: NuxtContext) => string)
   detectBrowserLanguage?: DetectBrowserLanguageOptions | false
+  disableDefaultRedirect?: DisableDefaultRedirect;
```
rchl: The naming of that option is a bit too generic for my taste since it applies only in very specific case, only when using specific strategy. We can think more about it after other changes are addressed.

src/templates/plugin.main.js
```diff
@@ -153,14 +153,20 @@ export default async (context) => {
...
     if (
       !isStaticGenerate &&
       !app.i18n.differentDomains &&
       options.strategy !== Constants.STRATEGIES.NO_PREFIX &&
       // Skip if already on the new locale unless the strategy is "prefix_and_default" and this is the default
-      // locale, in which case we might still redirect as we prefer unprefixed route in this case.
-      (getLocaleFromRoute(route) !== newLocale || (options.strategy === Constants.STRATEGIES.PREFIX_AND_DEFAULT && newLocale === options.defaultLocale))
+      // locale, in which case we might still redirect as we prefer unprefixed route in this case, but let user a
+      // possibility to disable this behavior by switching disableDefaultRedirect option to true.
+      (isSameLocale || (isPrefixAndDefaultStrategy && isDefaultLocale && isEnabledRedirectForRoute))
```
rchl: I feel like having this would only solve part of the issue as linking to such page using `localePath('about')` would still resolve and navigate to locale-less path. If you are willing to work on it more to also handle that case then I'll be willing to accept it.

Note that if this would be handled in the path resolving code (`localePath` and co.) then you probably wouldn't even need to change this part.

rchl (changes requested):
This would also need documentation to be updated with new option and some tests.

src/templates/plugin.main.js
```diff
@@ -153,18 +153,22 @@ export default async (context) => {
...
       // Skip if already on the new locale unless the strategy is "prefix_and_default" and this is the default
-      // locale, in which case we might still redirect as we prefer unprefixed route in this case.
-      (getLocaleFromRoute(route) !== newLocale || (options.strategy === Constants.STRATEGIES.PREFIX_AND_DEFAULT && newLocale === options.defaultLocale))
+      // locale, in which case we might still redirect as we prefer unprefixed route in this case, but let user a
+      // possibility to disable this behavior by switching disableDefaultRedirect option to true.
+      (isDifferentLocale || (isPrefixAndDefaultStrategy && isDefaultLocale))
     ) {
-      // The current route could be 404 in which case attempt to find matching route using the full path since
-      // "switchLocalePath" can only find routes if the current route exists.
-      const routePath = app.switchLocalePath(newLocale) || app.localePath(route.fullPath, newLocale)
```
Renhor: [rchl](https://github.com/rchl) With my current implementation switchLocalePath() always redirects me to the prefixed path and I couldn't think of anything better yet :(

Renhor:
[rchl](https://github.com/rchl) Thanks for the quick review. Unfortunately, I will only be able to continue working on the task today. I decided to abandon the route array and changed the implementation following your advice. Updated the information in the first post

types/index.d.ts
Renhor: What about 
```
type PrefixAndDefaultRule = 'default' | 'prefix'

interface PrefixAndDefaultRules {
  switchLocal: PrefixAndDefaultRule // 'default'
  routing: PrefixAndDefaultRule // 'default'
}

i18n: {
  prefixAndDefaultRules?: PrefixAndDefaultRules
}
```

src/templates/utils-common.js
Renhor: Thanks for advice, i begin to use ufo

src/templates/plugin.utils.js
Renhor: Currently i abandon the route array cuz i can't find it useful in the current implementation

docs/content/en/api.md
```diff
@@ -28,12 +28,18 @@ All [Vue I18n properties and methods](http://kazupon.github.io/vue-i18n/api/#vue
 
   - **Arguments**:
     - locale: (type: `string`)
+    - forcePrefix: (type: `boolean`)
```
rchl: I'd want to keep the changes as simple as possible while fixing your use case.

Is this new argument necessary? I have hard time imagining why would only sometimes want or not want the locale prefix.

docs/content/en/options-reference.md
```diff
@@ -186,6 +186,18 @@ Set this to `true` when using different domains for each locale. If enabled, no
 
 Whether [custom paths](/routing#custom-paths) are extracted from page files using babel parser.
 
+## `prefixAndDefaultRules`
+
+- type: `object`
+- default: `{ routing: 'default', switchLocale: 'default' }`
```
rchl: Is it necessary to have separate settings for `routing` and `switchLocale`?

rchl (changes requested):
I have not looked properly at the changes yet because many tests are failing.

docs/content/en/options-reference.md
Renhor: I agree, it's better to keep only one property

docs/content/en/api.md
Renhor: By adding a split for `{ routing: 'default', switchLocale: 'default' }`, I thought that in some situations we wouldn't want to globally change the behavior of the method to something else, so I decided to add this argument.

However, I now think that such a situation is actually hard to imagine.

docs/content/en/options-reference.md
Renhor: I agree, let's keep a single property for the behavior

Renhor:
Corrected the code that was causing the tests to crash

juliomontilla100:
I need this feature ! please, thanks... @Renhor any update bro?

DanielHefti:
This feature would be highly appreciated. Get in touch with me, if there is anything I can do.

Issue #1161, opened 2021-04-30
Different domains with multiple languages each


toooni:
I am not a 100% sure if this really isn't possible yet, but it seems like it isn't. We have multiple TLD's and one of them (.ch) needs to have two languages which should result in the following domain/path/locale combinations:
- weekend4two.ch `de-CH`
- weekend4two.ch/fr `fr-CH`
- weekend4two.de `de-DE`
- weekend4two.at `de-AT`
- weekend4two.fr/fr OR weekend4two.fr `fr-FR` (We do not really care since the .fr TLD does only need one language)

I've tried a few configs to achieve what I need but had no success. This one was the closest:

```
/* nuxt.config.js */
...
i18n: {
  differentDomains: true,
  strategy: 'prefix_except_default',
  locales: [
    {
      code: 'de-CH',
      iso: 'de-CH',
      domain: 'weekend4two.ch',
    },
    {
      code: 'fr-CH',
      iso: 'fr-CH',
      domain: 'weekend4two.ch',
    },
    {
      code: 'fr-FR',
      iso: 'fr-FR',
      domain: 'weekend4two.fr',
    },
    {
      code: 'de-DE',
      iso: 'de-DE',
      domain: 'weekend4two.de',
    },
    {
      code: 'de-AT',
      iso: 'de-AT',
      domain: 'weekend4two.at',
    },
  ],
}
...
  
/* index.vue */
...
<script>
export default {
  nuxtI18n: {
    paths: {
      de: '/',
      fr: '/fr',
      'fr-CH': '/fr',
    },
  },
}
</script>
...
```

The seems to result in a working solution where all the domains/path exist. But when going to `weekend4two.ch/fr`, the locale is set to `de-CH`. Whatever I do, it takes the locale of the first entry with the same domain.

### Describe the solution you'd like
It would be cool if the locales config in `nuxt.config.js` could be extended with a `prefix` property like this:
```
/* nuxt.config.js */
locales: [
  {
    code: 'de-CH',
    iso: 'de-CH',
    domain: 'weekend4two.ch',
    prefix: '', /* could be removed too */
  },
  {
    code: 'fr-CH',
    iso: 'fr-CH',
    domain: 'weekend4two.ch',
    prefix: '/fr',
  },
  ...
]
```

### Describe alternatives you've considered
I am open to all alternatives. It may be possible that it even is possible in a way I haven't figured out yet. I am new to NUXT.

andorfermichael:
[...]

PR #1461, opened 2022-06-23
Feat: allow multiple locales under a single domain

mbomfim33:
This PR is still a WIP. I'm working on implementing the feature mentioned on #1161 based on the work by [lfglopes](https://github.com/lfglopes). Since there were quite a few updates after this issue first got worked on, there are some things that still need some work, as well as some refactor once I'm done. 

Checklist before movinfg this PR to ready-to-review:
- [ ] Fix `switchLocalePath` when switching languages so it points to `newdomain + path` instead of `currentdomain + locale-code + path`.
- [ ] Fix tests that are failing probably due to the above
- [ ] Add tests to new feature
- [ ] Fix missing types
- [ ] Update documentation

### DIFF
--- a/src/helpers/routes.js
+++ b/src/helpers/routes.js
@@ -20,7 +20,7 @@ export function makeRoutes (baseRoutes, {
   defaultLocaleRouteNameSuffix,
   differentDomains,
   includeUnprefixedFallback,
-  localeCodes,
+  locales,
   pages,
   pagesDir,
   parsePages,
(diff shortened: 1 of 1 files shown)


src/helpers/routes.js
```diff
@@ -145,7 +157,11 @@ export function makeRoutes (baseRoutes, {
       )
 
       if (shouldAddPrefix) {
-        path = `/${locale}${path}`
+        path = `/${locale.code}${path}`
```
mbomfim33, line 160: When running `no-lang-switcher`, `locale.code` is undefined, so I have to figure out a way of checking the type of locale (`object` for multiple-locales, and `string` for `no-lang-switcher`

PR #1471, opened 2022-07-30
feat: add redirectStatusCode option

Matiyeu:
This PR add ability to customize the status code to redirect unprefixed route to the default locale(when strategy: prefix) or default locale to unprefixed route (when strategy: prefix_and_default)

example : 
```js
{
  strategy: 'prefix',
  redirectStatusCode: 301
}
```
domain.com => 301 => domain.com/en
domain.com/about-us => 301 => domain.com/en/about-us

### DIFF
--- a/docs/content/en/options-reference.md
+++ b/docs/content/en/options-reference.md
@@ -172,6 +172,13 @@ Set to a path to which you want to redirect users accessing the root URL (`/`). 
 }
 ```
 
+## `redirectStatusCode`
+
+- type: `number`
+- default: `302`
+
+Customize the status code to redirect unprefixed route to the default locale.
+
 ## `differentDomains`
 
 - type: `boolean`
--- a/src/helpers/constants.js
+++ b/src/helpers/constants.js
@@ -42,6 +42,7 @@ export const DEFAULT_OPTIONS = {
   lazy: false,
   langDir: null,
   rootRedirect: null,
+  redirectStatusCode: 302,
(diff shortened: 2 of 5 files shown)


rchl:
I don't like how generic this setting is named and at the same time not explaining exactly what it does. To average user it won't be clear in what circumstances this redirect applies without analyzing the code.

But probably the bigger problem is that in most cases it wouldn't be safe to override this setting and people might corner themselves into issues because of using it.

For example with this option configured to respond with 301, if someone comes to a site and gets redirected from an english to german locale but then wants to manually switch to english then it won't be possible because the browser will force-redirect back to german.

I think that an option like this would only make sense if it only applied in more specific cases where it would be appropriate. For example when the URL that triggered the redirect is 404. And it shouldn't apply if the redirect was due to auto-detecting browser locale as then the user might want to switch back to locale that it redirected from.
But that would require more serious changes in the logic.

baptistejamin:
According to Google, https://developers.google.com/search/docs/crawling-indexing/301-redirects , 301 redirects are very important signals.

For instance, at Crisp, we enforce /lang/page/ whoever, and some sites are still referencing us using /page/ , such an option would have important benefits for us in terms of SEO and crawl budget.

While it's right, most users won't care and will be happy with 302, there are many cases where we would be happy to get a such PR accepted, even if there is no fallback. We just want  strict 301 redirects rather than 302.

BobbieGoede:

PR #1488, opened 2022-08-13
feat: allow composable

dargmuesli:
Uses https://github.com/intlify/vue-i18n-composable/ to allow usage of this module in Vue 2.7's setup function. The currently supported usage outside of the setup functions remains working in my testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,8 @@
     "klona": "^2.0.5",
     "lodash.merge": "^4.6.2",
     "ufo": "^0.8.5",
-    "vue-i18n": "^8.27.2"
+    "vue-i18n": "^8.27.2",
+    "vue-i18n-composable": "^2.0.0"
   },
   "devDependencies": {
     "@babel/core": "7.18.10",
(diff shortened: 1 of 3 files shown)


dargmuesli:
Could close #1333 as well.

rchl:
According to its documentation, the v2 version only works with Vue 2.7 and not 2.6 so we can't use it.

rchl:
And note that this would only cover the `vue-i18n` APIs while this module provides a bunch of additional API that are specific to this module.

dargmuesli:
You're right, I knew this would be too easy 😄
So the only option is to wait for the next branch for composition support? Or are there any other possibilities to use the composition api with Vue 2.7 and vue-i18n v8 (= nuxt-i18n v7)?

dargmuesli:
I made different branches for different extensions of nuxt-i18n, like with vue-i18n-composable or vue-i18n-bridge in https://github.com/dargmuesli/nuxt-i18n-module/. For anyone interested in trying one of those out while v8 and v7.3 of nuxt-i18n are in progress, feel free to have a look!
Closing this PR in favor of `@next` and a Nuxt v3 migration 🤓.

PR #1491, opened 2022-08-23
[next] fix(module): remove nuxt 2 error

dargmuesli:
Nuxt 2 can be used with bridge, which supports running Nuxt 3 modules.
@kazupon

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -34,16 +34,6 @@ export default defineNuxtModule<NuxtI18nOptions>({
 
     checkOptions(options)
 
-    if (isNuxt2(nuxt)) {
-      throw new Error(
-        formatMessage(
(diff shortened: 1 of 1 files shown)


rchl:
I don't know if anything has changed since last time we've [talked about it with @kazupon ](https://github.com/nuxt-community/i18n-module/discussions/1287#discussioncomment-3042457) but the conclusion was that v8 won't support Nuxt 2 since that's easier to do for now.

So I'm not convinced that this is the right thing to do but @kazupon should hopefully be able to tell.

dargmuesli:
In my investigation I found vue-i18n v9 only being incompatible with Vue 2.7 at usages of `vue.Fragment`, `vue.Text` and `vue.createVNode`. I'll keep https://github.com/intlify/vue-i18n-next/pull/1130 open and close this PR for now. If v9 turns out to be compatible with Vue 2.7 as well, I'll reopen this PR.

PR #1521, opened 2022-09-19
[next] docs(module): point to alternatives for Nuxt 2

dargmuesli:
@kazupon

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -37,9 +37,9 @@ export default defineNuxtModule<NuxtI18nOptions>({
     if (isNuxt2(nuxt)) {
       throw new Error(
         formatMessage(
-          `We will release >=7.3 <8, See about GitHub Discussions https://github.com/nuxt-community/i18n-module/discussions/1287#discussioncomment-3042457: ${getNuxtVersion(
+          `v8 does not support Nuxt 2 and you're using ${getNuxtVersion(
             nuxt
-          )}`
+          )} (see https://github.com/nuxt-community/i18n-module/discussions/1287#discussioncomment-3042457). You might want to check out https://github.com/intlify/vue-i18n-next/tree/master/packages/vue-i18n-bridge for migration to vue-i18n@v9 or https://github.com/intlify/vue-i18n-composable/ to make use of the composition api with vue-i18n@v8.`
         )
       )
     }


src/module.ts
```diff
@@ -37,9 +37,9 @@ export default defineNuxtModule<NuxtI18nOptions>({
     if (isNuxt2(nuxt)) {
       throw new Error(
         formatMessage(
-          `We will release >=7.3 <8, See about GitHub Discussions https://github.com/nuxt-community/i18n-module/discussions/1287#discussioncomment-3042457: ${getNuxtVersion(
+          `v8 does not support Nuxt 2 and you're using ${getNuxtVersion(
```
rchl, line 40: Is it useful to print specific Nuxt version? Why not just say that Nuxt 2 is not supported since we know this is running under Nuxt 2?

src/module.ts
```diff
@@ -37,9 +37,9 @@ export default defineNuxtModule<NuxtI18nOptions>({
...
             nuxt
-          )}`
+          )} (see https://github.com/nuxt-community/i18n-module/discussions/1287#discussioncomment-3042457). You might want to check out https://github.com/intlify/vue-i18n-next/tree/master/packages/vue-i18n-bridge for migration to vue-i18n@v9 or https://github.com/intlify/vue-i18n-composable/ to make use of the composition api with vue-i18n@v8.`
```
rchl, line 42: I don't quite understand those suggestions. I think they are quite confusing.

Essentially those suggestions mention an alternative (`vue-i18n`) while this is the nuxt module so not quite the same. I'm fine with suggesting plain `vue-i18n` but it should be made clear that it's not the module and doesn't offer any of the module-specific functionality.

And if suggesting to use plain `vue-i18n` then probably there should be also proper instructions how to set it up in Nuxt including server-side support, maybe also yaml-loader stuff. I suppose the whole thing might require a page of its own.

src/module.ts
dargmuesli: True, I thought about removing it, but decided to keep it because @kazupon put it there and I didn't want to interfere with that. I can surely remove it.

src/module.ts
dargmuesli: Right, it'd be better to keep all possibilities under one umbrella in this repo.
Would it be a possiblity to just maintain two diverging branches as well, which use `vue-i18n-bridge` and `vue-i18n-composable`, as I did in my fork [[1](https://github.com/dargmuesli/nuxt-i18n-module/tree/feat/bridge)] [[2](https://github.com/dargmuesli/nuxt-i18n-module/tree/feat/composable)] and point to those instead?
That would allow people to keep using `@nuxtjs/i18n-module` and benefit from composition or bridge functionality, right?  They could use something like `"@nuxtjs/i18n": "nuxt-community/i18n-module#3b1c957",` or a dedicated tag on npm.

Please tell me if I understand something incorrectly. I want to enable v7 users to migrate more easily, but it seems quite a challenge. Please excuse me if my suggestions bother.

kazupon: Thank you for your contribution!

As @rchl said, if we use `vue-i18n-bridge` and `vue-i18n-composable`, I think we need to integrate it to work as the nuxt module fully.

I mean we should first provide a version of nuxtjs/i18n v7.4 on the nuxt bridge that is guaranteed to work with the options API so that the nuxt community can migrate to nuxtjs/i18n v8. Then, once that is released, we should backport the composables API from the nuxtjs/i18n v8 version and the composable API such as `useI18n` in vue-i18n to a version that works on the nuxt bridge as nuxtjs/i18n v7.5.

We believe that by preparing in the above release roadmap, it will be possible to migrate incrementally to nuxt2 + nuxtjs/i18n v7, -> nuxt2 + nuxt bridge + nuxtjs/i18n v7.4 or v7.5, and finally nuxt3 + nuxtjs/i18n v8.

dargmuesli: Gotcha, but isn't `@nuxtjs/i18n` v7 compatible with `nuxt bridge` right now already? I used it with `nuxt bridge` for about a year now.

kazupon: @dargmuesli 
We haven't yet tested if it works with nuxt2 + nuxt bridge combination. I believe that only after that test is done, we can officially announce the nuxt bridge support.
However, the support does not provide `localePath` provided by nuxt3, composable `useI18n` of vue-i18n and etc,  So when we make a formal announcement, we should also announce our roadmap.

dargmuesli: Ok, it seems I haven't yet understood what's intended to be supported at which version number. Aside, I understand why the changes I proposed here are confusing, so I'm closing this PR.

kazupon: @dargmuesli 
sorry, for my poor describing 🙇 
I mean the below:

- `@nuxtjs/i18n` v7.4: support nuxt bridge
- `@nuxtjs/i18n` v7.5: support some vue-i18n & nuxtjs/i18n composables that is ported from nuxt3

dargmuesli: Ok, I got that :+1: what is needed to be done to confirm Nuxt bridge compatibility? How can one help? :)

kazupon: Thanks! 💚 

I think we can test compatibility by nuxt bridge installed on e2e in the main branch of nuxtjs/i18n.
@rchl has more information about the existing environment. If you have any questions, you can ask @rchl.

rchl: Yeah, I think @kazupon has the right idea. If you can make all tests on the `main` branch work with Nuxt+Bridge then that would be a good indication that things work OK.

PR #1522, opened 2022-09-20
fix: improve error handling for vue-i18n v9 invalid message syntax

### LINKED ISSUE #1519: Having an "@" symbol in the locale JSON file breaks the plugin and results in a 500 error
### Version
@nuxtjs/i18n: 8.0.0-alpha.1
nuxt: 3.0.0-rc.9

### Nuxt configuration

```js
i18n: {
        baseUrl: 'https://studioterabyte.nl',
        locales: [
            {
                code: 'en',
                iso: 'en-US',
                file: 'en-US.js'
            },
            {
                code: 'nl',
                iso: 'nl-NL',
                file: 'nl-NL.js'
            }
        ],
        lazy: true,
        langDir: 'locales/',
        defaultLocale: 'nl',
        vueI18n: {
            legacy: false,
            fallbackLocale: 'nl',
        }
    }
```

### What is Expected?
For the correct locale to be loaded based on their respective `.json` file. 

When trying to load my locales via `JSON` files I kept encountering an error messages indicating that the message property could not be read:

```zsh
 ERROR  Cannot read properties of undefined (reading 'message')    
```

After some investigating I found that having an "@" symbol anywhere in the file will cause the plugin to return the error message above and render the 500 error page. I've recreated the error here:

https://stackblitz.com/edit/nuxt-starter-xohyax?file=package.json

Thanks!

kazupon:
resolve #1519

Invalid message syntax in vue-i18n v9 (e.g. `@` unescaped with literal intrepolation),  error handling was not handled properly in `@intlify/unplugin-vue-i18n` (`@intlify/bundle-utils`), resulting in a 500 error in nuxt, we can understand it difficult to know where message was.

That issue has been already fixed in the following PR so that we will upgrade `@intlify/unplugin-vue-i18n`.

related issue: [intlify/bundle-tools#164](https://github.com/intlify/bundle-tools/issues/164)
related PR: [intlify/bundle-tools#173](https://github.com/intlify/bundle-tools/pull/173)

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,9 +66,9 @@
     "test:spec": "yarn build && vitest run specs"
   },
   "dependencies": {
-    "@intlify/bundle-utils": "next",
+    "@intlify/bundle-utils": "^3.1.2",
     "@intlify/shared": "latest",
-    "@intlify/unplugin-vue-i18n": "latest",
+    "@intlify/unplugin-vue-i18n": "^0.6.0",
     "@nuxt/kit": "^3.0.0-rc.9",
     "cookie-es": "^0.5.0",
     "debug": "^4.3.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -580,9 +580,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@intlify/bundle-utils@npm:next":
-  version: 3.1.0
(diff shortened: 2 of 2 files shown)


kazupon:
image

kazupon:
The v8 docs build is failing, but CI is passing. So I will merge them (Later I will solve it).

PR #1549, opened 2022-10-08
refactor: seo docs

kazupon:


kazupon:
This PR is docs refactoring.
So, since there is no effect on CI, I will merge.

PR #1576, opened 2022-10-21
Add .ts and .js as supported file extensions

Aareksio:
Reviewing the source code of `next` branch I found no obstacles that would disallow loading locales from `js` and `ts` files. It does not support exporting (async) functions, but should be enough for requested feature of joining localization components:

```js
// en.js

import home from '../components/en/home.json'
import about from '../components/en/about.json'

export default {
  home,
  about,
}
```

### DIFF
--- a/specs/fixtures/lazy/lang/pl.ts
+++ b/specs/fixtures/lazy/lang/pl.ts
@@ -1,0 +1,6 @@
+export default {
+  home: 'Strona główna',
+  about: 'O nas',
+  posts: 'Publikacje',
+  dynamic: 'Dynamiczne'
+}
--- a/specs/lazy_load.spec.ts
+++ b/specs/lazy_load.spec.ts
@@ -24,6 +24,12 @@ await setup({
           iso: 'fr-FR',
           file: 'fr.json5',
           name: 'Français'
+        },
+        {
+          code: 'pl',
+          iso: 'pl-PL',
+          file: 'pl.ts',
+          name: 'Polski'
         }
       ]
     }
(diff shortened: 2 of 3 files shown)


rchl:
Ultimately it's up to @kazupon to decide but based on functionality available in the Nuxt 2 version, I would say that support for async functions is something that many would expect and use. So introducing "half-working" solution is probably not a great idea.

Aareksio:
@rchl I disagree with calling it "half working" idea.

This feature is requested by many users:
<https://github.com/nuxt-community/i18n-module/issues/1448#issue-1244029314>
<https://github.com/nuxt-community/i18n-module/issues/1448#issuecomment-1134362263>
<https://github.com/nuxt-community/i18n-module/issues/1448#issuecomment-1165144797>
<https://github.com/nuxt-community/i18n-module/issues/1462#issue-1288731808>
<https://github.com/nuxt-community/i18n-module/issues/1462#issuecomment-1194065154>
<https://github.com/nuxt-community/i18n-module/issues/1511#issue-1368692385>

The PR is intentionally kept very compact, to avoid any changes outside of the scope of allowing using **static** JS/TS files as translation source. Currently the module is in a state where everyone attempting to use JS/TS translations encounters a **silent** error. 

With this PR:

<table>
<tr>
	<td>Usage
	<td>Before PR
	<td>After PR
<tr>
	<td>JS/TS object
	<td>silent fail
	<td>works
<tr>
	<td>JS/TS function
	<td>silent fail
	<td>"Not support executable file (e.g. js, cjs, mjs)"
</table>

For me it's a win/win situation, instead of silent failure, [users get proper error message](https://github.com/nuxt-community/i18n-module/blob/next/src/runtime/internal.ts#L97).

Consciously limiting scope != half working.

kazupon:

There is a reason why we currently limit ourselves to formats such as json, json5, and yml.
In order to fully support i18n resources for executable files like js/ts, `@intlify/unplugin-vue-i18n` used as a vite plugin must support them.

Currently, vue-i18n v9 used as the core of @nuxtjs/i18n, transpiles locale messages into [message functions](https://vue-i18n.intlify.dev/guide/advanced/function.html) that can be executed as JS.

Currently supported formats such as json, json5 and, yml also support sourcemap when transpiled. sourcemap must be supported in order to support js/ts as a locale messages resource.
The case of message functions must also be considered.

Aareksio:
@kazupon This makes a lot of sense. It was too easy to enable the static file support. 

The [generator function in bundle-tools](https://github.com/intlify/bundle-tools/blob/main/packages/bundle-utils/src/json.ts) seems quite complex for JSON alone, I imagine for JS it's going to be even more extensive as there are more edge cases. Definitely out of realm for a first time contribution.

Is there an approach you would recommend for supporting splitting translation files into components?

```
|- lang
 |- en/
  |- home.json
  |- profile.json
```

Since `@intlify/unplugin-vue-i18n` we could perhaps leverage the generated JS for each component and join them somehow. Is it even worth a try in your opinion?

irishkooky:
I would like to use this feature, is this pull request going to be merged?

Aareksio:
The approach presented in this PR is lacking. It is **unlikely** to get merged.
Support must be added first to `@intlify/unplugin-vue-i18n`.

irishkooky:
@Aareksio 
I see. Thanks!

pumano:
any news on this feature? in my case I need to use async function for getting i18n data from server and just need to support functions and js files. That works well in nuxt 2 but not works in nuxt 3. Thanks very frustrating because there no breaking change info about it in documentation when updating.

@kazupon if you can tell more about what to do, I can find time for help with that support. Just need to know what to do for add support. Because that feature need me ASAP :)

pumano:
add for linking issue https://github.com/nuxt-modules/i18n/issues/1829

PR #1609, opened 2022-11-04
fix: wrong relative import path when srcDir is set

so1ve:


kazupon:
LGTM! 👍

PR #1617, opened 2022-11-06
fix(gen): consider named functions (#1616)

### LINKED ISSUE #1616: [8.0.0-beta-3 - Nuxt 3 RC13] Uncaught SyntaxError: Unexpected token '{'
### Version
@nuxtjs/i18n: 8.0.0-beta.3
nuxt: 3.0.0-rc.12

```
i18n: {
  lazy: true,
  langDir: 'locales',
  strategy: 'prefix',
  defaultLocale: 'en',
  detectBrowserLanguage: {
    useCookie: true,
    cookieKey: 'i18n_redirected',
    redirectOn: 'root',
  },
  locales: [
    {
      code: 'en',
      name: 'English',
      file: 'en.json',
      customFunctions: {
        greeting({ name }) {
           return `Hello ${name}.`
        },
      },
    },
    {
      code: 'es',
      name: 'Español',
      file: 'es.json',
    },
  ],
  vueI18n: {
    legacy: false,
    locale: 'en',
  },
}
```

### Description

`i18n.locales[0].customFunctions` will break https://github.com/nuxt-community/i18n-module/blob/next/src/gen.ts#L193 resulting in
```
{ "greeting": (greeting({name}) { return `Hello ${name}.`}) }
```
It needs to distinguish between arrow functions and named functions - I'll open a PR!

ineshbose:
resolve #1616 
Along with function name typo fix.

### DIFF
--- a/src/gen.ts
+++ b/src/gen.ts
@@ -113,7 +113,7 @@ export function generateLoaderOptions(
       codes += `}\n`
       return codes
     } else if (rootKey === 'additionalMessages') {
-      return `export const ${rootKey} = ${generaeteAdditionalMessages(rootValue, dev)}\n`
+      return `export const ${rootKey} = ${generateAdditionalMessages(rootValue, dev)}\n`
 	  } else {
 	    return `export const ${rootKey} = ${toCode(rootValue)}\n`
 	  }
@@ -150,7 +150,7 @@ function generateVueI18nOptions(options: Record<string, any>, dev: boolean): str
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-function generaeteAdditionalMessages(value: Record<string, any>, dev: boolean): string {
+function generateAdditionalMessages(value: Record<string, any>, dev: boolean): string {
   let genCode = 'Object({'
   for (const [locale, messages] of Object.entries(value)) {
     genCode += `${JSON.stringify(locale)}:[`
@@ -191,7 +191,7 @@ export function toCode(code: any): string {
   }
 
   if (isFunction(code) && code.toString) {
-    return `(${code.toString()})`
+    return `(${code.toString().replace(new RegExp(`^${code.name}`), 'function ')})`
   }
 
   if (isArray(code)) {
--- a/test/__snapshots__/gen.test.ts.snap
+++ b/test/__snapshots__/gen.test.ts.snap
@@ -68,9 +68,9 @@ export const resolveNuxtI18nOptions = async (context) => {
 },}),})
   nuxtI18nOptions.vueI18n = await vueI18nOptionsLoader(context)
   nuxtI18nOptions.locales = [Object({\\"code\\":\\"en\\",\\"file\\":\\"en.json\\",\\"path\\":\\"/path/to/en.json\\",\\"testFunc\\":((prop) => {
-          return \`Hello \${prop.toString()}\`;
+          return \`Hello \${prop}\`;
         })}),Object({\\"code\\":\\"ja\\",\\"file\\":\\"ja.json\\",\\"path\\":\\"/path/to/ja.json\\",\\"testFunc\\":((prop) => {
-          return \`Hello \${prop.toString()}\`;
+          return \`Hello \${prop}\`;
         })})]
   return nuxtI18nOptions
 }
@@ -95,9 +95,9 @@ export const resolveNuxtI18nOptions = async (context) => {
 },}),})
   nuxtI18nOptions.vueI18n = await vueI18nOptionsLoader(context)
   nuxtI18nOptions.locales = [Object({\\"code\\":\\"en\\",\\"file\\":\\"en.json\\",\\"path\\":\\"/path/to/en.json\\",\\"testFunc\\":(function (prop) {
-          return \`Hello \${prop.toString()}\`;
+          return \`Hello \${prop}\`;
         })}),Object({\\"code\\":\\"ja\\",\\"file\\":\\"ja.json\\",\\"path\\":\\"/path/to/ja.json\\",\\"testFunc\\":(function (prop) {
-          return \`Hello \${prop.toString()}\`;
+          return \`Hello \${prop}\`;
         })})]
   return nuxtI18nOptions
 }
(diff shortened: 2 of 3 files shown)


src/gen.ts
```diff
@@ -191,7 +191,7 @@ export function toCode(code: any): string {
   }
 
   if (isFunction(code) && code.toString) {
-    return `(${code.toString()})`
+    return `(${code.toString().replace(new RegExp(`^${code.name}`), 'function ')})`
```
kazupon, line 194: Could you add test cases to unit tests, please? 🙏

kazupon (changes requested):

kazupon:
@ineshbose 
GitHub Action failed in Lint
I think you need to format with `pnpm fix`

ineshbose:
My bad! Didn't see the pipeline result for that. Should be fixed.

kazupon (approved):

kazupon:
LGTM!

Thank you so much!

kazupon:
closes #1616

Aareksio:
@ineshbose @kazupon
This fails when the function is anonymous / has no `.name`:

```js
const createAnynomousFunction = () => (() => 1)
code = createAnynomousFunction()

console.log(`(${code.toString().replace(new RegExp(`^${code.name}`), 'function ')})`)
// Logs: "(function () => 1)" instead of "(() => 1)"
```

Real use case, implementing CLDR pluralization rules:
```js
const createPluralizationRule = (formatterFn) => {
  return (value, choicesList) => {
    const stringValue = String(value)
    const decimalPart = !stringValue.includes('.') ? '' : stringValue.split('.')[1]
    const decimalLength = decimalPart.length
    const numericValue = +value
    const integerPart = +(stringValue.split('.')[0])
    const truncatedDecimal = decimalPart.length === 0 ? 0 : +decimalPart.replace(/0+$/, '')

    const choice = formatterFn(numericValue, integerPart, decimalLength, +decimalPart, truncatedDecimal)
    return choice < choicesLength ? choice : choicesLength - 1
  }
}

export default {
  pluralizationRules: {
    de: createPluralizationRule((_n, i, v) => i === 1 && v === 0 ? 0 : 1)
  }
}
```

ineshbose:
@Aareksio thanks for the feedback. I tried it with anonymous functions (not quite the same way you have however using a higher order function), but in those cases, `.name` evaluates to `undefined` which is unlikely to be replaced in the template RegEx. Could you share your resulting `.nuxt/i18n.options.mjs`?

@kazupon I think as mentioned before, ideally it would be best to rewrite the script using an AST walker?

Aareksio:
```js
export const resolveNuxtI18nOptions = async (context) => {
  const nuxtI18nOptions = Object({})
  const vueI18nOptionsLoader = async (context) => Object({"pluralizationRules":Object({"de":(function (value, choicesLength) => {
    const stringValue = String(value);
    const decimalPart = !stringValue.includes('.') ? '' : stringValue.split('.')[1];
    const decimalLength = decimalPart.length;
    const numericValue = +value;
    const integerPart = +stringValue.split('.')[0];
    const truncatedDecimal = decimalPart.length === 0 ? 0 : +decimalPart.replace(/0+$/, '');

    const choice = formatterFn(numericValue, integerPart, decimalLength, +decimalPart, truncatedDecimal);
    return choice < choicesLength ? choice : choicesLength - 1;
  })}),})
  nuxtI18nOptions.vueI18n = await vueI18nOptionsLoader(context)
  // ...
  return nuxtI18nOptions
}
```

This approach doesn't work anyway, as `function.toString()` doesn't pass the context required for the above code to work. You can see `formatterFn` is undefined - must use file loader (currently broken on Windows). 

That said, the input is **valid** javascript, while the output has invalid syntax.

Probably easiest _dirty_ fix would be to do something like the below, to avoid `/^/` regex when `code.name === ''`:
``(${code.toString().replace(new RegExp(`^${code.name || 'undefined'}`), 'function ')})``

ineshbose:
@kazupon already on it! 😉 

@Aareksio I'm able to understand and replicate your issue with all versions of `@nuxtjs/i18n@v8.0.0-beta.x`. This change was included in beta 5 and aims to resolve issues with syntax error while using named functions in an object as examples in #1616. Even prior to this change, the same issue occurs where `formatterFn` is undefined - hence this PR is unrelated to the raised issue (please feel free to correct me if I'm wrong and provide a reproduction for beta.4 if possible where it works).

What I'm analysing is that in

```js
export default {
  pluralizationRules: {
    de: createPluralizationRule((_n, i, v) => i === 1 && v === 0 ? 0 : 1)
  }
}
```

The function `createPluralizationRule` is a function that is CALLED and hence giving a return value instead of pointing to a function, so in the case where you had

```js
export default {
  pluralizationRules: {
    de: createPluralizationRule
  }
}
```

would be perfectly valid with `formatterFn` also defined, but I understand that is not what you want. `code.toString()` (implemented and used through all versions of `@nuxtjs/i18n@v8.0.0-beta.x`) is also not a very good solution either, I agree, which is part of the problem, so the relevant file (`src/gen.ts`) should have a rewrite using a walker (same treatment as `src/pages.ts` through https://github.com/nuxt-community/i18n-module/commit/bc06d8993c4c525ed1f6aa914887729858c4f639) depending on @kazupon's thoughts.

Meanwhile it would be great to open a separate issue for this (titled like "Function context undefined due to code.toString()" or "gen.ts does not support higher order functions"). 🙂

Aareksio:
> Even prior to this change, the same issue occurs where formatterFn is undefined - hence this PR is unrelated to the raised issue (please feel free to correct me if I'm wrong and provide a reproduction for beta.4 if possible where it works).

Exactly, I pointed this out in the reply above. The `formatterFn` is undefined after and before this change - not important for the case. I am also not raising an issue out of this problem, as the [documentation clearly states alternative solution](https://v8.i18n.nuxtjs.org/options/vue-i18n#vuei18n), for complex data types, which is providing a path to the config file.

---

What is important, is that a function with a name `''` is processed into **invalid syntax** `function () => {}` because of the regex.
As per reproduction:

```js
const createFunctionWithNoName= () => (() => 1)
code = createFunctionWithNoName()

console.log(`(${code.toString().replace(new RegExp(`^${code.name}`), 'function ')})`)
// Logs: "(function () => 1)" instead of "(() => 1)"
```

ineshbose:
Thanks for the reproduction. I see your point and also agree the dirty fix would work. I've taken a note of this and trying to rework the file to consider all possibilities (such as returning functions, using contexts, etc, as you raised). Meanwhile please use the alternative solution where possible. 🙏🏻

PR #1622, opened 2022-11-08
Changed canonical queries example

Simlor:
In my local project, I got an error message until I changed it to this notation. Perhaps I just overlooked something else?


kazupon (approved):
Good catch!
Thanks!

PR #1653, opened 2022-11-18
fix: resolve deps modules

### LINKED ISSUE #1518: [NEXT] Crashes on rc.9 and rc.10 `Only file and data URLs are supported by the default ESM loader`
### Version
@nuxtjs/i18n: `@nuxtjs/i18n@8.0.0-alpha.1-27710980.d1ee9d2`
nuxt: `3.0.0-rc.9`

### Nuxt configuration

 - [ ] Applies to a site deployed to a static server (site generated with `nuxt generate`)
 - [x] Applies to a site deployed to a server with a Node backend

### Reproduction Link

I tried updating from rc8 to rc9 today and got the error message below when trying to start in dev mode.
Tried deleting .nuxt, node_modules as well as package-lock.json to get a fresh install but it all resulted in the same error. Same story for rc10 and edge rc11. 

https://github.com/warflash/nuxt_7426

### Steps to reproduce

Clone the repo on windows and run `npm run dev`

### What is Expected?

The instance starts as normal

```shell
500
Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'f:'

at new NodeError (node:internal/errors:371:5)
at defaultResolve (node:internal/modules/esm/resolve:1016:11)
at ESMLoader.resolve (node:internal/modules/esm/loader:422:30)
at ESMLoader.getModuleJob (node:internal/modules/esm/loader:222:40)
at ModuleWrap. (node:internal/modules/esm/module_job:76:40)
at link (node:internal/modules/esm/module_job:75:36)
```

### LINKED ISSUE #1613: [NEXT] Cannot find module vue-i18n.mjs
### Version
@nuxtjs/i18n: 8.0.0-beta.3
nuxt: 3.0.0-rc.13
vue-i18n: "^9.2.2"

### Steps to reproduce
Start nuxi dev

### What is Expected?
Starting dev

Cannot start nuxt:  Cannot find module 'nuxt\node_modules\vue-i18n\dist\vue-i18n.mjs'

kazupon:
resolve #1518
resolve #1613

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "magic-string": "^0.26.7",
     "mlly": "^1.0.0",
     "pathe": "^1.0.0",
+    "pkg-types": "^1.0.1",
     "ufo": "^1.0.0",
     "unplugin": "^1.0.0",
     "vue-i18n": "^9.3.0-beta.7",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -44,6 +44,7 @@ importers:
       npm-run-all: ^4.1.5
       nuxt: 3.0.0-rc.12
       pathe: ^1.0.0
+      pkg-types: ^1.0.1
       playwright: ^1.27.1
       prettier: ^2.7.0
       rimraf: ^3.0.2
@@ -71,6 +72,7 @@ importers:
       magic-string: 0.26.7
       mlly: 1.0.0
       pathe: 1.0.0
+      pkg-types: 1.0.1
       ufo: 1.0.0
       unplugin: 1.0.0
       vue-i18n: 9.3.0-beta.7_vue@3.2.45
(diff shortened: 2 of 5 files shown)


rchl:
This seems absolutely horrible in my opinion and I don't think anyone should ever need to do such things.

Sorry that I don't have better suggestion since I haven't looked into the issue but I would hope there would be some generic solution that doesn't require package-manager-specific handling.

kazupon:
I know that this solution is terrible too. 
I really struggle with different hoist patterns for different package managers...

rchl:
I don't have understanding on how it works in Nuxt 3 but in Nuxt 2 the issue was only with dependencies used in plugins since plugins were added in `.nuxt/...` where it might not resolve the same way as when importing from the module's package itself. So to fix those cases it was enough to create an alias during module execution and change plugins to import from that alias instead.

Is this is not enough in Nuxt 3 anymore?

And why would we need different alias resolving logic for different package managers? Resolving using "import" or "require" should in theory work more or less the same during module execution.

I know you are busy @danielroe and @pi0 but if you have some take on this then please share.

PR #1786, opened 2023-01-08
fix: don't expose full project path throuh `langDir` option

kazupon:
### ❓ Type of change

### 📚 Description

From e-mail:

An absolute path to the language directory may be exposed in build artifacts if you specify `langDir` option.

Nuxt-i18n expands `langDir` during build process.
https://github.com/nuxt-modules/i18n/blob/c4f7465a289b76d6985b9ac65dfb5908d36daff4/src/index.js#L34

#### Steps to reproduce

```sh
yarn create nuxt-app my-nuxt-app
cd my-nuxt-app
yarn add @nuxtjs/i18n
# After this, you have to modify nuxt.config.ts to use `langDir` option.
yarn run generate
grep -r "${PWD}" dist/ 
# Then the path which contains your current directory will be outputted.
```

#### Affected versions

this behavior is introduced in #1084
So, versions from v6.20.5 to v7.3.0 may be affected.

https://github.com/nuxt-modules/i18n/pull/1084/files
https://github.com/nuxt-modules/i18n/compare/v6.20.4...v6.20.5

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "@babel/traverse": "^7.18.10",
     "@intlify/vue-i18n-extensions": "^1.0.2",
     "@intlify/vue-i18n-loader": "^1.1.0",
-    "@nuxt/utils": "2.15.8",
+    "@nuxt/utils": "2.x",
     "cookie": "^0.5.0",
     "devalue": "^2.0.1",
     "is-https": "^4.0.0",
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
     "@babel/traverse": "^7.18.10",
     "@intlify/vue-i18n-extensions": "^1.0.2",
     "@intlify/vue-i18n-loader": "^1.1.0",
+    "@nuxt/utils": "2.15.8",
     "cookie": "^0.5.0",
     "devalue": "^2.0.1",
     "is-https": "^4.0.0",
--- a/src/index.js
+++ b/src/index.js
@@ -27,13 +27,14 @@ export default async function (moduleOptions) {
   /** @type {Record<string, string>} */
   const localeFileMap = {}
 
-  /**
-   * Helper for `langDir` resources resolving
-   * @param {any[]} args
-   */
-  const relativeToBuild = (...args) => relativeTo(this.options.buildDir, ...args)
-
   if (options.langDir) {
+    /**
+     * Helper for `langDir` resources resolving
+     * @param {any[]} args
+     */
+    const relativeToBuild = (...args) => relativeTo(this.options.buildDir, ...args)
+    const resolvedLangDir = this.nuxt.resolver.resolveAlias(options.langDir)
+
     if (!options.locales.length || typeof options.locales[0] === 'string') {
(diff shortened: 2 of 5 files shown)


kazupon:
@rchl  /cc @danielroe
this is a code fix for Nuxt2 (before nuxtjs/i18n v7.3), so please review it. 🙏

kazupon:
if we need to provide this fix for `v6.x`, I can do it with another PR.

package.json
```diff
@@ -102,6 +102,7 @@
     "@babel/traverse": "^7.18.10",
     "@intlify/vue-i18n-extensions": "^1.0.2",
     "@intlify/vue-i18n-loader": "^1.1.0",
+    "@nuxt/utils": "2.15.8",
```
rchl: Using range should be better so that we don't lock it to this specific version if new version comes out.

Also I wonder whether we should allow all older versions too because I feel like using 2.15.8 could cause issues if someone is using  2.14.0, for example. So we could use a range like `2.x`

src/index.js
```diff
@@ -31,7 +42,12 @@ export default async function (moduleOptions) {
         throw new Error(formatMessage(`All locales must be objects and have the "file" property set when using "langDir".\nFound none in:\n${JSON.stringify(locale, null, 2)}.`))
       }
     }
-    options.langDir = this.nuxt.resolver.resolveAlias(options.langDir)
+    const resolvedLangDir = this.nuxt.resolver.resolveAlias(options.langDir)
+    for (const locale of options.locales) {
+      if (typeof locale !== 'string' && locale.file) {
+        localeFileMap[String(locale.file)] = relativeToBuild(resolvedLangDir, locale.file)
+      }
+    }
```
rchl: Not that important but could do with less duplication and better variable scoping with something like:


```js
  /** @type {Record<string, string>} */
  const localeFileMap = {}

  if (options.langDir) {
    /** @param {any[]} args */
    const relativeToBuild = (...args) => relativeTo(this.options.buildDir, ...args)
    const resolvedLangDir = this.nuxt.resolver.resolveAlias(options.langDir)

    if (!options.locales.length || typeof options.locales[0] === 'string') {
      throw new Error(formatMessage('When using the "langDir" option the "locales" must be a list of objects.'))
    }
    for (const locale of options.locales) {
      if (typeof (locale) === 'string' || !locale.file) {
        throw new Error(formatMessage(`All locales must be objects and have the "file" property set when using "langDir".\nFound none in:\n${JSON.stringify(locale, null, 2)}.`))
      }
      localeFileMap[String(locale.file)] = relativeToBuild(resolvedLangDir, locale.file)
    }
  }
```

kazupon:
@rchl 
Thank you for your review!
I've just updated the code of your review comments.

rchl (approved):

rchl:
Released in https://github.com/nuxt-modules/i18n/releases/tag/v7.3.1

kazupon:
Thank you very much! ❤️

PR #1801, opened 2023-01-19
feat: extend lazy loading

### LINKED ISSUE #1797: extend locale lazy-loading
## describe
Currently, lazy-loading loads the locale messages specified in the i18n.locales file.

BCP47 supports language tags such as language - script - region - variant - extension - privateuse.
https://en.wikipedia.org/wiki/IETF_language_tag

If you define locale messages as resources in a file according to that specification, there is a possibility of duplication.
For example, `en-US` and `en-GB` have the same language, just different regions.
In that case, most resources would be the same.

For convenience in managing resources for locale messages, it would be useful to have a mechanism to define resources only for the difference of region and variant and to be able to load them lazily.

## usecase projects
https://github.com/elk-zone/elk/pull/1172

## more details for this feature request
https://discord.com/channels/1044887051155292200/1045783579520684062/1063952424764719154

### Additional information

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution) (The contribution guideline of nuxt-modules/i18n is compliant with Nuxt too).
- [X] Check existing [discussions](https://github.com/nuxt-modules/i18n/discussions) and [issues](https://github.com/nuxt/framework/issues).

kazupon:
resolve #1797

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/8.lazy-load-translations.md
+++ b/docs/content/2.guide/8.lazy-load-translations.md
@@ -75,7 +75,7 @@ The `files` can load lazily multiple files.
 
 This is useful because it is efficient to manage multiple files that define only differences without duplicating locale messages.
 
-For example, let’s think for the case of Spanish language supporting. There are **79 Spanish-speaking countries** according to [wikipedia](https://en.wikipedia.org/wiki/Spanish_language)!
+For example, let’s think for the case of Spanish language supporting. According to [wikipedia](https://en.wikipedia.org/wiki/Spanish_language), there are **over 15 countries** where Spanish is spoken as an official language!
 
 If these countries use all using `file`, that’s so hard to maintain because of the duplication of locale messages for each contry.
 
@@ -88,7 +88,7 @@ nuxt-project/
 ├── lang/
 │   ├── es.json     # locale messages for common Spanish
 │   ├── es-AR.json  # locale messages for Argentina
-│   ├── es-UY.json  # locale messages for for Uruguay
+│   ├── es-UY.json  # locale messages for Uruguay
 │   ├── es-US.json  # locale messages for Estados Unidos
 |   ...             # other contries ...
 ├── nuxt.config.ts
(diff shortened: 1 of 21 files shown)


docs/content/2.guide/8.lazy-load-translations.md
```diff
@@ -67,3 +68,86 @@ export default defineNuxtConfig({
...
 
 ::
+
+## Multiple files lazy loading
+
+The `files` can load lazily multiple files.
+
+This is useful because it is efficient to manage multiple files that define only differences without duplicating locale messages.
+
+For example, let’s think for the case of Spanish language supporting. There are **79 Spanish-speaking countries** according to [wikipedia](https://en.wikipedia.org/wiki/Spanish_language)!
```
userquin: 19 countries listed on ISO (official): only these https://github.com/elk-zone/elk/pull/1172/files#diff-fb2fd6012fe1954c1021496a3bc2260acb3878612e64f1e0dbdbf30862bbf8f8R93

There are also a few countries on what spanish was official but no rn: Filipinas for example

https://es.wikipedia.org/wiki/Anexo:Pa%C3%ADses_donde_el_espa%C3%B1ol_es_idioma_oficial

docs/content/2.guide/8.lazy-load-translations.md
```diff
@@ -67,3 +68,86 @@ export default defineNuxtConfig({
...
+})
+```
+
+Attention the `files`, so the above configuration specifies multiple file names.
+
+@nuxtjs/i18n will load locale messages with lazy loading in the order of the array specified in `files`. It then overrides the locale messages in the order in which they were loaded.
+
+In the above `es-AR` example, it has `es.json` and `es-AR.json` into `files`. In this case, @nuxtjs/i18n lazy-load `es.json`, and the it lazy-load `es-AR.json` and overrides for `es.json` locale messages.
+
+In the example above, only two files are defined for `files`, ofcourse you can specify more files over 2 files. In that case, the files will be loaded and override in array order too.
```
userquin: wording: `of course` (space)

kazupon:
@userquin 
Thanks! I've updated docs from your feedback!

docs/content/2.guide/8.lazy-load-translations.md
userquin: ~~Still 79, it should be 19, read the article again, these are estimated people who speak Spanish in those countries: in Spain there are people who speak Japanese but in Spain Japanese is not official, you don't have to count Spain as countries  speaking japanese ;)~~

Check `Geographical distribution`, it just mention 20 countries, you're using `Spanish speakers by country` entry.

userquin: upps (updated to 15), why showing changes shows previouws version?

Issue #1859, opened 2023-02-15
`redirectOn` is not work on `prefix_and_default`


kazupon:
> Seems to work on the root page ("/") or with strategy no_prefix, but NOT on sub pages with `prefix_and_default`, even though redirectOn is set to "all" or "no prefix".
Somehow `detectBrowserLanguage` gets called twice (once through the plugin init and once in the middleware), and the second time it just returns no language because both checks ( https://github.com/nuxt-modules/i18n/blob/next/src/runtime/internal.ts#L407 and https://github.com/nuxt-modules/i18n/blob/next/src/runtime/internal.ts#L411) are false.

Reproduction: https://stackblitz.com/edit/github-agwtyv-1bsafu

_Originally posted by @Alex--C in https://github.com/nuxt-modules/i18n/issues/1632#issuecomment-1419531146_

borghol:
@kazupon I believe this issue can be closed

kazupon:
Oh, Oops.
I'll close this issue.
I appreciate your contribution!

Thanks!

PR #1883, opened 2023-02-22
fix: alwaysRedirect does not work with redirectOn 'all' and 'no prefix' 

borghol:
### 🔗 Linked issue

#1859 

### ❓ Type of change

### 📚 Description

When the strategy is set to anything except 'no_prefix' and with alwaysRedirect enabled, redirection only happens on the root path. This change modifies the check when alwaysRedirect is set to check redirectOn strategy before choosing to redirect.

Previously, going to a page like '/about' would not redirect to '/fr/about' because it is not the root. This fix will allow us to redirect correctly so that '/about' and '/ja/about' will redirect to '/fr/about' when using 'all', and redirect '/about' to '/fr/about' when using 'no prefix'

### 📝 Checklist

- [X] I have linked an issue or discussion.


borghol:
I will look into why these tests failed and reopen it afterwards

Issue #1889, opened 2023-02-24
Can't navigate to page with `defineI18nRoute(false)`


lukaszflorczak:
### Environment

- Operating System: `Darwin`
- Node Version:     `v19.6.1`
- Nuxt Version:     `3.2.2`
- Nitro Version:    `2.2.2`
- Package Manager:  `pnpm@7.27.1`
- Builder:          `vite`
- User Config:      `modules`, `i18n`
- Runtime Modules:  `@nuxtjs/i18n@8.0.0-beta.10`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/nuxt-starter-c377bi?file=app.vue

### Describe the bug

If the page component has `defineI18nRoute(false)` it is not possible to navigate to this page on the client side.

I would expect that I can visit that page (URL has no lang prefix), and after returning to another (localized), I will still have the original language set.

kazupon:
This issue has been fixed.
You can install the edge channel version.
Thanks!

kazupon:
re-open, I've reverted this PR.
https://github.com/nuxt-modules/i18n/pull/2153

lukaszflorczak:
Hi @kazupon! 

What is the current situation of this issue? Unfortunately, this bug disqualifies beta.13 for me 😔

kazupon:
Oops. I'll work on this issue!

saifobeidat:
this is happening with me as well,. did you find any workaround?

or is there any version that doesn't have this issue?

lukaszflorczak:
@saifobeidat At this moment I use `8.0.0-beta.12`

kazupon:
@lukaszflorczak 
Hi!
I fixed this issue, so you can try to install edge channel version.
Happy coding!

PR #1902, opened 2023-03-02
fix: defineI18nRoute do not work client-side

borghol:
### 🔗 Linked issue
#1889 

### ❓ Type of change

### 📚 Description

When attempting to check redirection, the detectRedirect function matches a non-defined i18n route back to the root path. 

The detectRedirect function was modified to detect whether the route matched by the router is an i18n route. If it is, we will continue with the logic as-is. If it is not, we will skip redirection.

### 📝 Checklist

- [X] I have linked an issue or discussion.


kazupon:
That's excellent work! ❤️ 
Thanks!

kazupon (approved):

PR #1919, opened 2023-03-09
refactor!: functions in the config to be used through hooks

ineshbose:
### ❓ Type of change

### 📚 Description

As a step towards stability, the module would also aim for stable code generation and handling that occurs to interact successfully with Nuxt. It's realised that passing functions through the module configuration may not be ideal, and hence this PR aims to use runtime hooks provided by Nuxt for such purposes.

### 📝 Checklist


src/runtime/plugins/i18n.ts
```diff
@@ -224,8 +224,12 @@ export default defineNuxtPlugin(async nuxt => {
           _getLocaleCookie(nuxt.ssrContext, { ...nuxtI18nOptions.detectBrowserLanguage, localeCodes })
         composer.setLocaleCookie = (locale: string) =>
           _setLocaleCookie(locale, nuxt.ssrContext, nuxtI18nOptions.detectBrowserLanguage || undefined)
-        composer.onBeforeLanguageSwitch = nuxtI18nOptions.onBeforeLanguageSwitch
-        composer.onLanguageSwitched = nuxtI18nOptions.onLanguageSwitched
+
+        // @ts-ignore
+        composer.onBeforeLanguageSwitch = (o, n, i, ctx) => nuxt.callHook('i18n:before-lang-switch', o, n, i, ctx)
```
kazupon: if it's possible, I hope you would resolve type. 🙏

src/module.ts
```diff
@@ -284,5 +284,12 @@ declare module '@nuxt/schema' {
 
   interface NuxtHooks {
     'i18n:extend-messages': (messages: LocaleMessages<DefineLocaleMessage>[], localeCodes: string[]) => Promise<void>
+    'i18n:before-lang-switch': <Context = unknown>(
+      oldLocale: string,
+      newLocale: string,
+      initialSetup: boolean,
+      context: Context
+    ) => string | void
+    'i18n:lang-switched': (oldLocale: string, newLocale: string) => void
```
kazupon: We need to edit docs that we are now using Nuxt hooks instead of `onBeforeLanguageSwitch` and `onLanguageSwitched`.

src/runtime/plugins/i18n.ts
ineshbose: I do want to propose changing the parameters for these functions to be an object with the values in properties instead, so it would be

```ts
type Props<Context = unknown> = {
  oldLocale: string,
  newLocale: string,
  initialSetup: boolean,
  context: Context
}

export type BeforeLanguageSwitchHandler = <C = unknown>(prop: Props<C>) => string | void

// then..
composer.onBeforeLanguageSwitch = (p) => nuxt.callHook('i18n:before-lang-switch', p)
```

The function type should not be an issue, however passing `'i18n:before-lang-switch'` becomes a warning since the built-in hooks doesn't have that option and I have seen other modules use ts-ignore for it. Perhaps I need to dig deeper into docs and see what can be done!

src/module.ts
ineshbose: Yes. I would like to update the docs once we have the implementation ready (keep in mind that the PR is still a draft)!

src/runtime/plugins/i18n.ts
kazupon: Thanks! I've understood your proposal.

>I do want to propose changing the parameters for these functions to be an object with the values in properties instead

Indeed, It makes sense to pass objects instead of params. With objects, it is less likely to break the function's argument signature if we add something we want to pass to function.

we want to rename the nuxt hooks defined for the new i18n module as follows:

- `'i18n:before-lang-switch'` -> `'i18n:beforeLocaleSwitch'`
- `'i18n:locale-switched'` -> `'i18n:localeSwitched'`

If there are naming guidelines for nuxt hooks, they should be followed.

The following are lifecycle hooks:
https://nuxt.com/docs/api/advanced/hooks#lifecycle-hooks

ineshbose: In that case, should I rename the existing `i18n:extend-messages` hook to be camel case as well?

ineshbose:
Hi @kazupon 

I realised that this hook would be _runtime_ instead of build time like `i18n:extend-messages`, so I've made changes accordingly, but will require your expertise at this point to thoroughly review and test! (I tried to change `callbacks.spec.ts` but I believe I'm unable to get the other snapshots passing)

src/module.ts
```diff
@@ -286,3 +289,15 @@ declare module '@nuxt/schema' {
...
 }
+
+declare module '#app' {
+  interface RuntimeNuxtHooks {
+    'i18n:beforeLocaleSwitch': <Context = unknown>(
+      params: LocaleSwitch & {
+        initialSetup: boolean
+        context: Context
+      }
+    ) => MaybePromise<void>
```
kazupon, line 300:
```suggestion
    ) => MaybePromise<string | void>
```

kazupon: See docs:
https://v8.i18n.nuxtjs.org/guide/callbacks#onbeforelanguageswitch
>Returns: string or nothing

src/runtime/types.d.ts
```diff
@@ -26,15 +26,15 @@ type BeforeLanguageSwitchHandler = (
   newLocale: string,
   initialSetup: boolean,
   context: NuxtApp
-) => string | void
+) => Promise<unknown>
```
kazupon:
```suggestion
) => Promise<string | void>
```

src/runtime/types.d.ts
```diff
@@ -26,15 +26,15 @@ type BeforeLanguageSwitchHandler = (
...
 
 /**
  * Called after the app's locale is switched.
  *
  * @param oldLocale - The app's locale before the switch
  * @param newLocale - The app's locale after the switch.
  */
-type LanguageSwitchedHandler = (oldLocale: string, newLocale: string) => void
+type LanguageSwitchedHandler = (oldLocale: string, newLocale: string) => Promise<unknown>
```
kazupon:
```suggestion
type LanguageSwitchedHandler = (oldLocale: string, newLocale: string) => Promise<void>
```

src/runtime/types.d.ts
ineshbose: I'm not sure if the string can be returned from the hook. There would be a minor workaround required from developers to assign the locale instead of returning it.

src/types.ts
```diff
@@ -41,14 +41,12 @@ export type BeforeLanguageSwitchHandler = <Context = unknown>(
   newLocale: string,
   initialSetup: boolean,
   context: Context
-) => string | void
+) => Promise<unknown>
```
kazupon: `BeforeLanguageSwitchHandler`  should be removed, so because it's swtiched to nuxt hooks, 
It is no longer necessary to configure in nuxt.config.ts.

src/types.ts
```diff
@@ -41,14 +41,12 @@ export type BeforeLanguageSwitchHandler = <Context = unknown>(
...
 
-export type LanguageSwitchedHandler = (oldLocale: string, newLocale: string) => void
+export type LanguageSwitchedHandler = (oldLocale: string, newLocale: string) => Promise<unknown>
```
kazupon: This should also be removed.
Here is [same reason](https://github.com/nuxt-modules/i18n/pull/1919/files#r1133094629
LanguageSwitchedHandler) too.

kazupon:
And these exports on `options.d.ts` should be removed too.
https://github.com/nuxt-modules/i18n/blob/next/src/options.d.ts#L40-L42

src/runtime/plugins/i18n.ts
kazupon: Yeah!
The topic differs from this PR, so we should handle it with another PR.

ineshbose:
What do you think about the logic of the change - any underlying implementation that may have been overlooked?

src/module.ts
ineshbose: ~~Instead of return, it may need to be something like `nuxtApp.$i18n.locale = 'en'`~~

kazupon:
I think that your logic is sound.
The `callbacks.spec.ts` needs to be changed to nuxt hooks, but what you are testing should be the same.

kazupon:
One more request 🙏 
Changing to nuxt hooks is breaking change for nuxt2 apps using v7.
So, we need to explain it in the migration docs.
https://v8.i18n.nuxtjs.org/guide/migrating

ineshbose:
Just working on the changes and making sure the tests are passing.

What is your opinion on `beforeLocaleSwitch` not being able to return string anymore?

This is a kind of workaround:

```ts
import { defineNuxtPlugin } from '#imports'

export default defineNuxtPlugin(nuxtApp => {
  nuxtApp.hook('i18n:beforeLocaleSwitch', async ({ oldLocale, newLocale, initialSetup }) => {
    console.log('onBeforeLanguageSwitch', oldLocale, newLocale, initialSetup)

    if (newLocale === 'en') {
      await nuxtApp.$i18n.setLocale('fr')
    }
  })
})

```

However, the function is async, and `callHook` is also async so we need to await it somehow in the composer.

Edit: I'm able to await the function here, but without a return value for localeOverride, it may not work. Let me investigate further.

```ts
  const localeOverride = await onBeforeLanguageSwitch(i18n, oldLocale, newLocale, initial, context)
  const localeCodes = getLocaleCodes(i18n)
  if (localeOverride && localeCodes && localeCodes.includes(localeOverride)) {
    if (localeOverride === oldLocale) {
      return [ret, oldLocale]
    }
    newLocale = localeOverride
  }
```


### Returning locale works! Just needed to await it.

src/module.ts
ineshbose: This won't work with `callHook` it seems!

src/runtime/utils.ts
```diff
@@ -201,11 +201,11 @@ export async function loadAndSetLocale<Context extends NuxtApp = NuxtApp>(
 
   // set the locale
   if (useCookie) {
-    setCookieLocale(i18n, newLocale)
+    await setCookieLocale(i18n, newLocale)
```
kazupon: I’m not sure if there is any js engine overhead in running as `await` on a synchronous function, but since `setCookieLocale` and `setLocale` are synchronous functions, I think that we don't need to `await`.

src/options.d.ts
```diff
@@ -37,7 +37,5 @@ export {
   NuxtI18nInternalOptions,
   DetectBrowserLanguageOptions,
   RootRedirectOptions,
-  LanguageSwitchedHandler,
-  BeforeLanguageSwitchHandler,
   LanguageSwitchedHandler
```
kazupon: we need to remove this line (`LanguageSwitchedHandler`)
I don't remember why I exported the same type. And typescript doesn't warn us. 😅

src/runtime/utils.ts
ineshbose: I think there's no harm in it but rather safer - but I've removed it!

kazupon:
@ineshbose 
Thanks for all your hard work!
I'll merge this PR!

kazupon (approved):

PR #1926, opened 2023-03-12
refactor: using logger from `@nuxt/kit`

ineshbose:
### ❓ Type of change

### 📝 Checklist


ineshbose:
Hi @kazupon, would you be able to provide advice on eliminating `createDebug`? I have no experience with it, but could be very good to remove it as a dependency if the logger is suitable.

kazupon:
@ineshbose 
I know that the logger provided by nuxt kit is consola.

consola can be filtered by log level using the environment variable `CONSOLA_LEVEL`.
https://github.com/unjs/consola#level

This is useful because it gives you control over features not available in debug module.

But, the ability to control log output with tag was not clear from my reading of the documentation.

The reason I use debug is that I can control consola's tag in detail using the `DEBUG` environment variable.
https://github.com/debug-js/debug

so this reason, I like to use it in my vue-i18n, intlify project and nuxt i18n module!
It would be nice to have the same functionality in consola.

@Atinux @pi0 
Sorry to ping you and ask, as it is obvious if you read the consola code 🙇‍♂️ 
Is it possible to do something like filter by debug namespace in consola?

atinux:
It would be very nice to add such feature in consola actually, could you open an issue there?

ineshbose:
I'll make a temporary function around it for now.

@kazupon I'll wait to progress on this after #1938 has been merged as I see some log statements there along with a potential logger conflict. And then I'll move to #1943

ineshbose:
Closing. Will reopen at a convenient time (unless someone picks it as a good first issue 🙂).

PR #1929, opened 2023-03-13
fix: defineI18nRoute not working outside setup sugar syntax pages

### LINKED ISSUE #1928: defineI18nRoute macro does not work for pages without setup syntax
### Environment

- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.2.3`
- Nitro Version:    `2.2.3`
- Package Manager:  `npm@7.17.0`
- Builder:          `vite`
- User Config:      `modules`, `vite`, `i18n`
- Runtime Modules:  `normalizedModule()`, `@nuxtjs/i18n@8.0.0-beta.10`

### Reproduction

This can be reproduced using the playground project in this repository, see https://stackblitz.com/edit/github-zvo7zt?file=pages%2Fabout%2Findex.vue

Navigate to the about page and notice the other languages lack translations while they should be configured with:
```ts
defineI18nRoute({
  paths: {
    en: '/about-us',
    fr: '/a-propos',
    ja: '/about-ja'
  }
})
```

### Describe the bug

This is probably the same issue as https://github.com/nuxt-modules/i18n/issues/1852

It looks like the macro is not handled when not using `<script setup>` syntax. I will try to fix it and submit a PR!

### Additional context

_No response_

### Logs

_No response_

BobbieGoede:
### 🔗 Linked issue
#1928 

### ❓ Type of change

### 📚 Description
Resolves #1928

It looks like `defineI18nRoute` was only checked  in components/pages using the setup sugar syntax, maybe as an optimization? I changed the logic to use work with either approaches.

In case it was an optimization I have added a simple check whether the content of the file includes `defineI18nRoute` before continuing, it may be too simple so I can remove that if needed.

### 📝 Checklist

### DIFF
--- a/src/pages.ts
+++ b/src/pages.ts
@@ -245,16 +245,19 @@ function readComponent(target: string) {
   try {
     const content = fs.readFileSync(target, 'utf8').toString()
     const { descriptor } = parseSFC(content)
-    if (!descriptor.scriptSetup) {
+
+    if (!content.includes('defineI18nRoute')) {
       return options
     }
+
     const desc = compileScript(descriptor, { id: target })
-    const { scriptSetupAst } = desc
+    const { scriptSetupAst, scriptAst } = desc
 
     let extract = ''
-    if (scriptSetupAst) {
+    const genericSetupAst = scriptSetupAst || scriptAst
+    if (genericSetupAst) {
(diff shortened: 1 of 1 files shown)


kazupon (approved):
Thank you so much always!
LGTM!

PR #1948, opened 2023-03-22
refactor!: read vue-i18n options from config path

ineshbose:
> **Note**: Possibly being chased/continued in #1973 

### ❓ Type of change

### 📚 Description

Options would need to be passed from a separate file that will be imported by the plugin. This idea is inspired from looking at other modules that require runtime configuration. Please feel free to discuss over Discord about this change.

**WORK IN PROGRESS**

### 📝 Checklist


ineshbose:
@ kazupon, please feel free to share thoughts at your convenience. I expect the change to be slightly big in certain areas but I'll only finalise things after the previously proposed PRs have been tackled.

Btw, should the default config file be named `i18n.config` you think? (it follows `*.config` convention, is shorter than `vue-i18n.{config,options}` and it's also what I'm using in my personal Nuxt project) It would be customisable regardless - your call 🙂

kazupon:
>should the default config file be named i18n.config you think? 

Yeah! your idea sounds good.

src/types.ts
```diff
@@ -52,7 +52,9 @@ export type NuxtI18nOptions<Context = unknown> = {
   skipSettingLocaleOnNavigate?: boolean
   // sortRoutes?: boolean
   strategy?: Strategies
-  vueI18n?: I18nOptions | string
+  vueI18n?: {
+    configFile?: string
+  }
```
kazupon, line 66: I hope the i18n options keep `I18nOptions | string` because it would break compatibility with v7.

ineshbose: It would need to be `string` for this change. Sorry if I'm misunderstanding but why is compatibility with v7 important (since this is another major version)?

kazupon: Indeed, v8 is a major version, so breaking changes are inevitable.
Upgrading to a major version for a large application that has already been built is a huge effort.

Some frameworks (e.g. larave, rails, and react) have been released so that major versions do not cause too much work for the user's application. They are releasing features supported in previous versions in working order, just as Vue3 supports the Options API.

I have once again thought about how much breaking change would detract from the migration experience for the user.

In this PR, it is a change in the configuration interface, so the affected code point is local.
Changing the vueI18n option would not be much of a problem.
Of course, we need to write a migration guide in the documentation.

ineshbose: Absolutely - it would be great to have as less breaking changes as possible.

Similar to #1919 also moving some fragile configuration options to other solutions that Nuxt provides for stability, I'm aiming to do the same here. This may have been a small weak-point in v7, so having this solved in v8 would be great. The migration would indeed be documented.

ineshbose:
PR would also possibly tackle this-

https://github.com/nuxt-modules/i18n/blob/7943ca10a11ed07dc60b4bf3aa9896f8069ecd0f/TODO.md?plain=1#L9

kazupon:
Could you prioritize this PR over other tasks? 🙏 
I hope to release an RC version soon, and I want this to be the last opportunity involving breaking changes before that release.

docs/content/1.getting-started/2.basic-usage.md
```diff
@@ -8,34 +8,37 @@ The basics to get started with the Nuxt i18n module is to translate with Vue I18
...
+    },
+    fr: {
+      welcome: 'Bienvenue'
     }
   }
 })
 ```
 
-The `vueI18n` option is the same as `createI18n` function option of Vue I18n. `vueI18n` option is passed to the `createI18n` function via the nuxt plugin of this module internally.
+The `defineI18nConfig` function is the same as `createI18n` function option of Vue I18n. The configuration is passed to the `createI18n` function via the nuxt plugin (runtime) of this module internally.
```
kazupon: I think we should name `defineIVue18nConfig`, not `defineI18nConfig`, because that is coniguration for`vueI18n` option. And some people might misunderstand it as a nuxt i18n setting.

docs/content/3.options/1.vue-i18n.md
```diff
@@ -6,26 +6,37 @@ Related Vue I18n options.
...
+
+Configuration for `createI18n` can be passed using a configuration file. By default, the module will scan for a `i18n.config.{js,mjs,ts}` if nothing is specified.
+
+```ts {}[nuxt.config.ts]
+export default defineNuxtConfig({
+  modules: [
+    '@nuxtjs/i18n'
+  ],
+  i18n: {
+    vueI18n: { configPath: './nuxt-i18n.js' } // custom path example
```
kazupon, line 22: ~~The `vueI18n` option may no longer be optional. It's recommended to autoload `i18n.config.ts` if we have a config that conforms to the convention. If we have a nonconventional config file name, we can specify it in the `vueI18n` option!~~

kazupon: ah, sorry, Never mind this comment.
I was misreading. 😅

kazupon:
In additions, we also need to fix the basic usage docs.
https://v8.i18n.nuxtjs.org/getting-started/basic-usage#translate-with-vue-i18n

ineshbose:
Hi @kazupon 

I wanted to formalise the context to this PR with complete transparency so that everyone is on the same page!

This module provides enabling [`vue-i18n`](https://vue-i18n.intlify.dev/) with Nuxt 3 and SSR. On further deep-diving into the code of this repository and intlify packages, I realised that there is a lot of code-generation work happening (from JavaScript to JavaScript) which has a reputation of not being the most stable or reliable, so I'm looking for ways around it. Since I'm not an expert in i18n solutions, I cannot comment on the requirement for this. Nuxt 3 also involves a level of code generation through `addTemplate`, etc, and I have been using these functionalities for a while now to gain a better understanding; I've learnt that code generation should be minimum but also functions shouldn't be serialized (hence #1919 was also made). `vue-i18n` is excellent for internationalization and it provides [modifiers](https://vue-i18n.intlify.dev/guide/essentials/syntax.html#custom-modifiers) that are passed as functions. From previous versions of this module, `vueI18n` is the configuration option for it, and then it gets serialized using code generation - but that has the limitations of manual code generation including understandability of code, efficiency and stability.

Based on Nuxt's official guide on a project's [contexts](https://nuxt.com/docs/guide/going-further/internals#runtime-context-vs-build-context) - _"they are not supposed to share state, code, or context"_ (also why [`nuxt.config.ts`](https://nuxt.com/docs/guide/directory-structure/nuxt.config) and [`app.config.ts`](https://nuxt.com/docs/guide/directory-structure/app-config) are different and declared differently, why [plugins](https://nuxt.com/docs/guide/directory-structure/plugins) and [modules](https://nuxt.com/docs/guide/directory-structure/modules) are different and declared differently, why [app hooks](https://nuxt.com/docs/api/advanced/hooks#app-hooks-runtime) and [nuxt hooks](https://nuxt.com/docs/api/advanced/hooks#nuxt-hooks-build-time) are different and [declared differently](https://nuxt.com/docs/guide/going-further/hooks#lifecycle-hooks)) - since `vue-i18n` is on the runtime context as a plugin, the configuration shouldn't be in `nuxt.config.ts` (build context) so that would greatly help with ensuring that developers are aware of the different contexts (as Nuxt aims to help with) and more importantly - code-generation since we don't have to take the build-time configuration to build/serialize it into runtime configuration as discouraged by the [guide](https://nuxt.com/docs/guide/going-further/internals#runtime-context-vs-build-context). Instead, since the configuration is now different, in a different file, it could just be imported in runtime. [A great example of this is the FormKit module that only takes in configuration path](https://github.com/formkit/formkit/blob/757a2f2a95f5424e1d922f66a91d6b1af29cab95/packages/nuxt/src/module.ts#L35-L70) to create the [runtime plugin](https://github.com/formkit/formkit/blob/master/packages/nuxt/src/runtime/plugin.mjs) ([docs](https://formkit.com/getting-started/installation#with-vue)) - so that is a point of inspiration.

With all that said, this is a massive change that will need to be tested intensively, and for me, as an amateur, to carry this out since I only have such little understanding of how the module and code generation is working, I will need some great help and great guidance. I can't say what steps to take forward, but I'm looking to hear from the lead developer (aka you 😄) if have some plans or feedback on how this can be facilitated or just cancelled (which I'm open to, but I believe it would continue with the code debt). Let me know - and we can continue a close discussion over Discord if required. 🙂

kazupon:
Hi! @ineshbose 
Thank you for your detailed and clear explanation!!

I was understanding the current problems with the `vueI18n` option and that it does not follow the Nuxt development conventions 😅 
`vueI18n` option is not code generated by `gen.ts`, but by using `addPluginTemplate` and import in that template.
It's better to use `addPluginTemplate`, as it will be safely handled by the Nuxt plugin.

The implementation using `addPluginTemplate` needs huge refactoring in `runtime/i18n.ts` and `gen.ts`.
I think that this refactoring would delay the release.

So, I have considered whether there is any compromise for the RC release.

Fortunately, `vueI18n` option, when a path is specified as a string, the code generated in `i18n.options.mjs` reads the file specified in the path with dynamic import.

The loaded file is processed by the bundler (vite), so that even if functions such as modifiers are set, they can be safely processed.

`i18n.config` can be solved with `module.ts`, so all you have to do is specify the path, and the impact cost is low.

If `vueI18n` option is set with Object, it will output a warning to set it with `i18n.config`, when `nuxi dev`.

In the RC release, we can accept configuration by Object, but in the v8 official release or RC patch version release, we will discontinue that support and `vueI18n` option will only accept path configuration.

That way, I believe we can make changes to `vueI18n` option progressive without forcing the community to wait for RC release.
We can refactor for `vueI18n` option using `addPluginTemplate` until v8 official release.

What do you think?

Issue #1942, opened 2023-03-18
Rename `i18n:extend-messages` to `i18n:extendMessages`


kazupon:
In that case, should I rename the existing `i18n:extend-messages` hook to be camel case as well?

_Originally posted by @ineshbose in https://github.com/nuxt-modules/i18n/pull/1919#discussion_r1133077515_

ineshbose:
Brilliant issue to get contributing to this repo! 😄

kazupon:
We might be able to deplicate i18n:extend-message. Because I think we can do the same thing with the recently supported Layers. /cc @BobbieGoede

https://v8.i18n.nuxtjs.org/guide/layers

The reason for deprecated is that this hook pre-compiles the messages passed by vue-i18n at build-time, and generates code in i18n.options.mjs in the template.
This is javascript to javascript and not safe.
Also, if large messages are set in this hook, i18n.options.mjs will become large, resulting in a performance issue.

reference discord link:
https://discord.com/channels/1044887051155292200/1063952424764719154/1064230457098645645

BobbieGoede:
@kazupon that should be possible, I can look into adapting for modules. Are there any advantages over extending messages with objects instead of files with the current approach?

Is there a preferred way of registering/discovering modules that want to provide messages/locales? Using a hook I figure it could work something like this for modules:
```ts

export default defineNuxtModule({
  async setup(options, nuxt) {

    // could potentially be simplified inside a function to handle resolve and hook
    const resolver = createResolver(import.meta.url)
    nuxt.hook('i18n:registerModule', registerI18nModule => {
      registerI18nModule({
        // paths may be more unpredictable compared to layers?
        langDir: resolver.resolve('./locales'),
        locales: [
          {
            code: 'en',
            iso: 'en-US',
            file: 'en.json',
            name: 'English'
          },
          {
            code: 'nl',
            iso: 'nl-NL',
            file: 'nl.json',
            name: 'Nederlands'
          }
       ]
    })
  }
)
```

I'm not sure if `iso` and `name` are relevant for adding messages from modules.

The discord link doesn't work for me so I am not sure what it references 😅

PR #1955, opened 2023-03-27
feat: add basic module locale merging

BobbieGoede:
### 🔗 Linked issue
[#1942](https://github.com/nuxt-modules/i18n/issues/1942#issuecomment-1483898669)

### ❓ Type of change

### 📚 Description
Allows modules to provide locales in a similar way to layers (#1925), as a potential replacement for `i18n:extend-messages` hook.

Current usage is using a hook to register your module with its i18n configuration, similar to configuring lazy-load locales. Other locale configuration have not been implemented yet. If a different approach is preferred please let me know!
```ts
export default defineNuxtModule({
  async setup(options, nuxt) {
    const { resolve } = createResolver(import.meta.url)
    nuxt.hook('i18n:registerModule', registerI18nModule => {
      registerI18nModule({
        langDir: resolve('./locales'),
        locales: [
          {
            code: 'en',
            iso: 'en-US',
            file: 'en.json',
            name: 'English'
          },
          {
            code: 'fr',
            iso: 'fr-FR',
            file: 'fr.json',
            name: 'Francais'
          },
          {
            code: 'nl',
            iso: 'nl-NL',
            file: 'nl.json',
            name: 'Nederlands'
          }
        ]
      })
    })
  }
})
```

This PR is still a work in progress, implementation will be more generic to prevent code duplication as it has in its current state. I will add tests and docs later if it's clear that we want this functionality to be added and the usage is agreed on.

### 📝 Checklist


kazupon:
This is great PR! ❤️ 

We can replace the `i18n:extend-messages` hook.
The problem of `i18n:extend-messages` hook is that at nuxt build-time, object locale is pre-compiled to javascript, so that vue-i18n can handle it. And then pass it to template (`i18n options.mjs`) in `gen.ts`.
This was javascript to javascript and unsafe. And if the locale object was huge, `i18n options.mjs` was huge too.

This PR solves these problems by lazy loading of the locale on runtime builds and Nuxt layers.
We can deprecated the `i18n:extend-messages` hook.

BobbieGoede:
@kazupon I added some documentation as well as a basic test and I haven't touched the `i18n:extend-messages` hook to keep this PR from breaking existing functionality, so deprecation can be done in a separate PR.

The `i18n:registerModule` hook implementation only supports adding lazy-load translations (`locales: LocaleObject[]`), I feel that adding support for strings only (`locales: ['en', 'fr']`) wouldn't make sense as no messages would be added but I can add it if needed.

kazupon:
@BobbieGoede 
Thank you very much!
Great job!

> I haven't touched the i18n:extend-messages hook to keep this PR from breaking existing functionality, so deprecation can be done in a separate PR.

I will write the documentation for ` i18n:extend-messages` deprecation/

>The i18n:registerModule hook implementation only supports adding lazy-load translations (locales: LocaleObject[]), I feel that adding support for strings only (locales: ['en', 'fr']) wouldn't make sense as no messages would be added but I can add it if needed.

Yes! `locales: LocaleObject[]` only is enough! `locales: string[]` makes no sense at all.

PR #2049, opened 2023-05-08
chore(deps): relax upper node version constraint

DamianGlowala:
Inspired by https://github.com/nuxt/nuxt/commit/3133c6fb84affca6854b7c7fa8cfbab9fe2aebe6.

### ❓ Type of change

### 📝 Checklist


package.json
```diff
@@ -152,6 +152,6 @@
     "access": "public"
   },
   "engines": {
-    "node": "^14.16.0 || ^16.11.0 || ^17.0.0 || ^18.0.0 || ^19.0.0"
+    "node": "^14.16.0 || >=16.11.0"
```
kazupon, line 155: I think that node v14 should be dropped since that passed the LTS.
https://github.com/nodejs/release#release-schedule

/cc @danielroe

danielroe: It's up to you. It certainly doesn't need to be supported, but I won't be removing from the Nuxt engines until we merge features that are actively incompatible with Node 14.

kazupon: Thanks!

I noticed that node v16 will be out of support in about 6 months. So it would be more efficient to drop it together with node v16.
I will accept this PR!

kazupon (approved):

PR #2079, opened 2023-05-15
fix: remove fallback routes

### LINKED ISSUE #1288: Unprefixed routes are always added in full static mode when strategy is "prefix"
### Version
@nuxtjs/i18n: 7.0.3
nuxt: 2.15.7

### Nuxt configuration

 - [ ] Applies to a site deployed to a server with a Node backend

```
i18n: {
    strategy: 'prefix',
}
```

### Reproduction Link
[https://codesandbox.io/s/nuxt-nuxtjs-i18n-base-template-forked-ron65?file=/pages/index.vue](https://codesandbox.io/s/nuxt-nuxtjs-i18n-base-template-forked-ron65?file=/pages/index.vue)

### Steps to reproduce
1. Set `target: static` in nuxt.config.js
2. Set `strategy: prefix` in i18n config

### What is Expected?
There should only be routes prefixed with locale codes

There are routes with prefixed locale codes as well as routes with unprefixed locale codes.
As you can see, the reproduction link prints all routes and for each page there is the prefixed version, ie: /en/about /es/about but also the unprefixed version: /about

I'd say that with setting the strategy to "prefix" this is an unexpected behaviour, especially considering it only happens in full static mode.

The reason why this happen is this logic:
https://github.com/nuxt-community/i18n-module/blob/c710ad18d30dc80c9e71502fbf0e722e596952c7/src/core/hooks.js#L13-L16

When full static mode is active or when `nuxt generate` runs, the variable `includeUprefixedFallback` is set to true which forces the route generation to include the unprefixed version of each route.

My question is why is this logic present? I guess there must be a good reason if it's there 😄

dword-design:
### 🔗 Linked issue

Closes #1288

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/pages.ts
+++ b/src/pages.ts
@@ -41,12 +41,6 @@ export function setupPages(
     return !options.differentDomains && DefaultLocalizeRoutesPrefixable(opts)
   }
 
-  let includeUprefixedFallback = nuxt.options.ssr === false
-  nuxt.hook('nitro:init', () => {
-    debug('enable includeUprefixedFallback')
-    includeUprefixedFallback = true
-  })
-
   const pagesDir = nuxt.options.dir && nuxt.options.dir.pages ? nuxt.options.dir.pages : 'pages'
   const srcDir = nuxt.options.srcDir
   const { trailingSlash } = additionalOptions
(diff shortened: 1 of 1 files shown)


kazupon:
@rchl 
Could you review this PR for v7? 🙏 
I'm busy so I can't handling everything? 😅

BobbieGoede:

PR #2108, opened 2023-05-25
feat: add configurable runtime only setting

### LINKED ISSUE #2075: Lazy-Load Dynamic message files.
### Describe the feature

Currently if you use lazy-loading for the translations of a website, they get pre-rendered / pre-compiled at build time which is not really feasible if you have a large/community driven translation project and want to have the translations dynamically update over let's say an API endpoint which provides the string data when you visit the page. Same goes for dynamically being able to add/remove locales (from the locales array) by fetching an API endpoint to find out what languages the website supports. I am not 100% sure if this isn't already possible because when I tried to do this by using a `ts` file to fetch the language it got pre-compiled once I built the project for SSR which I don't want.

### Additional information

- [X] Would you be willing to help implement this feature?

### Final checks

Vanger888:
### 🔗 Linked issue

https://github.com/nuxt-modules/i18n/issues/2075

### ❓ Type of change

### 📚 Description
Added `runtimeOnlyVueI18nBundle` under the experimental flag as described in the linked task. This flag is true by default, so it shouldn't affect current behaviour.
I faced lazy loading limitations, we need to load translations at runtime because they change quite often. I found some workarounds with empty messages at build time and setting dynamic messages at runtime with` i18n.setLocaleMessage.` But with runtime only vue-i18n we can't use any params in translations. I hope this feature helps to solve this limitation.

Partially resolved #2075

### 📝 Checklist

### DIFF
--- a/docs/content/3.options/6.misc.md
+++ b/docs/content/3.options/6.misc.md
@@ -7,7 +7,7 @@ Miscellaneous options.
 ## `experimental`
 
 - type: `object`
-- default: `{ jsTsFormatResource: false }`
+- default: `{ jsTsFormatResource: false, runtimeOnlyVueI18nBundle: true }`
 
 Configure the flag for experimental features of the nuxt i18n module.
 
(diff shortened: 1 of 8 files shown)


kazupon:

Bundling vue-i18n with runtime (i.e. message compiler) is not guaranteed to work.
This is because of CSP limitations and may not work in some user environments.
https://vue-i18n.intlify.dev/guide/advanced/optimization.html#optimization

We can support runtime vue-i18n, but we **need to understand the limitations of CSP**.

We can offer it as **an experimental feature** if many users want it.

Issue #2132, opened 2023-06-03
`redirectOn: 'no prefix'` does not redirect on the root URL


phanect:
### Environment

(Here's Stackblitz' environment information)

- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.5.2
- Nitro Version:    2.4.1
- Package Manager:  npm@7.17.0
- Builder:          vite
- User Config:      ssr, components, modules, i18n
- Runtime Modules:  @nuxtjs/i18n@8.0.0-beta.12
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-2yxmzd

### Describe the bug

I added `strategy: 'prefix'` and `redirectOn: 'no prefix'` to nuxt.config.ts.

nuxt.config.ts

```ts
export default defineNuxtConfig({
  // ...
  i18n: {
    // ...
    strategy: 'prefix',
    // ...
    detectBrowserLanguage: {
      useCookie: true,
      cookieSecure: true,
      alwaysRedirect: true,
      fallbackLocale: 'en',
      redirectOn: 'no prefix',
    },
  },
});
```

If I understand correctly, with this config, if you open / (root URL), it should return 302 and redirect to /en or /ja. (Depending on your `Accept-Language`)
However, when you open / (root URL), it returns 404.

phanect:
This issue might be related to #2131.

phanect:
I guess the following code causes this issue:

https://github.com/nuxt-modules/i18n/blob/8dc25cbd88212563b5f9709740246a4915b8462f/src/runtime/utils.ts#L294-L299

`isI18nRouteDefined(route)` always returns `false` because `route.matched[0]?.meta.nuxtI18n` is always `undefined` on my end. (see the following `isI18nRouteDefined()` function:

https://github.com/nuxt-modules/i18n/blob/8dc25cbd88212563b5f9709740246a4915b8462f/src/runtime/utils.ts#L514-L517

Therefore, `detectRedirect()` returns `redirectPath` (=== empty string (`''`)) and [the following redirection process](https://github.com/nuxt-modules/i18n/blob/8dc25cbd88212563b5f9709740246a4915b8462f/src/runtime/utils.ts#L301-L348) is not executed.

I'm stuck here because I'm unsure how to make `route.matched[0]?.meta.nuxtI18n` a non-undefined value...

kazupon:
Thanks!

You can workaround with `dedefaultLocale` option.

[...]

Issue #1632, opened 2022-11-10
detectBrowserLanguage Parameter configuration does not work


senher:
### Version
@nuxtjs/i18n:^8.0.0-beta.3 
nuxt:3.0.0-rc.11 

### Nuxt configuration

Please change to `[x]` if relevant for this issue:

 - [ ] Applies to a site deployed to a static server (site generated with `nuxt generate`)
 - [x] Applies to a site deployed to a server with a Node backend

### Steps to reproduce

```
i18n: {
    // ...
    detectBrowserLanguage: {
      useCookie: true,
      cookieKey: 'i18n_redirected',
      redirectOn: 'root',  // recommended
    }
  },
```

### What is Expected?
When I open the main site link in a browser in a different language, should he switch to the site in the corresponding language?
Is the cookie set to the corresponding language code?

### What is actually happening?
They don't seem to be working properly

rchl:
Fill out all the requested information including specifying versions and providing reproduction.

stephAc:
Hello, having the same issue where it seems that the browser language is not detected by the pluging and keeps on going for the defaultLocale config, It also applies to the static deployment. 
Thank's for any info on this issue

mjbates7:
Hello, I'm also experiencing the same issues as the above comments. It always goes back to the default locale rather than what is detected. I would have thought if the accept language detects the locale (and redirect is on) it would go to that `{locale}/[...slug]` for example. If no redirect then it should know what locale is detected.

Thank you

markus-gx:
When using `locale` `const { locale, availableLocales } =  useI18n()` it stays the same as the previous comments mentioned. 
Inspected the request headers and they are different than the fallbackLocale configured. So no browser detectino is working. Running on beta.4 and Nuxt 3.0.0 Stable.

Iamlowlo:
I'm using Nuxt 3.0.0 and @nuxtjs/i18n 8.0.0-beta.6
Same issue with browser detection.
I tried to config with and without cookie usage with no luck. Even when useCookie is set to true, the cookie is not even set.
If defaultLocale is not set, project root route returns a 404 (cause that route doesn't exists indeed and needs a redirection that it's not happening)

The config I'm using is the following:
```
        modules: [
            '@nuxtjs/i18n',
        ],
        i18n: {
            strategy: 'prefix',
            locales: ['en-us', 'es'],
            detectBrowserLanguage: {
                useCookie: true,
                cookieKey: 'asdasd'
            },
[...]

PR #2153, opened 2023-06-13
Revert "fix: defineI18nRoute do not work client-side"

kazupon:
Reverts nuxt-modules/i18n#1902


kazupon:
ref: https://github.com/nuxt-modules/i18n/issues/2132#issuecomment-1579595503
ref: https://github.com/nuxt-modules/i18n/issues/1632#issuecomment-1563245295

PR #2163, opened 2023-06-14
fix: fix order of locale merging when using layers and lazy:false

### LINKED ISSUE #2162: Domain-property of locale doesn't get merged correctly when using lazy:false and a Nuxt-Layer
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.15.0`
- Nuxt Version:     `3.2.3`
- Nitro Version:    `2.2.3`
- Package Manager:  `npm@9.5.0`
- Builder:          `vite`
- User Config:      `extends`, `hooks`, `shopware`, `runtimeConfig`, `modules`, `i18n`
- Runtime Modules:  `modules/domainRuntimeInfo`, `nuxt-schema-org@2.1.2`
- Build Modules:    `-`


### Reproduction

I'll provide a reproduction as soon as possible.

### Describe the bug

When using nuxt layers and the "differentDomains" feature, it seems like the `domain` property of the locales do not get merged correctly.
In my specific case the merged configuration contains the `domain` property of the locale defined in the parent-layer.
The child-layer uses the same locales in the configuration as the parent-layer, but with different domains set for each locale.

### Additional context

_No response_

### Logs

```shell
@nuxtjs/i18n:layers project layer lazy option false +1ms                                                                                                                                                                                     12:28:57
@nuxtjs/i18n:layers layer.config.i18n.locales [                                                                                                                                                                                              12:28:57
  {
    code: 'en-GB',
    file: 'en.json',
    domain: 'http://child.example.com'
  },
  {
    code: 'de-DE',
    file: 'de.json',
    domain: 'http://child.example.de'
  }
] +0ms
@nuxtjs/i18n:layers layer.config.i18n.locales [                                                                                                                                                                                              12:28:57
  {
    code: 'en-GB',
    file: 'en.json',
    domain: 'http://parent.example.com'
  },
  { code: 'de-DE', file: 'de.json', domain: 'http://parent.example.de' },
  {
    code: 'en-US',
    file: 'en.json',
    domain: 'http://parent-us.example.com'
  }
] +0ms
@nuxtjs/i18n:layers layer.config.i18n.locales [] +0ms                                                                                                                                                                                        12:28:57
@nuxtjs/i18n:layers layer.config.i18n.locales [] +0ms                                                                                                                                                                                        12:28:57
@nuxtjs/i18n:layers merged locales [                                                                                                                                                                                                         12:28:57
  {
    code: 'en-GB',
    file: 'en.json',
    domain: 'http://parent.example.com'
  },
  { code: 'de-DE', file: 'de.json', domain: 'http://parent.example.de' },
  {
    code: 'en-US',
    file: 'en.json',
    domain: 'http://parent-us.example.com'
  }
] +0ms

```
```

niklaswolf:
### 🔗 Linked issue
#2162 

### ❓ Type of change

### 📚 Description
I'm pretty sure that this resolves #2162 . Technically my solution is a breaking change, so I marked it as such.

When not using the lazy mode, the order of the layers was the wrong way round before merging the locales. This results in the project layer (the first in the layers-array) having the least priority and beeing overriden by following layers, which is not the behaviour described in the documentation (and it makes no sense like this too 😆 )
This can be seen in the changes I made to the playground (actually I don't know if changes to the playground should be comitted, as there are many many commented cases etc.)

I tested also in lazy mode, there the debug-log shows that the locales are merged correctly, so as I see it, the problem only is present when using non-lazy-mode at the moment.

To see the debug output, run it with `DEBUG=@nuxtjs/i18n:layers pnpm dev `

### 📝 Checklist


kazupon:
Welcome contribution! ❤️ 

Look good too me!

@BobbieGoede 
if you have the time, Could you review this PR?

BobbieGoede:
@niklaswolf 
Thank you for contributing! I'm trying to run the playground/reproduction but I'm getting the following error: `[vite-node] [VITE_ERROR] /.nuxt/i18n.options.mjs` can you confirm it works locally without other changes? It could be something wrong with my setup 😅

Based on the debug logs it looks like this PR fixes the issue, can you edit the test for layers with `lazy: false` in `layers_simple.spec.ts` so that it confirms the layer locale domains are overwritten?

niklaswolf:
I'll try to do so. I think I have a new test added locally that should work with the new code, but somehow it keeps failing.
Is it possible that the local tests do not use the local source-code, but the latest version from npm?
https://github.com/nuxt-modules/i18n/blob/next/specs/fixtures/domain/package.json#L12

BobbieGoede:
The tests should be using the local build, the dependency in the fixtures are overridden by the lockfile. See https://github.com/nuxt-modules/i18n/blob/next/pnpm-lock.yaml#LL4C25-L4C25 and https://github.com/nuxt-modules/i18n/blob/next/pnpm-workspace.yaml, are you using pnpm? I think it won't work using npm or yarn but I haven't tried.

I've written some tests for this PR in the meantime, I'll add them to your branch!

BobbieGoede (approved):
The tests should confirm that your changes have resolved the issue, thank you for contributing! ❤️

niklaswolf:
@BobbieGoede thanks for taking the time to add the test, I'll have a look at it to be able to provide some tests next time faster :)

PR #2180, opened 2023-06-26
fix(alias): do not use runtime version during nuxt prepare

yassilah:
### ❓ Type of change

### 📚 Description

Using `vue-i18n` runtime version only in dev mode results in wrong/no types when running `nuxt prepare`. This PR fixes it.

### 📝 Checklist


kazupon (approved):
Good contribution!
Thanks!

Issue #2141, opened 2023-06-07
Migrating hook i18n:extend-messages to i18n:registerModule does not work


devonik:
### Environment

------------------------------
- Operating System: Darwin
- Node Version:     v18.14.0
- Nuxt Version:     3.5.3
- Nitro Version:    2.4.1
- Package Manager:  pnpm@8.6.0
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

Write a simple nuxt module and try to register messages to i18n with the new hook i18n:registerModule within modules setup().

https://stackblitz.com/edit/github-asw9fc

### Describe the bug

My new hook in the module setup() looks like this:
```
nuxt.hook('i18n:registerModule', (register: any) => {
      register({
        // langDir path needs to be resolved
        langDir: resolver.resolve('./runtime/lang'),
        experimental: {
          jsTsFormatResource: true
        },
        locales: [
          {
            code: 'de',
            file: 'general.ts',
          },
          {
            code: 'en',
            file: 'general.ts',
          },
        ]
      })
    })
```
In https://v8.i18n.nuxtjs.org/guide/lazy-load-translations I saw the required experimental.jsTsFormatResource but with no effect
The defineI18nLocale() in general.ts function is not called by server start.

### Additional context

My old setup with hook i18n:extend-messages was running
```
nuxt.hook(
      'i18n:extend-messages',
      async (additionalMessages, localeCodes) => {
        const messages = await fetchApi(config.api)
        additionalMessages.push(messages)
      },
    )
```

kazupon:
Thank you for your feedback!

>experimental.jsTsFormatResource
The register callback in the `i18n:registerModule` hook does not accept it. Only `langDir` and `locales`.

And, please give us a minimal reproduction of the code in the stackblize or github repo 🙏 
we cannot check your issue with text-only information for reproduction.

devonik:
> Thank you for your feedback!
> 
> > experimental.jsTsFormatResource
> > The register callback in the `i18n:registerModule` hook does not accept it. Only `langDir` and `locales`.
> 
> And, please give us a minimal reproduction of the code in the stackblize or github repo 🙏 we cannot check your issue with text-only information for reproduction.

I put a stackblitz link into description 👍

Randoooom:
Any updates on this?

devonik:
Nice that there seems a fix. When it will be released I would like to test it

[...]

PR #2188, opened 2023-06-30
fix: registerModule hook not working for js/ts format resources

BobbieGoede:
### 🔗 Linked issue
#2141 

### ❓ Type of change

### 📚 Description
It seems that only locales within the project `langDir` were being resolved in the following line: https://github.com/nuxt-modules/i18n/blob/next/src/utils.ts#L62C1-L62C92

This results in locales added via the `i18n:registerModule` hook and possibly layers too would not get a `type` or `types` assigned to it, which as far as I understand is required for js/ts format locales (I'm still catching up with the changes made in the last few months 😅). 

Note that my changes removes the file extension check/filter in the resolve function, I could add a basic matcher to ensure each file has the correct extension if needed.

### 📝 Checklist

### DIFF
--- a/playground/module-experimental/index.ts
+++ b/playground/module-experimental/index.ts
@@ -1,0 +1,36 @@
+import { defineNuxtModule, addPlugin, createResolver } from '@nuxt/kit'
+
+// Module options TypeScript interface definition
+export interface ModuleOptions {}
+
+export default defineNuxtModule<ModuleOptions>({
+  meta: {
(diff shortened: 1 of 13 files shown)


kazupon (approved):
Good job!
Thank you! ❤️

jaydeep-stallion:
I added **i18n:registerModule** hook in my index.ts module's file but its not translating the messages

image

PR #2193, opened 2023-07-02
fix: lazy load i18n files

userquin:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Resolves #2192

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #2212, opened 2023-07-08
fix: missing hash query param in `load` hook for virtual conf

userquin:
### 🔗 Linked issue

#2211 

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Resolves #2211

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #2219, opened 2023-07-10
Revert "fix: lazy load i18n files"

kazupon:
Reverts nuxt-modules/i18n#2193

### DIFF
--- a/src/runtime/utils.ts
+++ b/src/runtime/utils.ts
@@ -122,7 +122,7 @@ export async function loadInitialMessages<Context extends NuxtApp = NuxtApp>(
       await Promise.all(fallbackLocales.map(locale => loadLocale(context, locale, setter)))
     }
     // load initial messages
-    const locales = lazy ? localeCodes : [...new Set<Locale>().add(defaultLocale).add(initialLocale)]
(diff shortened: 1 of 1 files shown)


userquin:
@kazupon why do you revert also the lazy commit ?

kazupon:
Lazy load does not works.
You can try it in the nuxt-i18n playground.

userquin:
It is working in my vuetify nuxt module with single and multiple files per locale:
https://github.com/userquin/vuetify-nuxt-module

I Will check it later in my local.

userquin:
@kazupon the i18n playground seems to be working with a few changes from #2219 , the playground should change the lang when clicking on the links no?

imagen

imagen

userquin:
EDIT: it seems something wrong with the imports, check the screenshot here: https://github.com/nuxt-modules/i18n/pull/2212#issuecomment-1634312968

About the null entry, in the playground here also missing some entry/request, the virtual entry is not there.

I guess it is something in Nuxt, Nuxt DevTools inspector isn't able to intercept virtual modules when virtual module using default Vite name convention (prefixing ids with `\0`).

I'll try to change the resolved virtual ids here and check if it is working...


imagen

kazupon:
Thanks!
I look forward to PR from you :)

PR #2229, opened 2023-07-13
feat: add NuxtLinkLocale component

### LINKED ISSUE #2222: Provide a nuxt-link-locale component who combine nuxt-link + localePath
### Describe the feature

I see many code where code like this is used:
```html
<nuxt-link :to="localePath('/')">{{ $t('home') }}</nuxt-link>
```
so i could imagine it would lead to a better dev experience when a own component would exists,
which improve the read ability (may also easier to convert existing code)

So idea would be something like that:
```html
<nuxt-link-locale to="/">{{ $t('home') }}</nuxt-link-local>
```

It would be also allow easier to tracking code places where locale is not done (for good/bad reasons).

to think this further:
```html
 <!-- with lang para -->
<nuxt-link :to="localePath('index', 'en')">Homepage in English</nuxt-link>
<nuxt-link-locale to="index" lang="en">Homepage in English</nuxt-link-local>

<!-- change language -->
<nuxt-link :to="switchLocalePath('en')">English</nuxt-link>
<nuxt-link-locale switch="en">English</nuxt-link-local>
```

About "Could this feature be implemented as a module?" i don't know enough internal stuff about vue/nuxtjs/i18n to awnser this.

### Additional information

### Final checks

BobbieGoede:
### 🔗 Linked issue
#2222 

### ❓ Type of change

### 📚 Description
Resolves #2222 

Adds a new component `<NuxtLinkLocale />` which is built on top of `<NuxtLink />`, this component will internally use the `localePath` function to resolve a localized `<NuxtLink />`. It also supports passing a `locale` prop to force the localization to resolve the link with the passed language.

### 📝 Checklist


playground/pages/index.vue
```diff
@@ -75,7 +75,8 @@ definePageMeta({
       <NuxtLink :to="localePath('/')">Home</NuxtLink> | <NuxtLink :to="localePath({ name: 'about' })">About</NuxtLink> |
       <NuxtLink :to="localePath({ name: 'blog' })">Blog</NuxtLink> |
       <NuxtLink :to="localePath({ name: 'category-id', params: { id: 'foo' } })">Category</NuxtLink> |
-      <NuxtLink :to="localePath({ name: 'history' })">History</NuxtLink>
+      <NuxtLinkLocale :to="{ name: 'history' }" class="history-link">History</NuxtLinkLocale> |
+      <NuxtLinkLocale :to="'/'" lang="ja" activeClass="link-active">Home (Japanese)</NuxtLinkLocale>
```
kazupon: I think it should named `locale` prop, not `lang` prop, because in the i18n context, we are mostly naming `locale`, not `lang`.

The ECMAScript Internationalization API also named with locale. The first argument of the `Intl.DateTimeFormat` constructor is called as `locales`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

src/runtime/components/NuxtLinkLocale.vue
kazupon, line 1: Like library such as Nuxt i18n module, we should make components with render functions, not with SFC. because it is easier to control and optimize in JavaScript (TypeScript) than in the SFC compiler.

If you don't know what this means, you may want to read the implementation code for the build-in components provided by Nuxt.

kazupon (changes requested):
Wow! Great job!
Your PR is really helpful for more discussions.

kazupon:
@dargmuesli @ineshbose 
If you have the time, please review.
we'd like to get more feedback from you!

src/runtime/components/NuxtLinkLocale.vue
kazupon: e.g. nuxt-link
https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/components/nuxt-link.ts

playground/pages/index.vue
BobbieGoede: That makes sense and is consistent with the rest of the module, I've changed the prop to `locale`.

src/runtime/components/NuxtLinkLocale.vue
BobbieGoede: I've changed the component to be a render function, I didn't know that it was a performance consideration 😅. It was a smaller change than I expected, curious if my change is what you had in mind.

BobbieGoede:
The CI checks kept failing so I added the following npm script `"postinstall": "nuxi prepare"` to generate the types used when importing from `#imports`. This is similar to the [official project starter](https://github.com/nuxt/starter/tree/v3), I saw the [official module starter](https://github.com/nuxt/starter/tree/module) uses types generated from inside the `playground` directory, I felt that wouldn't make much sense in the context of a CI job.

BobbieGoede:
I've added a simple components page to the docs describing the component and its usage. I also reordered the api pages, so that the components page comes second, similar to the official Nuxt docs, I can revert this reorder if preferred (it's hardly a necessary change in hindsight 🫣).

docs/content/4.API/2.components.md
```diff
@@ -0,0 +1,36 @@
...
+This component is built on top of [`<NuxtLink>`](https://nuxt.com/docs/api/components/nuxt-link#nuxtlink) but changes the default behavior by internally using [`localePath()`](/api/vue#localepath) to make it easier to link to localized routes.
+
+### Examples
+
+#### Basic usage
+```vue
+<template>
+  <NuxtLinkLocale to="/">{{ $t('home') }}</NuxtLinkLocale>
+  <!-- equivalent to -->
+  <NuxtLink :to="localePath('/')">{{ $t('home') }}</NuxtLink>
```
kazupon, line 23: Strictly, `localePath` requires a call to `useLocalePath` with the script block.

I think it would be better to write the equivalent separately as follows, instead of writing it here.

```vue
<script setup>
</script>

<template>
  <NuxtLink :to="localePath('/')">{{ $t('home') }}</NuxtLink>
</template>
```

src/runtime/components/NuxtLinkLocale.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import type { PropType } from 'vue'
+import type { RouteLocationRaw } from 'vue-router'
+import type { RawLocation, RouteLocation } from '@intlify/vue-router-bridge'
+
+const NuxtLinkLocale = defineNuxtLink({ componentName: 'NuxtLinkLocale' })
+
+export default defineComponent({
+  name: 'NuxtLinkLocale',
+  props: {
```
kazupon, line 11: With this method of implementation, if new props are added or deprecated in NuxtLink, I think we would not be able to maintain compatibility and there would be maintenance to follow 🤔 

@danielroe 
Is there a better way to do… NuxtLink props inherit without hard-coding as follows?

danielroe: I think you can directly spread the props of NuxtLink here. (`...NuxtLinkLocale.props`)

kazupon: Oh, I see, NuxtLink is exported!

When I looked at the code, it didn't look exported to me, which is why I asked you the mention!
https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/components/nuxt-link.ts

Thanks ❤️

package.json
```diff
@@ -63,7 +63,8 @@
     "test": "run-s test:types test:unit test:spec",
     "test:types": "tsc --noEmit",
     "test:unit": "vitest run test",
-    "test:spec": "vitest run specs"
+    "test:spec": "vitest run specs",
+    "postinstall": "nuxi prepare"
```
rchl: Shouldn't do that. It will also run (and fail) when people install the package.

src/runtime/components/NuxtLinkLocale.ts
BobbieGoede: I somehow forgot it was possible to spread the props of a component, much better this way, thanks! 🚀

docs/content/4.API/2.components.md
BobbieGoede: Agreed, I've changed the examples to be more complete/realistic.

package.json
BobbieGoede: Hmm I didn't consider that, I moved `nuxi prepare` to the `test` script now. This way the CI test job won't fail, but anyone contributing will initially be missing types for `#imports`. Should we add a `prepare` script and run it for both `test` and `dev` scripts?

kazupon: > Should we add a prepare script and run it for both test and dev scripts?

Yes, we should.

At this point, I can only come up with that idea.
The monorepo for the nuxt i18n module, unfortunately, that is root, not `packages`, so we can't use `postinstall`.
In the future, we should refactor the directory structure and make it a package structure like `docs`, `playground` and  `i18n`

kazupon: ref: nuxt-devtools: https://github.com/nuxt/devtools/blob/main/package.json

BobbieGoede: For now I have added a `dev:prepare` script which runs before `dev` and `test`.

kazupon (approved):

kazupon:
@BobbieGoede 
Well done!
Thank you so much!
I’ll Merge this PR!

Tush1810:
This component forcefully removes trailing slashes. Is there a way to add trailing slashes to it?

PR #2268, opened 2023-07-28
docs(layers): correct configuration requirements

dargmuesli:
### 🔗 Linked issue

n/a

### ❓ Type of change

### 📚 Description

In my testing it is required to also have `locales` set in the project which extends another. Also, the module does not have to be redefined in a project that extends another as the module is already defined in the project which is extended.

Beyond that, It would be interesting as well to have the documentation explain *why* it is necessary to repeat some of the configuration in the extending project! :bulb: I'm wondering why it is like that and don't know :)

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/15.layers.md
+++ b/docs/content/2.guide/15.layers.md
@@ -20,7 +20,7 @@ Page routes defined in `i18n.pages` in each layer configuration will be merged a
 
 ### Locales
 ::alert{type="warning"}
-A project extending layers containing lazy-loaded translations is still required to have `lazy` and `langDir` options configured.
+A project extending layers containing lazy-loaded translations is still required to have `lazy`, `langDir` and `locales` options configured.
 ::
 
 Extending your project with a layer that contains locales can be done as follows:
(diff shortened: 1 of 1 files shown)


kazupon (approved):
Thanks! LGTM!

@BobbieGoede 
Do you have anything to add?

BobbieGoede:
Looks like that property _is_ required, thanks for the PR!

> Beyond that, It would be interesting as well to have the documentation explain why it is necessary to repeat some of the configuration in the extending project! 💡 I'm wondering why it is like that and don't know :)

Especially when using multiple layers, properties can get overwritten by each layer, so if a project is consuming a layer with `lazy: true` and a layer with `lazy: false`, depending on the order of `extend` the `lazy` property will be different. If `lazy` ends up being `true` then main the project is required to have `langDir` set.

Another reason for this is that files defined in `locales` are resolved relative to `langDir`, so we need a `langDir` in the main project to resolve `locales` defined in layers containing lazy loaded locales.

This behavior is a bit unintuitive and hard to explain well, ideally the main project extending `i18n` layers should not be required to configure anything. I'm actually working on a change to make this possible, so hopefully I can get that done soon!

BobbieGoede (approved):

PR #2280, opened 2023-07-31
chore: update to latest nuxt packages

danielroe:
### ❓ Type of change

### 📚 Description

This updates Nuxt and uses `PublicRuntimeConfig` interface (to avoid overwriting other properties on `RuntimeConfig`.

Note that there's also a mismatch with `vue-i18n-routing` as we support redirects with object syntax, not just strings. I've also swapped a test assertion around - I think it was wrong before(?) But let me know if that's incorrect and there's actually a bug to address here.

### 📝 Checklist

### DIFF
--- a/specs/lang_switcher.spec.ts
+++ b/specs/lang_switcher.spec.ts
@@ -34,8 +34,8 @@ test('switching', async () => {
 
   // page path
   expect(await getData(page, '#home-use-async-data')).toMatchObject({
-    aboutPath: '/about',
-    aboutTranslation: 'About us'
+    aboutPath: '/fr/about',
+    aboutTranslation: 'À propos'
   })
   expect(await page.getAttribute('#lang-switcher-with-nuxt-link a', 'href')).toMatch('/')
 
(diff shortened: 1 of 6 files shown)


kazupon (approved):
Thanks! ❤️ 
I'm happy to the PR from you, so I can see my mistake.

specs/lang_switcher.spec.ts
```diff
@@ -34,8 +34,8 @@ test('switching', async () => {
 
   // page path
   expect(await getData(page, '#home-use-async-data')).toMatchObject({
-    aboutPath: '/about',
-    aboutTranslation: 'About us'
+    aboutPath: '/fr/about',
+    aboutTranslation: 'À propos'
```
kazupon, line 38: oh, nuxt i18n has still bug.
Since your PR is an update of the deps module, we will accept it, and I’ll open the issue.

NOTE:
cannot redirect to target locale when change the locale with `setLocale`

kazupon: I've just opened #2283

Issue #2206, opened 2023-07-07
nuxt 3 module i18n dont working on nuxt layers


productdevbook:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v18.16.0
- Nuxt Version:     3.6.1
- Nitro Version:    2.5.2
- Package Manager:  pnpm@8.5.1
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://github.com/productdevbook/i18n-bugs-layer

### Describe the bug

pnpm install
pnpm dev admin

✔ Nitro built in 364 ms                                                                                                                        nitro 11:55:20 AM
[intlify] Not found 'hello' key in 'en-US' locale messages.                                                                                           11:55:21 AM
[intlify] Fall back to translate 'hello' key with 'en' locale.                                                                                        11:55:21 AM
[intlify] Not found 'hello' key in 'en' locale messages.                                                                                              11:55:21 AM
[intlify] Not found 'hello' key in 'en-US' locale messages.                                                                                           11:55:21 AM
[intlify] Fall back to translate 'hello' key with 'en' locale.                                                                                        11:55:21 AM
[intlify] Not found 'hello' key in 'en' locale messages.                                                                                              11:55:21 AM

### Additional context

pnpm install
pnpm dev admin

thaikolja:
Same here. All packages default (using Nuxt Content).

Tigriz:
I got the same problem and tried `nuxi upgrade --force` once again (even though I updated the day 3.6.1 released) and it worked, dunno what happened tbh.

productdevbook:
> I got the same problem and tried `nuxi upgrade --force` once again (even though I updated the day 3.6.1 released) and it worked, dunno what happened tbh.

I updated the repo as you said, still the same problem persists.

kazupon:
Hi!

I’ve checked your minimal reproduction repo.
I’ve noticed that you need to set up two things to make nuxt i18n work with nuxt layer.


### 1. set your custom nuxt module at `modules` options

You need to set your nuxt module defined in the nuxt layer (`layer` dir) of the `layer` dir to `nuxt.config.ts` as follows

```ts
// https://nuxt.com/docs/api/configuration/nuxt-config
import MyModule from './module'
export default defineNuxtConfig({
  modules: [
    // https://i18n.nuxtjs.org/
    MyModule,
    '@nuxtjs/i18n',
  ],
[...]

PR #2290, opened 2023-08-02
feat: project relative layer locale resolution

BobbieGoede:
### 🔗 Linked issue
#2206 

### ❓ Type of change
These changes fix some unintuitive behaviors we have regarding layer support, it could be seen as a bug fix or feature enhancement.

At this moment projects are required to do some configuring to be able to make use of extended i18n layers. All locale files are resolved relative to the main project's `langDir` so the directory has to exist and be defined at `i18n.langDir`, which in turn requires `i18n.locales` to contain `LocaleObject` entries. (see [related issue discussion](https://github.com/nuxt-modules/i18n/issues/2206#issuecomment-1627676007) and [related PR discussion](https://github.com/nuxt-modules/i18n/pull/2268))

This PR changes the way the locales are resolved by looping through all layers (project and extended) and resolving all locale files relative to the main project, which should make extending i18n layers easier.

~I believe this may make some breaking edge cases possible, I still need to test what happens when mixing `lazy` and non `lazy` layers or mixing layers with `string` locales vs `LocaleObject` locales.~ The final merged value of `lazy` is used to treat all provided locales as either lazy or not, this seems to work fine.

Extending a project's `i18n` config with locales and messages is getting easier and easier, should we think of ways for projects to filter out locales as well?

⚠️ Important note ⚠️
I have added a workaround/hack to be able to properly test our layer implementation, I did this by changing the `setup` function inside `@nuxt/test-utils` included in this repo (`specs/utils`). The changes made to the test-utils can be removed, but we will have to be mindful to use the added `overrides` property to set project layer configurations.

The `@nuxt/test-utils` package does not apply `nuxtConfig` overrides to the project/first layer (`nuxt.options._layers[0]`), overrides are only applied to the fully merged options. The layer implementation intentionally ignores merged options as files need to be handled/resolved per layer. To work around this I have added an `overrides` key, essentially implementing our own override approach, which is handled/applied by the module itself. if there is a better way to do this please let me know. 

Some tests had to be changed as the layer implementation incorrectly allowed some broken configurations. Also, the `gen.test.ts` unit tests results are now technically faulty as the `langDir` locale paths are not resolved inside `gen.ts` but beforehand.

### 📝 Checklist

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -14,13 +14,13 @@ importers:
         version: 9.3.0-beta.24
       '@intlify/unplugin-vue-i18n':
         specifier: ^0.12.2
-        version: 0.12.2(rollup@3.27.0)(vue-i18n@9.3.0-beta.24)
+        version: 0.12.2(rollup@3.28.0)(vue-i18n@9.3.0-beta.24)
       '@mizchi/sucrase':
         specifier: ^4.1.0
         version: 4.1.0
       '@nuxt/kit':
         specifier: ^3.6.5
-        version: 3.6.5(rollup@3.27.0)
+        version: 3.6.5(rollup@3.28.0)
       '@vue/compiler-sfc':
         specifier: ^3.3.4
         version: 3.3.4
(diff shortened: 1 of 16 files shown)


kazupon (approved):
Thanks!
I got about importance note!
LGTM!

dargmuesli:
Does this change anything about the content I edited in https://github.com/nuxt-modules/i18n/pull/2268? Like, do the docs need an update?

BobbieGoede:
@dargmuesli 
It does, I should have included it in this PR as it's not necessary to configure any options to extend on an i18n layer now. I will open a PR to fix it! 😅

PR #2297, opened 2023-08-04
feat: allow to set the plugin as parallel

huang-julien:
### 🔗 Linked issue

Hi :wave: 
Since a few version in nuxt (3.5?), we are able to use plugin with an object syntax. 
This syntax allow us to set `parallel` to be true so we can get better performances by not blocking awaiting a plugin before loading the next one.
Since users tend to have async plugins, this PR can improve the performance of projects not relying on i18n within their plugins. 

### ❓ Type of change

### 📝 Checklist


src/runtime/plugins/i18n.ts
kazupon, line 1: I think this file has been all formatted, so we makes it hard to review 😅 
Can you please make it so that it's not formatted unnecessarily? 🙏

src/constants.ts
```diff
@@ -68,7 +68,8 @@ export const DEFAULT_OPTIONS = {
   pages: {},
   skipSettingLocaleOnNavigate: false,
   types: 'composition',
-  debug: false
+  debug: false,
+  parallelPlugin: false
```
kazupon, line 72: Oh, I see. That's instresting!
https://github.com/nuxt/nuxt/pull/20460

When adding this option, we need to add docs about the option.
Can you also write it to docs?

src/types.ts
```diff
@@ -88,6 +88,7 @@ export type NuxtI18nOptions<Context = unknown> = {
   types?: 'composition' | 'legacy'
   debug?: boolean
   dynamicRouteParams?: boolean
+  parallelPlugin: boolean
```
kazupon:
```suggestion
  parallelPlugin?: boolean
```
That option should be typed defined above. Because options is optional.

kazupon (changes requested):
Thank you for your contribution! ❤️ 
I’ve just reviewed. 👀 
Please check it!

src/runtime/plugins/i18n.ts
huang-julien: Will do :+1:

huang-julien: hmmm since the plugin has been converted to Object syntax plugin, I don't think we can remove the format (i'm not sure how tbh). 
Basically, the plugin has been changed with a `setup` function, a `name` and a `parallel` boolean.
The content of the setup function hasn't changed, justed moved into `setup`

kazupon: > Basically, the plugin has been changed with a setup function, a name and a parallel boolean.

Okay, I've just checked it!

kazupon:
You need to resolve build error on github actions
https://github.com/nuxt-modules/i18n/actions/runs/5813601526/job/15768570502?pr=2297

src/constants.ts
huang-julien: hmm... There's no info on the parallel option (on nuxt.com) it seems. I'll reference the PR, update the nuxt doc and then update this one.

huang-julien: nvm i'm blind :joy: .... there's indeed a doc for parallel plugins

kazupon:
oh, you need to update unit tests too 😅 
https://github.com/nuxt-modules/i18n/actions/runs/5825092419/job/15803087501?pr=2297

huang-julien:
oops 😅 i'm a bit ashamed by this one. Let me update it

kazupon:
@huang-julien 
Thanks!

This PR has still conflicts, can you resolve it? 🙏 
You are one step closer to having this PR merged!

huang-julien:
locally
- [x] test
- [x] lint
- [x] test:unit
- [x] test:spec
- [x] test:types
- [x] format

huang-julien:
I have a strange behavior using vitest, there's a context conflict with nuxt and unctx

kazupon (approved):
Well done!
Thanks!

PR #2306, opened 2023-08-10
ci: fix playwright cache being ignored due to old version

BobbieGoede:
### ❓ Type of change

### 📚 Description
While trying to debug the test action I noticed we are using/used the `ci.yml` of `vitejs` as an example to set up `playwright`. After checking their current I found that they fixed an issue with the way the playwright browser is being cached which we still have (ref: https://github.com/vitejs/vite/pull/11254). 

Right now in the test job during the install playwright step, you will always see the old cached browsers being removed in favor of downloading the latest available version. This PR changes this to cache it and prevent redownloading, using the `playwright` version in `package.json` as cache key.

### 📝 Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,8 @@ name: CI
 
 env:
   CI: true
+  # install playwright binary manually (because pnpm only runs install script once)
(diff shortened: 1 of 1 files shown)


kazupon (approved):
Thanks!

Issue #2343, opened 2023-08-24
[PNPM] WARN  Ignore Vue I18n configuration file does not exist at i18n.config on layer


LeCoupa:
### Environment

Nuxi 3.6.5                                                           8:46:32 PM
                                                                     8:46:33 PM
[8:46:33 PM] RootDir: /Users/lecoupa/Documents/TheCompaniesAPI/thecompaniesapi-app
Nuxt project info: (copied to clipboard)                             8:46:33 PM

------------------------------
- Operating System: Darwin
- Node Version:     v20.5.0
- Nuxt Version:     3.6.5
- Nitro Version:    2.5.2
- Package Manager:  pnpm@8.6.2
- Builder:          vite
- User Config:      build, devServer, experimental, nitro, extends, imports, runtimeConfig, vite, components, css, modules, app, typescript, devtools, i18n, pinia, plausible, pwa
- Runtime Modules:  @nuxtjs/color-mode@3.3.0, @nuxtjs/i18n@8.0.0-rc.3, @pinia/nuxt@0.4.11, @unocss/nuxt@0.53.6, @vite-pwa/nuxt@0.1.0, @vueuse/nuxt@10.2.1, @nuxtjs/plausible@0.2.1, floating-vue/nuxt, vue-recaptcha/nuxt
- Build Modules:    -
------------------------------

### Reproduction

Here is my the i18n config object if it can help

```typescript
i18n: {
    defaultLocale: 'en',
    langDir: './locales/',
    lazy: true,
    strategy: 'prefix_except_default',
    vueI18n: './config/i18n.config.ts',

    detectBrowserLanguage: {
      useCookie: true,
      fallbackLocale: 'en',
    },

    locales: [
      {
        code: 'en',
        iso: 'en',
        file: 'en.ts',
      },
      {
        code: 'fr',
        iso: 'fr',
        file: 'fr.ts',
      },
    ],
  },
```

### Describe the bug

I don't know what happens but wherever I put it the `i18n.config.ts` file is never found. I have this strange error (notice the nuxt-seo-kit inside the url).

```
[...]

PR #2347, opened 2023-08-26
fix: i18n layer vue i18n resolution

BobbieGoede:
### 🔗 Linked issue
#2343 

### ❓ Type of change

### 📚 Description
This changes `resolveLayerVueI18nConfigInfo` to only try to resolve Vue I18n configuration paths for layers that contain configuration for `@nuxtjs/i18n`. 

@kazupon would this be correct or is it possible and should we take into account layers could have an `i18n.config` file without using `@nuxtjs/i18n`?

### 📝 Checklist

### DIFF
--- a/src/layers.ts
+++ b/src/layers.ts
@@ -140,9 +140,11 @@ export async function resolveLayerVueI18nConfigInfo(nuxt: Nuxt, buildDir: string
   const layers = [...nuxt.options._layers]
   layers.shift()
   return await Promise.all(
-    layers.map(layer => {
-      const i18n = getLayerI18n(layer)
-      return resolveVueI18nConfigInfo(i18n || {}, buildDir, layer.config.rootDir)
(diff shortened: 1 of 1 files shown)


kazupon:
> would this be correct or is it possible and should we take into account layers could have an i18n.config file without using @nuxtjs/i18n?

Yes, this is correct.
We shouldn't take into account an `i18n.config` file without using `@nuxtjs/i18n`.

kazupon (approved):

dargmuesli:
How is the vue-i18n config meant to be inferred from multiple layers? I would imagine configurations like `fallbackWarn: false` to bubble up, but it seems that a project has to set that configuration itself even if a layer it uses set it already.
I looked at the source code a bit and it seems to me that only `messages` are collected from the multiple vue-i18n configs? Is that right?

Issue #2185, opened 2023-06-28
Unexpected lazy-loading, fetch behavior with multiple locales


tomsdob:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v16.20.0
- Nuxt Version:     3.6.1
- Nitro Version:    2.5.1
- Package Manager:  npm@9.4.2
- Builder:          vite
- User Config:      modules, i18n
- Runtime Modules:  @nuxtjs/i18n@8.0.0-beta.12
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/github-nmhzwq-gudpmp

### Describe the bug

Unexpected translation behavior when using lazy-loaded translations with a fetch request and multiple locales.

### Additional context

I18n does not apply the current locale translations while in default locale - `'lv'`. The translations appear in `'en'` locale instead. 
In dev mode sometimes half of the translations are correct, half are wrong after hydration. 
In build mode all of the translations are wrong and don't respect the default locale.
Translation `id` param can be seen by using `?id=123` query in the URL.

agracia-foticos:
same issue lazy load not runs!

BobbieGoede:
It looks like there's a few things that could be causing the described behavior. I changed your reproduction to demonstrate the different configurations that could be at fault [here](https://stackblitz.com/edit/github-nmhzwq-xqfhnt?file=nuxt.config.ts,package.json).

The reproduction can be run with the following variations:
* `npm run dev` basically your original reproduction
* `npm run dev:no-browser-detect` this turns off browser language detection, which I think is causing the switching of languages in the first place
* `npm run dev:unique-file` this sets the locales to different locale files to check if caching is at issue (see the bottom info notice on this [doc page](https://v8.i18n.nuxtjs.org/guide/lazy-load-translations))
* `npm run dev:both` this combines the two adjustments above

It seems that the elements using `v-html` do not update whenever i18n variables are involved, I have added some elements to demonstrate this. When `detectBrowserLanguage` is enabled and the locale is changed, only plain variables in `v-html` are updated and variables inserted as normal using `{{ }}`.

When using `npm run dev:no-browser-detect` the elements using `v-html` do not seem to update and keep the translation of the language fetched on page load.

When using `npm run dev:unique-file` not much seems to change.

[...]

PR #2355, opened 2023-08-30
feat: disable caching for dynamic locale messages

BobbieGoede:
### 🔗 Linked issue
#2185 

### ❓ Type of change

### 📚 Description
This changes the messages caching behaviour, only message file keys are used to store and retrieve cached messages. If a message loader is dynamic (ie. a function) then its key and messages are not used for caching. 

As described in https://github.com/nuxt-modules/i18n/issues/2185#issuecomment-1695083083 we want to disable caching by default for `js` and `ts` files, this PR is a step into that direction but does not offer control over caching yet.

It may be preferred to delay this until I have made a solution that provides control over the caching via a `cache` property.

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/8.lazy-load-translations.md
+++ b/docs/content/2.guide/8.lazy-load-translations.md
@@ -78,10 +78,7 @@ About `defineI18nLocale` details, see the [here](/api/composables#defineI18nLoca
 ::
 
 ::alert{type="warn"}
-
-The js / ts format is currently an experimental feature and is disabled by default. 
-
-If you want to use it, you must set the `experimental.jsTsFormatResource` module option to `true`.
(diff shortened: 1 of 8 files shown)


kazupon (commented):
Good job!
Thanks!
Your PR is a great for improving next step of lazy-loading caching.

> we want to disable caching by default for js and ts files, this PR is a step into that direction but does not offer control over caching yet.

I hope that you will put the docs about it.
https://v8.i18n.nuxtjs.org/guide/lazy-load-translations

Near the future it may not be necessary with `cache` property, but we need to understand for caching of `js` / `ts` resources.

BobbieGoede:
I have added a warning about the caching behaviour! Will add a section about cache when adding support for the `cache` property.

kazupon (approved):

PR #2357, opened 2023-08-31
feat: locale message file cache configuration

### LINKED ISSUE #2293: Option to clear or disable cache when using lazy:true
### Describe the feature

It would be great if there could be an option to disable cache when using lazy:true.

I have a problem where i use mergeLocaleMessage() in my code on some route changes.
Why because, I like to fetch translations based on route.
In my scenario, i have A LOT of pages with A LOT of translations for each page and it would be expensive to load them all on initial page load.
So now I am fetching translations on route change and then using mergeLocaleMessage() to add them.
But the cache breaks everything if the initial route is different. It loads only translations on client-side.

Little basic reproduction: https://stackblitz.com/edit/nuxt-starter-nus1pf?file=app.vue

### Additional information

### Final checks

BobbieGoede:
### 🔗 Linked issue
#2185
#2293

### ❓ Type of change

### 📚 Description
Resolves #2293

Changes locale message file configuration, `file` and `files` on locales can now be set to `{ path: string; cache?: boolean }`, if cache is left unset caching be enabled by default for locale files that are not `dynamic`.

It was a bit difficult to think of a proper way to test whether cache configuration is taking effect as it happens internally, the added test checks this based on debug logs, perhaps there is a better way to do this.

### 📝 Checklist

### DIFF
--- a/src/runtime/utils.ts
+++ b/src/runtime/utils.ts
@@ -46,7 +46,7 @@ import type { I18n, Locale, FallbackLocale, LocaleMessages, DefineLocaleMessage 
 import type { NuxtI18nOptions, DetectBrowserLanguageOptions, RootRedirectOptions } from '#build/i18n.options.mjs'
 import type { DetectLocaleContext } from '#build/i18n.internal.mjs'
 import type { DeepRequired } from 'ts-essentials'
-import type { LocaleOption } from './types'
+import { LocaleFile } from '../types'
 
 export function _setLocale(i18n: I18n, locale: Locale) {
   return callVueI18nInterfaces(i18n, 'setLocale', locale)
@@ -517,10 +517,9 @@ export function extendBaseUrl<Context extends NuxtApp = NuxtApp>(
   }
 }
 
-export const getLocalePaths = (locale: LocaleOption): string[] => {
+export const getLocalePaths = (locale: LocaleObject): string[] => {
   if (locale.file != null) {
-    // @ts-ignore
-    return [locale.file as LocaleFile].map(x => (typeof x === 'string' ? x : x.path))
+    return [locale.file as unknown as LocaleFile].map(x => (typeof x === 'string' ? x : x.path))
   }
 
   if (locale.files != null) {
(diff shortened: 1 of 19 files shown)


src/types.ts
```diff
@@ -32,7 +32,20 @@ export type LocaleInfo = {
   paths?: string[]
   hashes?: string[]
   types?: LocaleType[]
-} & LocaleObject
+} & Omit<LocaleOption, 'files'> & { files: LocaleFile[] }
+
+export type LocaleFile = { path: string; cache?: boolean }
+
+export type LocaleOption = {
```
kazupon: Can we extend with [vue-i18n-routing.d.ts](https://github.com/nuxt-modules/i18n/blob/next/src/vue-i18n-routing.d.ts)?

src/runtime/utils.ts
```diff
@@ -516,4 +517,17 @@ export function extendBaseUrl<Context extends NuxtApp = NuxtApp>(
   }
 }
 
+export const getLocalePaths = (locale: LocaleOption): string[] => {
```
kazupon: Yeah, this is one of those duplicate codes as your mention...

Nuxt has [the concept of runtime conext and build context](https://nuxt.com/docs/guide/going-further/internals#runtime-context-vs-build-context). we want to be able to use it in both contexts.

One way to solve this, I think have a separate package for use within the nuxt i18n monorepo workspace.

I wish there was a better way to do more code sharing in context... 

@danielroe 
Do you have best solution?

src/types.ts
BobbieGoede: I have changed the types to use the `vue-i18n-routing.d.ts` file, it seems to work for the most part. Typescript gets confused about the `file` property for some reason, so I have to add type assertions.  As seen in the duplicate utility function:
* https://github.com/nuxt-modules/i18n/pull/2357/files#diff-18f829432b971231f742904deac44b9c2b2eb287e72dc2c84684c6050305d700R522
* https://github.com/nuxt-modules/i18n/pull/2357/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R405

src/runtime/utils.ts
BobbieGoede: I managed to remove the duplicate code by preparing the options during generation instead of at runtime

kazupon (approved):
Great works!
Thank you so much!

PR #2363, opened 2023-09-04
FIX protocol inside function getDomainFromLocale with differentDomain:true

karam12:
### ❓ Type of change

### 📚 Description

I fixed the function getDomainFromLocale which sometimes return an URL that begin with `https:://` (that contains 2 ":" instead of 1). This issue happens because the function `URL(window.localtion.origin).protocol` will return the protocol with a ":" at the end. So we don't need to add them again when we return the URL.

### 📝 Checklist

### DIFF
--- a/src/runtime/internal.ts
+++ b/src/runtime/internal.ts
@@ -491,11 +491,11 @@ export function getDomainFromLocale(localeCode: Locale, locales: LocaleObject[],
       const {
         node: { req }
       } = useRequestEvent(nuxt)
-      protocol = req && isHTTPS(req) ? 'https' : 'http'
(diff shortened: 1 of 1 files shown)


kazupon (approved):
you're right!

Thank you so much!

kazupon:
@karam12 
Hi!
I have updated some next branch, because CI is currently not going through (actions/checkout@v4, and timing issue for e2e spec case).
Could you please sync and push them to your branch?

kazupon:
I have tested this fix locally on my own.
It works fine.
I will merge your PR as I do not want to block it for official release.

Thanks!

PR #2402, opened 2023-09-12
fix: disable typescript check for generated `i18n.options.mjs`

### LINKED ISSUE #2392: type error in `.nuxt/i18n.options.mjs` with `checkJs: true` in `tsconfig.json`
### Environment

- Operating System: Linux
- Node Version:     v16.20.0
- Nuxt Version:     3.7.1
- CLI Version:      3.8.0
- Nitro Version:    2.6.3
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      modules
- Runtime Modules:  @nuxtjs/i18n@8.0.0-rc.4
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-xigg7k

run `npm run check` in the console of the above stackblitz repo

### Describe the bug

when `checkJs: true` is enabled in `tsconfig.json`, i get the following type errors when running `vue-tsc`:

```
.nuxt/i18n.options.mjs:1:14 - error TS7005: Variable 'localeCodes' implicitly has an 'any[]' type.

1 export const localeCodes = []
               ~~~~~~~~~~~

.nuxt/i18n.options.mjs:6:46 - error TS7006: Parameter 'context' implicitly has an 'any' type.

6 export const resolveNuxtI18nOptions = async (context) => {
                                               ~~~~~~~

.nuxt/i18n.options.mjs:12:37 - error TS7006: Parameter 'loader' implicitly has an 'any' type.

12  const vueI18nConfigLoader = async (loader) => {
                                       ~~~~~~

.nuxt/i18n.options.mjs:13:48 - error TS7006: Parameter 'r' implicitly has an 'any' type.

13             const config = await loader().then(r => r.default || r)
                                                  ~


Found 4 errors in the same file, starting at: .nuxt/i18n.options.mjs:1


```

### Additional context

_No response_

### Logs

_No response_

BobbieGoede:
### 🔗 Linked issue
#2392 

### ❓ Type of change

### 📚 Description
Resolves #2392 

Adds `// @ts-nocheck` to the start of `i18n.options.mjs` to prevent `checkJs` from checking it, it feels like a lazy fix, but if I understand correctly the end-user should not use this file directly. 

At first I tried adding a declaration file but I don't know how typescript decides whether to use a declaration file to use as types for a `js,mjs` file. As always, I'm open to suggestions!

### 📝 Checklist

### DIFF
--- a/src/gen.ts
+++ b/src/gen.ts
@@ -100,7 +100,7 @@ export function generateLoaderOptions(
     })
   }
 
-  let genCode = ''
+  let genCode = '// @ts-nocheck\n'
   const localeInfo = options.localeInfo || []
   const syncLocaleFiles = new Set<LocaleInfo>()
   const asyncLocaleFiles = new Set<LocaleInfo>()
--- a/test/__snapshots__/gen.test.ts.snap
+++ b/test/__snapshots__/gen.test.ts.snap
@@ -1,7 +1,8 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`basic 1`] = `
-"import locale_en_json_en from \\"../en.json\\" assert { type: \\"json\\" };
+"// @ts-nocheck
+import locale_en_json_en from \\"../en.json\\" assert { type: \\"json\\" };
 import locale_ja_json_ja from \\"../ja.json\\" assert { type: \\"json\\" };
 import locale_fr_json_fr from \\"../fr.json\\" assert { type: \\"json\\" };
 export const localeCodes = [\\"en\\",\\"ja\\",\\"fr\\"]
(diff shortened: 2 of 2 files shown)


kazupon:
Thank you so much!

I was just trying to figure out a solution to this issue.

One is to use your `@ts-ignore` idea, another is to use `addTemplate` / `addTypeTempalte` to generate `i18n.options.d.ts` as mentioned in the documentation.
https://nuxt.com/docs/guide/going-further/modules/#adding-templatesvirtual-files
https://nuxt.com/docs/guide/going-further/modules/#adding-type-declarations

I'm not sure if nuxt i18n can correctly resolve types that way.

@danielroe 
Is it correct to use `addTemplate` / `addTypeTempalte` to generate types?

danielroe:
You could also generate a `i18n.options.d.mts` file with the corresponding types. You would just use `addTemplate` in this case as `addTypeTemplate` is meant for augmentations. However, I think it's reasonable to disable type checking on this file if it will not be imported into user code. I am a bit surprised it's being checked though. Is there anything adding it to the project, such as an import or reference in another file which _is_ meant to be checked?

(Another option could be possibly just generating `i18n.options.ts` file instead of `.mjs`.)

kazupon:
Thank you for your quick reply!

> Is there anything adding it to the project, such as an import or reference in another file which is meant to be checked?

`i18n.options.mjs` is referenced on the runtime side, but we do not provide it as an API, so in effect users cannot use it.
So I think it's reasonable to use `@ts-ignore` to work around it.

PR #2419, opened 2023-09-16
refactor: reduce option generation complexity

BobbieGoede:
### ❓ Type of change

### 📚 Description
I think it would be preferable to use a template for the generated options file instead of constructing it programmatically. This is a step toward that, moving logic outside of the generation script will eventually allow us to replace it with a template.

### 📝 Checklist


kazupon (commented):
Yeah!
i18n.options is very fragile, so it makes sense to generate it in the template.
Your refactoring is the first step towards a healthier nuxt i18n!

kazupon:
Great works!
Thank you! ❤️

PR #2422, opened 2023-09-18
refactor: options template generation

BobbieGoede:
### ❓ Type of change

### 📚 Description
This is PR is built on top of #2419.

This changes the options generation to use a template file, the template still requires some of the options to be formatted/generated to work (see `options.importStrings`, `options.localeMessages` and `options.vueI18nConfigs`). Some of the options in the template are rendered using loops, it doesn't look pretty but the generated file looks readable.

The `gen.test` doesn't test the generated file but only checks the returned options, maybe this could be considered a regression in accuracy? 

### 📝 Checklist

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -7,3 +7,4 @@ coverage
 docs/components/content/Logo.vue
 docs/components/content/VoltaBoard.vue
 .eslintcache
+src/runtime/templates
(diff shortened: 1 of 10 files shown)


kazupon (approved):
Great works!
Thank you so much!

Issue #2443, opened 2023-09-25
Support different domains config at runtime


existe-deja:
### Describe the feature

One build to rule them all!
A good practice is to be able to deliver a project on a staging or prod environment just but changing some environment variables. It would be nice to have the ability to set the related domain for a locale at runtime level.

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution) (The contribution guideline of nuxt-modules/i18n is compliant with Nuxt too).
- [X] Check existing [discussions](https://github.com/nuxt-modules/i18n/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

cjpearson:
I also need this behavior and have a WIP patch. It allows overriding a locale's domain at runtime by setting the `NUXT_PUBLIC_I18N_LOCALES_{locale}_DOMAIN` environment variable.

BobbieGoede:
Resolved by #2446!

existe-deja:
I think I missed something in the setup or maybe it's not working at all.

When I build my project with an certain `.env` file, domain values are not updated when I run it with another `.env` file.

To reproduce:

**1/ edit your `/etc/hosts` with at least the following content to be able to have different local domains**
```
127.0.0.1	localhost local-dev-en local-dev-nl local-dev-ja local-prod-en local-prod-nl local-prod-ja
 ```
 
**2/ clone the project https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-jvph93?file=.template.env and `yarn dev` with the following `.env`**
```
NUXT_PUBLIC_I18N_LOCALES_EN_DOMAIN=local-dev-en:3000
NUXT_PUBLIC_I18N_LOCALES_NL_DOMAIN=local-dev-nl:3000
NUXT_PUBLIC_I18N_LOCALES_JA_DOMAIN=local-dev-ja:3000
```
Everything works fine, the domain is correctly mapped to the locale

**3/ with the same .env file, `yarn build` and then run it with the `.env`**
In Ubuntu I use this command `set -a && source .env && set +a && node .output/server/index.mjs`

Everything works fine as in dev mode.

**4/ update the `.env` file with the following values** 
```
NUXT_PUBLIC_I18N_LOCALES_EN_DOMAIN=local-prod-en:3000
[...]

PR #2441, opened 2023-09-24
docs: host v7 docs

kazupon:
### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

When we release v8 in near the future, we will host on the same domain as the v8 domain (i18n.nuxtjs.org).
This way advantage of hosting as part of v8 docs is that there is no need to maintain a separate domain.

One disadvantage is that in some cases, we will cause confusion for users, to host both versions.

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

Issue #1998, opened 2023-04-14
Test action unreliable - action fails then succeeds without code change


BobbieGoede:
### Environment

-

### Reproduction

Unable to reproduce reliably, first noticed on my PR https://github.com/nuxt-modules/i18n/pull/1997 but behaviour seems to be present as seen in actions triggered by latest commit https://github.com/nuxt-modules/i18n/actions/runs/4697036451

[Test run 1 (failed)](https://github.com/nuxt-modules/i18n/actions/runs/4699507910/attempts/1)
[Test run 2 (succeeded)](https://github.com/nuxt-modules/i18n/actions/runs/4699507910/attempts/2)

### Describe the bug

Noticed tests failed on a small PR and succeeded after rerunning the job without making changes. I'm not sure what the cause is, it may be a race condition or a change in test cleanup upstream. Since it's hard to reproduce it's also hard to pinpoint the cause, I suspect it could be caused by the recent bump in dependency versions.

[Test run 1 (failed)](https://github.com/nuxt-modules/i18n/actions/runs/4699507910/attempts/1)
[Test run 2 (succeeded)](https://github.com/nuxt-modules/i18n/actions/runs/4699507910/attempts/2)

kazupon:
@BobbieGoede 
Can I close this issue?

BobbieGoede:
I think we should keep this issue open, while I have not been able to reproduce it, this issue seems to still be present. As seen here:
* [fix: drop `vueI18n` option warning (succeeded)](https://github.com/nuxt-modules/i18n/actions/runs/5887232980)
* [fix: drop `vueI18n` option warning (failed)](https://github.com/nuxt-modules/i18n/actions/runs/5887298678)

And here as well:
* [docs: tweak docs contents structure (succeeded)](https://github.com/nuxt-modules/i18n/actions/runs/5887636094)
* [docs: tweak docs contents structure (failed)](https://github.com/nuxt-modules/i18n/actions/runs/5887707415)

As far as I can tell there is no code difference between the test runs, the jobs are triggered as they have been merged into `next`. But as demonstrated in this issue description it also happens on PRs before merge.

It's mostly an inconvenience, the tests usually succeed after rerunning, but it does prevent automatic releases and makes the test seem less trustworthy.

PR #2446, opened 2023-09-26
feat: set locale domains via runtimeConfig (#2443)

cjpearson:
### 🔗 Linked issue

#2443 

### ❓ Type of change

### 📚 Description

This change allows setting locale domains via the runtime config. Currently when enabling `differentDomains` the app must be recompiled for each environment because the domains are set at compile time. With this change a single build can be used on different environments with different .env files.

e.g. 

production .env

```
NUXT_PUBLIC_I18N_LOCALES_EN_DOMAIN=en.example.test
NUXT_PUBLIC_I18N_LOCALES_DE_DOMAIN=de.example.test
```

staging .env
```
NUXT_PUBLIC_I18N_LOCALES_EN_DOMAIN=en.staging.example.test
NUXT_PUBLIC_I18N_LOCALES_DE_DOMAIN=de.staging.example.test
```

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/10.different-domains.md
+++ b/docs/content/2.guide/10.different-domains.md
@@ -109,3 +109,19 @@ export default defineNuxtConfig({
   // ...
 })
 ```
+
+With the above config, a build would have to be run for staging and production with different .env files that specify `DOMAIN_UK` and `DOMAIN_FR`.
+
+Alternatively, to avoid the need for multiple builds, the locale domains can be overridden via runtime environment variables. The variable name should follow the format `NUXT_PUBLIC_I18N_LOCALES_{locale code}_DOMAIN`
+
+For example:
+
+```env {}[production .env]
+NUXT_PUBLIC_I18N_LOCALES_UK_DOMAIN=uk.example.test
+NUXT_PUBLIC_I18N_LOCALES_FR_DOMAIN=fr.example.test
+```
+
+```env {}[staging .env]
+NUXT_PUBLIC_I18N_LOCALES_UK_DOMAIN=uk.staging.example.test
+NUXT_PUBLIC_I18N_LOCALES_FR_DOMAIN=fr.staging.example.test
+```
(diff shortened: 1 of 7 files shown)


lukasaric (approved):

cjpearson:
Is this test maybe flaky? I can't reproduce it locally and it also seems to be causing the failures in #2442 and https://github.com/nuxt-modules/i18n/commit/ee5e026c7e9ae02837911e94e424ca29dfa3d7da

BobbieGoede:
@cjpearson 
Yes the tests are a bit flaky in general (tracked here https://github.com/nuxt-modules/i18n/issues/1998), so will be rerunning it. Fortunately it seems to only throw false failures and not the other way around 😅

BobbieGoede:
Thanks for contributing! 

I think this feature looks good, I prefer it over the approach currently described in the docs here: https://v8.i18n.nuxtjs.org/guide/different-domains#runtime-environment-variables

I think if you change/replace the example in the linked documentation to use this feature it would be perfect! What do you think @kazupon?

cjpearson:
Thanks! I made a pass at updating the guide with an example

kazupon:
I think this feature looks good too.
we would like to merge this feature, when you will ready for example and docs :)

cjpearson:
Hi @kazupon I updated the docs for runtime config and routing and added an example to the different domains guide. Is there anything else I should change in the docs?

BobbieGoede:
I have raised the test job timeout, if you rebase your branch on `next` the tests should succeed (at least it won't fail because of the timeout)

BobbieGoede:
Not sure why the tests keep failing repeatedly now, I know that it's not because of your code as they pass locally. Looks like the flaky test debt has caught up with us 😭, will take some time to look into this.

kazupon:
Yeah. 
The Other PRs such as #2441 are also flaky testing… we should fix that issue. 😿

PR #2458, opened 2023-09-30
test: reuse fixtures

BobbieGoede:
### ❓ Type of change

### 📚 Description
Some of the fixtures are basically identical, with the changes in #2457 we should be able to safely reuse fixtures in different tests if they only differ in config overrides. I hope to find some common structures between tests that don't depend on build time configuration, that would allow us to have fewer fixtures and merge some of the test files, potentially making tests faster and more predictable.

### 📝 Checklist


kazupon (approved):
Good job!
Thanks!

PR #2477, opened 2023-10-04
chore: update nuxt2

wattanx:
### ❓ Type of change

### 📚 Description
It must be `nuxt@^2.17.0` to use the nuxt bridge.
I would like to update this as I may add more test cases including nuxt bridge in the future.

### 📝 Checklist


tsconfig.json
```diff
@@ -15,6 +15,7 @@
     "strict": true,
     "noEmit": true,
     "baseUrl": ".",
+    "skipLibCheck": true,
```
rchl: Don't want that here. This can easily hide real issues.

I've investigated the error and it looks like it has to be fixed in `@nuxt/types`. See https://github.com/nuxt/nuxt/pull/23531

package.json
```diff
@@ -145,9 +145,10 @@
     "jest-dev-server": "6.1.1",
     "jsdom": "20.0.0",
     "messageformat": "2.3.0",
-    "nuxt": "2.15.8",
+    "nuxt": "2.17.2",
     "playwright-chromium": "1.24.2",
     "release-it": "15.2.0",
+    "sass": "^1.69.7",
```
wattanx, line 151: Type error for `sass`.
It is difficult and takes time to solve with upstream, so I added sass as a temporary solution.
https://github.com/nuxt/nuxt/pull/24118

rchl:
What is your goal with this? Are you planning to work on this branch to get it to work with bridge or are you trying to use this library in your project with bridge installed?

wattanx:
My goal is to get it to work with nuxt/bridge.

rchl:
Unless something major has changed in nuxt/bridge, I don't think that will be possible without breaking types. But feel free to try anyway :)

In the meantime I've created https://github.com/nuxt-modules/i18n/pull/2661 to update Nuxt 2 and more dependencies since I wasn't able to update your PR.

PR #2478, opened 2023-10-05
feat: add property `bundle.onlyLocales`

imslepov:
### ❓ Type of change

### 📚 Description

Refers to https://github.com/intlify/bundle-tools/pull/301

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -32,7 +32,8 @@ import {
   getPackageManagerType,
   mergeI18nModules,
   applyOptionOverrides,
-  getLocaleFiles
+  getLocaleFiles,
+  filterLocales
 } from './utils'
 import { distDir, runtimeDir, pkgModulesDir } from './dirs'
 import { applyLayerOptions, checkLayerOptions, resolveLayerVueI18nConfigInfo } from './layers'
@@ -126,6 +127,7 @@ export default defineNuxtModule<NuxtI18nOptions>({
      */
 
     applyLayerOptions(options, nuxt)
+    filterLocales(options, nuxt)
     await mergeI18nModules(options, nuxt)
 
     /**
--- a/src/module.ts
+++ b/src/module.ts
@@ -127,8 +127,8 @@ export default defineNuxtModule<NuxtI18nOptions>({
      */
 
     applyLayerOptions(options, nuxt)
-    filterLocales(options, nuxt)
     await mergeI18nModules(options, nuxt)
+    filterLocales(options, nuxt)
 
     /**
      * setup runtime config
--- a/docs/content/3.options/7.bundle.md
+++ b/docs/content/3.options/7.bundle.md
@@ -67,3 +67,7 @@ If you enable this option, **you should check resources in your application are 
 Specify the locales codes that need to be included, the rest will be removed.
 
 It can be useful if you have one code base (e.g. [Nuxt Layers](https://nuxt.com/docs/getting-started/layers)) for several similar projects using different languages.
+
+::alert{type="warning"}
+The value of this **option will not be merged with other Nuxt Layers**. This option should only be specified in the final project config.
+::
--- a/docs/content/3.options/7.bundle.md
+++ b/docs/content/3.options/7.bundle.md
@@ -58,3 +58,12 @@ If you use this option, you need to enable the `compilation.jit` option.
 ::alert{type="warning"}
 If you enable this option, **you should check resources in your application are pre-compiled with nuxt i18n module**. If you will be loading resources dynamically from the back-end via the API, enabling this option does not work because there is no message compiler.
 ::
+
+### `onlyLocales`
+
+- type: `string | string[]`
+- default: `undefined`
+
+Specify the locales codes that need to be included, the rest will be removed.
+
+It can be useful if you have one code base (e.g. [Nuxt Layers](https://nuxt.com/docs/getting-started/layers)) for several similar projects using different languages.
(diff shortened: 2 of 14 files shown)


BobbieGoede:
It would definitely be useful to filter all provided locales like this, unfortunately I think the implementation will be a bit more complicated 😅. 

I have some points:
* It looks like this change will only exclude the messages defined in `<i18n>` blocks, other locales and their messages will still be accessible but will be missing the translations defined in those blocks.
* Currently, if a layer also defines `onlyLocales` it will merge with those defined by the project, say a layer has `['en', 'nl']` configured and the project has `['es']` I believe the result will be `['es', 'en', 'nl']`.
* The tests have to be expanded to confirm the feature works as intended.

imslepov:
@BobbieGoede thanks a lot for the comment!

I have a few thoughts about your points.

- Yes, you're right; the changes affect only messages and locales of the `<i18n>` block. In other places in the project, the developer will need to take care of excluding unused locales, for example, when lazy loading messages.
- From my point of view, it is better to specify the locales only in the project config without merging with layers. So, I think it will be clearer. Maybe even pick locales codes from the `locales` property?

```javascript
defineNuxtConfig({
  i18n: {
    locales: [
      { code: 'en', iso: 'en-US', file: 'en.js', dir: 'ltr' },
      { code: 'ar', iso: 'ar-EG', file: 'ar.js', dir: 'rtl' },
      { code: 'fr', iso: 'fr-FR', file: 'fr.js' }
    ]
  }
})

// `onlyLocales` will be automatically set to ['en', 'ar', 'fr']
```

- About tests - of course, i will do them.

BobbieGoede:
>From my point of view, it is better to specify the locales only in the project config without merging with layers. So, I think it will be clearer. Maybe even pick locales codes from the locales property?

I agree that specifying locales explicitly in the main project would make things clearer. However, merging configuration is the default behavior of Nuxt and projects expect the same behavior for modules. 

This approach would break existing projects, some projects separate the i18n configuration entirely into a layer, or use such a layer as a base and extend its i18n within their project.

>Yes, you're right; the changes affect only messages and locales of the <i18n> block. In other places in the project, the developer will need to take care of excluding unused locales, for example, when lazy loading messages.

Maybe we could move the `includeLocales` property to `i18n.includeLocales` and prevent that property from being merged. It would need to be documented it is a deviation from the default merging behavior. We could use it to filter the result of the merged `locales` array, what do you think?

imslepov:
If I understand you correctly, you want to move the `i18n.bundle.onlyLocales` property to `i18n.includeLocales`. I think this might be a little confusing because I'm not sure where to document the new option 🤔. I propose leaving the option as it is. What do you think?

If you agree with my proposal, then the changes during the PR will be as follows:

```javascript
const layer = defineNuxtConfig({
  i18n: {
    locales: [
      { code: 'en', iso: 'en-US', file: 'en.js', dir: 'ltr' }
    ],
    bundle: {
      onlyLocales: ['en']
    }
  }
})

const project = defineNuxtConfig({
  extends: ['layer'],
  i18n: {
    locales: [
      { code: 'ar', iso: 'ar-EG', file: 'ar.js', dir: 'rtl' },
      { code: 'fr', iso: 'fr-FR', file: 'fr.js' }
    ],
    bundle: {
      onlyLocales: ['fr']
    }
  }
})

// merged locales will be [en, ar, fr]
// but only the value of `bundle.onlyLocales` in the project will have an effect
// the resulting `locales` will only contain "fr"
```

I'll note in the docs about using non-default behavior when merging layers for the `bundle.onlyLocales` property.

BobbieGoede:
>If I understand you correctly, you want to move the i18n.bundle.onlyLocales property to i18n.includeLocales. I think this might be a little confusing because I'm not sure where to document the new option 🤔. I propose leaving the option as it is. What do you think?

Ah yes, I see what you mean, the docs could use some improvements in that regard. Let's keep `i18n.bundle.onlyLocales`, if we (ever) decide to change it, that would be something to consider for `v9`.

>If you agree with my proposal, then the changes during the PR will be as follows:

This looks good!

BobbieGoede:
Filtering the merged locales should be relatively easy after these lines: 

https://github.com/nuxt-modules/i18n/blob/b81801183c7a59f1ae43fce1eab321d693cb0751/src/module.ts#L124-L130

Using `nuxt.options._layers[0]` and the `getLayerI18n` utility function.

imslepov:
I made changes to the code and documentation, added e2e tests, but I encountered difficulties with unit tests. I couldn't figure out how to set them up. If it's necessary to add unit tests, I would appreciate it if you could provide guidance on how to do it.

src/module.ts
```diff
@@ -126,6 +127,7 @@ export default defineNuxtModule<NuxtI18nOptions>({
      */
 
     applyLayerOptions(options, nuxt)
+    filterLocales(options, nuxt)
```
BobbieGoede: It's safer to do this after `await mergeI18nModules(options, nuxt)`, modules could be extending messages of excluded locales, which will make the locales accessible again.

The test should be expanded to test if locales extended by an i18n module are excluded as well. The `layer-module` in playground can be copied for this.

BobbieGoede (changes requested):

src/module.ts
imslepov: Thanks, fixed it :)

BobbieGoede (approved):

BobbieGoede:
Great work, thanks for contributing! 🚀

Issue #2474, opened 2023-10-03
feat: support `nuxt-site-config`


dargmuesli:
### Describe the feature

As discussed on Discord with @BobbieGoede and @harlan-zw, there is Harlan's amazing [site-config](https://nuxtseo.com/site-config/getting-started/background) project which aims to provide a central method for general site configuration so not every other module has to implement this itself.

`nuxt-i18n` allows to specify a `baseUrl`, which is exactly such a reimplementation. I propose to support reading the `i18n.baseUrl` from `site.url` when it is available, keeping the fallback to `i18n.baseUrl`.

Of course `i18n.baseUrl` *could* be marked as deprecated until the final `v8` release if everything works fine outsourcing the `baseUrl` to `nuxt-site-config` to keep a separation of concerns.

### Additional information

- [x] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution) (The contribution guideline of nuxt-modules/i18n is compliant with Nuxt too).
- [X] Check existing [discussions](https://github.com/nuxt-modules/i18n/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

kazupon:
That sounds good! :+1:
nuxt i18n is a community based open-source project as well as Nuxt.

Let's discuss more if we need to work out any details about the specifications! 
of course, PR is welcome! :) /cc @BobbieGoede

RobArkesteijn:
Is there any progression for this? Or at least a workaround until this feature is implemented.

harlan-zw:
Nuxt Site Config integrates with Nuxt I18n which I think makes more sense as Nuxt I18n has way more ways to configure things already. Was there a specific issue you were trying to solve?

I'm going to close this for now as I don't think there's anything to action but happy to troubleshoot with you

PR #2481, opened 2023-10-09
feat(config): add `site.url` as an alternative to `baseUrl`

dargmuesli:
### 🔗 Linked issue

https://github.com/nuxt-modules/i18n/issues/2474

### ❓ Type of change

### 📚 Description

This PR adds support for reading Nuxt config's `site.url`, provided by [Nuxt SEO's site-config module](https://nuxtseo.com/site-config), marking `i18n.baseUrl` as deprecated.

**Draft** because:
- [ ] `nuxt-site-config` does not (yet?) support options as functions, which has to be discussed

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/16.migrating.md
+++ b/docs/content/2.guide/16.migrating.md
@@ -83,9 +83,9 @@ Deprecated for the same reason as the `localeLocation()` Option API.
 
 Use `localeHead()` instead for the Options API style.
 
-### Deprecated `baseUrl` component option
+### `baseUrl` component option
 
-Use `site.url` instead of `i18n.baseUrl`.
+You can use `site.url` instead of `i18n.baseUrl`.
 
 ::alert{type="info"}
 The `site` configuration option stems from [Nuxt SEO's `site-config` module](https://nuxtseo.com/site-config), which `nuxt-i18n` makes use of now.
--- a/docs/content/2.guide/16.migrating.md
+++ b/docs/content/2.guide/16.migrating.md
@@ -16,7 +16,7 @@ You can continue defining `vueI18n` options in `i18n.config`. Refer to the [Vue 
 
 ### Change the route key rules in `pages` option
 
-The key of route set in the `pages` option has been changed to be file-based relative to the `pages/` directory in Nuxt, and **excluding the leading `/`**. 
+The key of route set in the `pages` option has been changed to be file-based relative to the `pages/` directory in Nuxt, and **excluding the leading `/`**.
 
 The reason is that it is more intuitive to match Nuxt file-based routing.
 
@@ -83,6 +83,38 @@ Deprecated for the same reason as the `localeLocation()` Option API.
 
 Use `localeHead()` instead for the Options API style.
 
+### Deprecated `baseUrl` component option
+
+Use `site.url` instead of `i18n.baseUrl`.
+
+::alert{type="info"}
+The `site` configuration option stems from [Nuxt SEO's `site-config` module](https://nuxtseo.com/site-config), which `nuxt-i18n` makes use of now.
+::
+
+Nuxt2:
+```vue {}[nuxt.config.js]
+<script>
+import Vue from 'vue'
+
+export default Vue.extend({
+  nuxtI18n: {
+    baseUrl: 'https://localhost:3000',
+  }
+})
+</script>
+```
+
+Nuxt3:
+```vue {}[nuxt.config.js]
+<script setup>
+defineI18nRoute({
+  site: {
+    url: 'http://localhost:3000'
+  }
+})
+</script>
+```
+
 ### Deprecated `nuxtI18n` component option
 
 Replaced by the `defineI18nRoute()` compiler macro as it can be optimized with bundler.
@@ -122,7 +154,7 @@ Use the `customRoutes` option. because the option name `parsePages` is not intui
    modules: [
      '@nuxtjs/i18n'
    ],
- 
+
    i18n: {
      // ...
 -    parsePages: false,
@@ -141,7 +173,7 @@ Use `dynamicRouteParams` option. Because Vuex is no longer required by the Nuxt 
    modules: [
      '@nuxtjs/i18n'
    ],
- 
+
    i18n: {
      // ...
 -    vuex: true,
--- a/package.json
+++ b/package.json
@@ -88,6 +88,8 @@
     "knitwork": "^1.0.0",
     "magic-string": "^0.30.4",
     "mlly": "^1.4.2",
+    "nuxt-site-config": "^1.4.0",
+    "nuxt-site-config-kit": "^1.4.0",
     "pathe": "^1.1.1",
     "ufo": "^1.3.1",
     "unplugin": "^1.5.0",
--- a/playground/pages/blog.vue
+++ b/playground/pages/blog.vue
@@ -7,10 +7,11 @@ defineI18nRoute({
   }
 })
 
-const { t, locales, baseUrl } = useI18n({
+const { t, locales } = useI18n({
   useScope: 'local'
 })
-console.log('locales', locales, baseUrl)
+const { url } = useSiteConfig()
+console.log('locales', locales, url)
 </script>
 
 <template>
--- a/src/module.ts
+++ b/src/module.ts
@@ -66,14 +66,8 @@ export default defineNuxtModule<NuxtI18nOptions>({
     if (!options.compilation.jit) {
       logger.warn(
         'Opt-out JIT compilation. ' +
-        `It's necessary to pre-compile locale messages that are not managed by the nuxt i18n module (e.g. in the case of importing from a specific URL, you will need to precompile them yourself.) ` +
-        `And also, you need to understand that you cannot support use cases where you dynamically compose locale messages from the back-end via an API.`
-      )
-    }
-
-    if (options.baseUrl && options.baseUrl !== '') {
-      logger.warn(
-        "`i18n.baseUrl` is deprecated in favor of `site.url` as nuxt i18n now makes use of Nuxt SEO's site config."
+          `It's necessary to pre-compile locale messages that are not managed by the nuxt i18n module (e.g. in the case of importing from a specific URL, you will need to precompile them yourself.) ` +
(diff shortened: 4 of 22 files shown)


package.json
```diff
@@ -90,6 +90,8 @@
     "knitwork": "^1.0.0",
     "magic-string": "^0.30.4",
     "mlly": "^1.4.2",
+    "nuxt-site-config": "^1.4.0",
```
dargmuesli, line 91: @harlan-zw https://nuxtseo.com/site-config/getting-started/installation specifies that `nuxt-site-config` has to be installed along with `nuxt-site-config-kit`. As I only ever made use of a `nuxt-site-config-kit` import in this PR and ignoring the playground's `useSiteConfig` usage right now, I wonder: is it really required to install `nuxt-site-config` as well? And if so, why?

playground/pages/blog.vue
```diff
@@ -7,10 +7,11 @@ defineI18nRoute({
   }
 })
 
-const { t, locales, baseUrl } = useI18n({
+const { t, locales } = useI18n({
   useScope: 'local'
 })
-console.log('locales', locales, baseUrl)
+const { url } = useSiteConfig()
```
dargmuesli, line 13: TODO: this should require enabling the `nuxt-site-config` module on the playground.

specs/seo/baseUrl.spec.ts
```diff
@@ -8,8 +8,10 @@ await setup({
   // overrides
   nuxtConfig: {
     i18n: {
-      defaultLocale: 'en',
-      baseUrl: () => {
+      defaultLocale: 'en'
+    },
+    site: {
+      url: () => {
```
dargmuesli: @harlan-zw `nuxt-site-config` currently does not allow options to be set using functions as it is currently supported by `nuxt-i18n`. Are there plans to change this or alternatives to be used?
I'd guess this could be transformed into making use of the `site-config:resolve` hook, right?

BobbieGoede:
While I'm in favor of supporting/using the config module, if we want to add this for `v8` it will have to be backwards compatible. I feel it's important that we make an effort to prevent breaking changes between release candidates as we are working to stable release.

So hopefully there is a way to implement this while keeping compatibility and otherwise we will have to delay this until after `v8` stable release.

dargmuesli:
This change should be backwards compatible already and only make use of `site.url` if available. The only thing a user should see differ is the deprecation notice, which could theoretically be removed and reintroduced before v9.
I haven't done extensive testing yet though as I first want to resolve the threads above I opened myself before doing testing.

src/module.ts
```diff
@@ -72,6 +73,12 @@ export default defineNuxtModule<NuxtI18nOptions>({
       )
     }
 
+    if (options.baseUrl && options.baseUrl !== '') {
```
harlan-zw: I think a JSDoc deprecation should be enough without this. It's not really a problem if people want to keep their i18n config together

src/module.ts
```diff
@@ -128,6 +135,18 @@ export default defineNuxtModule<NuxtI18nOptions>({
...
+    /**
+     * source baseUrl from Nuxt SEO's `site.url`
+     * TODO: remove i18n's `baseUrl` module option when `site.url` has been in use for a sufficiently long time (starting October 2023)
+     */
+
+    await installNuxtSiteConfig()
+    const siteConfig = useSiteConfig()
+
+    if (siteConfig.url) {
+      options.baseUrl = siteConfig.url
```
harlan-zw: Is this needed? We want to use set the baseUrl in a runtime environment so that it can be easily switched

docs/content/2.guide/16.migrating.md
```diff
@@ -83,6 +83,38 @@ Deprecated for the same reason as the `localeLocation()` Option API.
 
 Use `localeHead()` instead for the Options API style.
 
+### Deprecated `baseUrl` component option
+
+Use `site.url` instead of `i18n.baseUrl`.
+
+::alert{type="info"}
+The `site` configuration option stems from [Nuxt SEO's `site-config` module](https://nuxtseo.com/site-config), which `nuxt-i18n` makes use of now.
```
harlan-zw, line 91: I think there will need to be some documentation on all the ways that site config can be set and hosted on this page. The docs aren't that friendly at the moment, something on my to-do list.

docs/content/2.guide/16.migrating.md
```diff
@@ -83,6 +83,38 @@ Deprecated for the same reason as the `localeLocation()` Option API.
 
 Use `localeHead()` instead for the Options API style.
 
+### Deprecated `baseUrl` component option
```
harlan-zw: We don't actually need to deprecate it, both would be valid and we can just push people to use site config instead. If they really want to use baseUrl then they should be free to.

harlan-zw:
@BobbieGoede would you mind sharing what your understanding of the `baseUrl` does? I want to make sure we're covering the test cases and actually implementing the module makes sense.

From my understanding, it's just used for i18n head tags? 

A couple of minor notes until I can do a proper review:
- end user shouldn't have to migrate `baseUrl` to `site.url`, both should be valid - for now (we nudge towards site url for easier consolidation of config)
- worth consolidating the `trailingSlash` config which nuxt-site-config also supports

src/module.ts
dargmuesli: Ah fair. Yeah, currently I only set it right here at build time, that should change.

BobbieGoede:
@dargmuesli 
>This change should be backwards compatible already and only make use of site.url if available. The only thing a user should see differ is the deprecation notice, which could theoretically be removed and reintroduced before v9.

Ah right, my bad. I mistook the deprecation notices as having the original baseURL handling removed for some reason 😅. 

It's good to support setting the value using `nuxt-site-config`, but deprecating `baseUrl` will be something to decide later/separately. For simplicity (for users) I think it is important to keep the ability to configure the module in a single place (the `i18n` key or module array).

>I haven't done extensive testing yet though as I first want to resolve the threads above I opened myself before doing testing.

One thing to consider is layer support, especially when supporting multiple ways of setting options. In the current implementation a layer will override a project's `baseUrl` if the layer uses this new approach. 

@harlan-zw
> From my understanding, it's just used for i18n head tags?

As I understand it, this is correct. Most of this happens in [`vue-i18n-routing`](https://github.com/intlify/routing/tree/main/packages/vue-i18n-routing), which uses the config to set up the head tags*. 

It is possible to override `baseUrl` by setting `domain` on locales, or by using a function to set it based on the Nuxt Context as [seen here](https://github.com/nuxt-modules/i18n/blob/b81801183c7a59f1ae43fce1eab321d693cb0751/src/runtime/utils.ts#L484C26-L484C26). When setting `domain` on locales, it is possible to override these [using runtime config](https://i18n.nuxtjs.org/guide/different-domains#runtime-environment-variables) as well. These options exist to allow projects to support different languages using different domains, I'm not sure how this would interact with `nuxt-site-config` or when other modules rely on it as well.

*On a separate note, I've been wanting to look into implementing/replacing the head tag handling in `vue-i18n-routing` with `unhead`. Unfortunately right now I'm not yet familiar enough with either of these packages to do this, if you have time to give it a look and/or point me in the right direction, that would be greatly appreciated! 😅

kazupon:
Sorry, I was preparing for vue fes japan, so I could not track of what was going on with this issue. :sweat_smile: 
How is going this PR?

dargmuesli:
It's on my TODO list, but as it should be a backwards compatible feature addition I had other tasks with higher priority. I think I can continue to work on it this week.

harlan-zw:
I also need to re-review this properly when I have some time

src/module.ts
dargmuesli: How would I add the JSDoc as that's [picked from `vue-i18n-routing`](https://github.com/nuxt-modules/i18n/blob/3c75cf91ec04909d93b00dda62393c0dc59a508e/src/types.ts#L115)?

src/module.ts
dargmuesli: In ea4a900dba9f48452360dbe8668cb37f07cd0d7e I made that change. There could be a better place of course as @BobbieGoede pointed out, but I don't feel knowledgeable enough to decide on that.

docs/content/2.guide/16.migrating.md
dargmuesli: I removed the deprecations.

src/module.ts
BobbieGoede: > How would I add the JSDoc as that's [picked from `vue-i18n-routing`](https://github.com/nuxt-modules/i18n/blob/3c75cf91ec04909d93b00dda62393c0dc59a508e/src/types.ts#L115)?

Something like this is possible, I think the original JSDoc would need to be copied and extended, unless there is some way to append to existing JSDoc that I'm unaware of 😅.

```diff
+  /**
+   * Custom JSDoc here!
+   */
+  baseUrl?: I18nRoutingOptions<Context>['baseUrl']
} & Pick<
  I18nRoutingOptions<Context>,
- | 'baseUrl'
```

dargmuesli (commented):
I've removed the deprecations, added separate `site.url` tests besides the `baseUrl` tests and moved the site config to the runtime. That should form the groundwork I intended to lay. For further changes I'd need assistance from you people :heart: 

Be invited to give examples on what needs to change or to submit PRs against this PR's base branch in my fork to get site config integrated into nuxt-i18n! :muscle: :partying_face:

src/module.ts
dargmuesli: Added :rocket:

BobbieGoede:
A lot has happened since this PR, I tried rebasing but I don't want to risk losing essential changes with a force push 😅 

Could you rebase/update your PR, we can keep the following
* dependency changes
* updated docs
* option setting 
* added tests
  *  let's keep the tests that use the old `baseUrl` config

dargmuesli:
Nuxt site config reads from i18n (https://nuxtseo.com/docs/site-config/guides/i18n) as noted in https://github.com/nuxt-modules/i18n/issues/2474#issuecomment-3032181364, so apparently it should be sufficient to specify `baseUrl` and have `site.url` adapt.

PR #2495, opened 2023-10-16
fix: handle special characters in route

### LINKED ISSUE #2382: Unjustified "Error: Avoided redundant navigation to current location"
### Environment

- Operating System: Windows_NT
- Node Version:     v20.5.0
- Nuxt Version:     3.7.1
- CLI Version:      3.7.3
- Nitro Version:    2.6.2
- Package Manager:  pnpm@8.6.10
- Builder:          -
- User Config:      ssr, modules, i18n
- Runtime Modules:  @nuxtjs/i18n@8.0.0-rc.4
- Build Modules:    -

### Reproduction

https://github.com/some-user123/nuxt-i18n-issue/

Run
```
pnpm install
pnpm dev
```

### Describe the bug

Preconditions:
- `ssr: false`
- `@nuxtjs/i18n` installed with `strategy: 'no_prefix'`

If you start above reproduction and point your browser to `http://localhost:3000/` you will be redirected to `http://localhost:3000/login?ret=%2F` due to a failing API request in the composable. (That redirect seems crucial. If you browse to this URL directly, the error will not occur. Similarly the error does not occur if the redirect goes to a URL without query parameter).

If you then click on the button, this code will be executed:
```js
  console.log('navigating to page2');
  const ret = await navigateTo({ path: '/page2', query: { q: 'foo' } });
  console.log('done', ret);
```

and leads to this output:
image

I think, this is wrong, as the navigation should go to `/page2?q=foo`, which is different from the current location `/login?ret=%2F`.

I suppose this is related to `@nuxtjs/i18n` as the error does not occur, if the module is not enabled.

### Additional context

_No response_

### Logs

_No response_

Achneoder:
### ❓ Type of change

### 📚 Description

Resolves #2382

Encodes the routePath via `encodeURI` for proper comaprison with `route.from.fullPath` 

### 📝 Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ git checkout -b my-new-feature
 ```sh
 pnpm fix
 pnpm test:unit
-pnpm test:e2e
+pnpm test:spec
 ```
 
 - Commit and push your changes
--- a/specs/fixtures/issues/2382/app.vue
+++ b/specs/fixtures/issues/2382/app.vue
@@ -1,0 +1,3 @@
+<template>
+  <NuxtPage />
+</template>
--- a/specs/fixtures/issues/2382/locales/en-US.json
+++ b/specs/fixtures/issues/2382/locales/en-US.json
@@ -1,0 +1,15 @@
+{
+  "Meta": {
+    "locale": "English"
+  },
+  "Pages": {
+    "Home": {
+      "title": "Home page",
+      "description": "Home page description"
+    },
+    "About": {
+      "title": "About page",
+      "description": "About page description"
+    }
+  }
+}
--- a/specs/fixtures/issues/2382/nuxt.config.ts
+++ b/specs/fixtures/issues/2382/nuxt.config.ts
@@ -1,0 +1,19 @@
+import type { LocaleObject } from '#i18n'
+
+const locales = [{ code: 'en', iso: 'en-US', file: 'en-US.json' }] as LocaleObject[]
+
+const defaultLocale = locales[0]
+
+export default defineNuxtConfig({
+  modules: ['@nuxtjs/i18n'],
+  i18n: {
+    locales,
+    defaultLocale: defaultLocale.code,
+    langDir: 'locales/',
+    lazy: true,
+    strategy: 'no_prefix',
+    detectBrowserLanguage: {
+      fallbackLocale: defaultLocale.code
+    }
+  }
+})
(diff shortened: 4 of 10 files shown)


Achneoder:
I'm probably missing something but the tests seem flaky. Executing my test locally works just fine, running the whole test suite results in failed tests, sometimes on of mine, sometimes others.

I merged my four individual tests to one with the same result.
I also tried to shard the execution of the specs, which results in sometimes in timeouts or just failures of random tests:
See https://github.com/Achneoder/nuxt-i18n/actions/runs/6545610627/attempts/1 - Attempt 1 was successful, Attempt 2 (just a re-run) failed

BobbieGoede:
Thanks for contributing!

You're right the tests are flaky, still trying to figure out the exact reason why... 😅 So far I have had some success in reducing the flakiness by adding `waitForURL` after navigations, which seems to have done the trick here too!

As for the issue this PR is fixing, as you mention in the comment note, this is most likely an issue in `vue-i18n-routing`. I tried to look into fixing it there after seeing your solution, but can't quite find the root cause (it uses `router.resolve` which I believe should encode the route as well?) or proof that the issue happens there in the first place. 

I think we could merge this PR so at least the issue is resolved in `@nuxtjs/i18n`, and open an issue on `vue-i18n-routing` to track if and when that gets solved, after which we can remove the fix you added as ideally this is fixed upstream. What do you think @kazupon?

Achneoder:
Cool! Thanks for your help!

I also tried to track down the origin of the path encoding issue and ended in `vue-router`. I hope to have a pull request ready soon there

BobbieGoede:
>I also tried to track down the origin of the path encoding issue and ended in vue-router. I hope to have a pull request ready soon there

Hmm, I suggest opening an issue there if it doesn't exist yet. If this behaviour stems from `vue-router` I'm curious whether maybe it is intentionally left unencoded. 🤔

Achneoder:
Apparently `vue-router` is working as intended:
https://github.com/vuejs/router/issues/2021

Maybe it's worth to check `vue-i18n-routing` again and apply the fix there, but then there's the same question: is `vue-i18n-rouing` responsible to provide properly encoded URLs when calling `switchLocalePath` (and also probably `localePath`)

kazupon:
@BobbieGoede 
> I think we could merge this PR so at least the issue is resolved in @nuxtjs/i18n, and open an issue on vue-i18n-routing to track if and when that gets solved, after which we can remove the fix you added as ideally this is fixed upstream. What do you think @kazupon?

Yeah, this issue should be opened in `vue-i18n-routing`.
You can open in there.

kazupon:
I've released vue-i18n-routing v1.1.2
https://github.com/intlify/routing/releases/tag/v1.1.2

You can bump nuxt i18n deps.

PR #2505, opened 2023-10-18
fix: prevent strategy `no_prefix` redirect detection locale change attempt

### LINKED ISSUE #2493: This module breaks ":to" links in nuxt v3
### Environment

- Operating System: Linux
- Node Version:     v18.17.1
- Nuxt Version:     3.7.4
- CLI Version:      3.9.0
- Nitro Version:    2.6.3
- Package Manager:  yarn@1.22.19
- Builder:          -
- User Config:      devtools, modules
- Runtime Modules:  @nuxtjs/i18n@8.0.0-rc.5
- Build Modules:    -

### Reproduction

https://github.com/florealcab/NuxtI18nBug

### Describe the bug

If you are in a page, and the URL ends with a slash (example: /sub/toto/), then any link on the page with the "to" parameter has no more effect (example : <NuxtLink to="/">Index</NuxtLink>) : click action is ignored, nothing in logs client side or server side.

If you just remove the trailing slash, all works well.

To reproduce it with the repo I give:
Go to http://localhost:3000/sub/test/ and try to click on the link "Go to index". Nothing happen. No logs.
Now, just remove the trailing slash (url become http://localhost:3000/sub/test). Click to the link "Go to index" again: it works!

### Additional context

This bug appears with the 8.0.0-rc3 version and does not exists before (tested with 8.0.0-beta10, rc1 and rc2), and always after (tested until rc5 which is the latest version for now.

This bug seems to be related to this module. I install all and then just downgrade or upgrade this module. I try to use the rc5 version and manually downgrade vue-i18n and vue-i18n-routing to the same version they had with the 8.0.0-beta10 of this module: the bug is still here.

But I don't know what causes this on this module, by the way...

### Logs

_No response_

BobbieGoede:
### 🔗 Linked issue
#2493 

### ❓ Type of change

### 📚 Description
Resolves #2493 

It looks like `detectRedirect` tried to change locale based on route while using `no_prefix` strategy. This causes issues as we can't retrieve the locale from unprefixed paths.

This resolves the referenced issue but potentially a related/underlying issue remains. When resolving the `routePath` here: https://github.com/nuxt-modules/i18n/blob/main/src/runtime/utils.ts#L315-L323 on url with a trailing slash will cause a redirect as `routePath` will resolve to the same path without a trailing slash. In the case of the referenced issue this logic is never hit as the locale does not change. 

### 📝 Checklist

### DIFF
--- a/specs/fixtures/basic/pages/category/[slug].vue
+++ b/specs/fixtures/basic/pages/category/[slug].vue
@@ -1,3 +1,4 @@
 <template>
   <p>This is cateory page on '{{ $route.params.slug }}'</p>
+  <NuxtLinkLocale id="return-home-link" to="/">Home</NuxtLinkLocale>
 </template>
--- a/specs/routing_strategies/no_prefix.spec.ts
+++ b/specs/routing_strategies/no_prefix.spec.ts
@@ -78,4 +78,24 @@ describe('strategy: no_prefix', async () => {
     // current locale
     expect(await getText(page, '#lang-switcher-current-locale code')).toEqual('fr')
   })
+
+  test('(#2493) should navigate from url with and without trailing slash', async () => {
+    const page = await createPage()
+    await page.goto(url('/category/nested/'))
+    await page.waitForURL('**/category/nested/')
+
+    await page.locator('#return-home-link').click()
+
+    await page.waitForURL(/:[0-9]+\/$/)
+
+    expect(page.url()).toEqual(url('/'))
+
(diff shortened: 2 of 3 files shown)


kazupon (approved):
Good job!
I'm happy to you worked on routing.
nuxt i18n routing is difficult that has some strategies and complex.
LGTM!

MaxCpp:
Sorry, when this fix release?

PR #2536, opened 2023-11-05
feat(package): add composables export

dargmuesli:
### 🔗 Linked issue

n/a

### ❓ Type of change

### 📚 Description

Allows to `import { useLocalePath } from '@nuxtjs/i18n/composables'` in addition to the already existing possibility to `import { useLocalePath } from '#i18n'`.

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,10 @@
       "require": "./dist/module.cjs",
       "types": "./dist/types.d.ts"
     },
+    "./composables": {
+      "import": "./dist/runtime/composables.mjs",
+      "types": "./dist/runtime/composables.d.ts"
+    },
     "./package.json": "./package.json"
   },
   "imports": {


dargmuesli:
Feel free to tell me if this is an anti-pattern.

kazupon:
Hi!
Thank you for your PR!

In my opinion, we don’t need export composable with `exports` field of `package.json`.
Because, there are no requests from communities other than nuxt to use composable. And we can import composables with `#i18n`.

nuxt does default auto import, so if you don't use auto import, we can import composable using `#i18n`.
(Oops, no docs for `#i18n` importing  😅)

dargmuesli:
True, I thought about this because e.g. nuxt also [has `./app` exported](https://github.com/nuxt/nuxt/blob/38b6d88cfab3477edabfa37d6f97b95eefc6ae6c/packages/nuxt/package.json#L31) while providing '#app' at the same time.
@danielroe is one of those identifiers to be preferred or recommended? :bulb:

danielroe:
I would recommend importing from `#imports` instead of either of them.

PR #2537, opened 2023-11-06
ci: use node version 18.18+

BobbieGoede:
### ❓ Type of change

### 📚 Description
After a long search I finally figured out why tests failed using Node version `18.18.2` or higher. Our tests set the `Host` header to test server responses for different domains, but due to https://github.com/nodejs/undici/pull/2322 this broke as `fetch` is not allowed to set the `Host` header for security reasons. 

I have added `undici` as a dev dependency and used it in tests where this behaviour is needed.

### 📝 Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


kazupon (approved):
Nice work!

Issue #2131, opened 2023-06-02
With strategy `prefix`, unprefixed routes are accessible


huang-julien:
### Environment

------------------------------
- Operating System: Windows_NT
- Node Version:     v16.20.0
- Nuxt Version:     3.5.2
- Nitro Version:    2.4.1
- Package Manager:  yarn@1.22.19
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/nuxt-starter-vdw3ue?file=pages%2Ftest.vue,pages%2Findex.vue,package.json

### Describe the bug

When using `prefix` strategy, we can see that the unprefixed route is still registered within the router.
When going to the unregistered route, we can still access to it without being redirected.

![image](https://github.com/nuxt-modules/i18n/assets/63512348/22422bb8-43cb-4117-b97a-e03e0cccab48)

kazupon:
Thank you for your reporting!
I could not reproduce this issue.
I got the below error:

```sh
 ERROR  Error while requiring module /path/to/2132/modules/routing/index.vue: SyntaxError: Unexpected token '<'                                  14:44:50


 ERROR  Cannot start nuxt:  Unexpected token '<'                                                                                                                                             14:44:50

  <template>
  ^

  SyntaxError: Unexpected token '<'
  at internalCompileFunction (node:internal/vm:73:18)
  at wrapSafe (node:internal/modules/cjs/loader:1176:20)
  at Module._compile (node:internal/modules/cjs/loader:1218:27)
  at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
  at Module.load (node:internal/modules/cjs/loader:1117:32)
  at Module._load (node:internal/modules/cjs/loader:958:12)
  at Module.require (node:internal/modules/cjs/loader:1141:19)
  at require (node:internal/modules/cjs/helpers:110:18)
  at jiti (node_modules/jiti/dist/jiti.js:1:243592)
  at requireModule (node_modules/@nuxt/kit/dist/index.mjs:274:26)

```

I seem your reproduction is updated after your opened this issue.

Could you give us your reproduction again please?  🙏

huang-julien:
oops it should be updated now 😅 . To reproduce it, you can simply make a hard reload on `/test`. The full path will be shown

kazupon:
Thank you for your updating!
I could confirm this issue.

We will try to fix.

[...]

Issue #2524, opened 2023-10-25
Inconsistent handling of `Accept-Language` header


rudolfbyker:
## Updates

- 2024/01/02: Updated reproduction and steps to reproduce at https://github.com/nuxt-modules/i18n/issues/2524#issuecomment-1873823456
- 2025/05/02: Updated reproduction for v9.5.4: https://github.com/nuxt-modules/i18n/issues/2524#issuecomment-2847024120

### Environment

Working directory: /home/projects/aoniwxvlm.github                                                                       13:55:30
Nuxt project info:                                                                                                       13:55:30

------------------------------
- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.8.0
- CLI Version:      3.9.1
- Nitro Version:    2.7.0
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools, modules, i18n
- Runtime Modules:  @nuxtjs/i18n@8.0.0-rc.5
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/github-fyaxh8

### Describe the bug

The root redirect takes the `Accept-Language` header into account:

```
curl -H "Accept-Language: fr" -I http://localhost:3000
```

```
HTTP/1.1 302 Found
access-control-allow-origin: *
location: /fr
content-type: text/html
date: Wed, 25 Oct 2023 11:49:52 GMT
connection: close
```

But everything else does not:

```
curl -H "Accept-Language: fr" -I http://localhost:3000/foo
```

```
HTTP/1.1 302 Found
access-control-allow-origin: *
location: /en/foo
content-type: text/html
date: Wed, 25 Oct 2023 11:49:41 GMT
connection: close
```

This also relates to #2131 because I'm using the `prefix` strategy, which means I should actually get a 404 on the `/foo` request. Therefore, I think fixing #2131 will also fix this issue, but not vice versa.

BobbieGoede:
To redirect using cookies or the `accept-language` header on routes other than the root route, you need to set `detectBrowserLanguage.redirectOn` to `'all'`. In which case `curl -H "Accept-Language: fr" -I http://localhost:3000/foo` will redirect to `http://localhost:3000/fr/foo`. Does this fit your use case?

Disabling browser language detection has some other behavior that seems a bit unintuitive with the `prefix` strategy, it still redirects to the default language (so technically no detection), I suppose it makes it possible to still render a catchall/404 page instead of returning a plain 404 response.

[...]

PR #2538, opened 2023-11-06
fix: exclude unprefixed routes for strategy `prefix`

BobbieGoede:
### 🔗 Linked issue
* #2524
* #2131

### ❓ Type of change

### 📚 Description
@kazupon 
What is the intended behavior for projects with `redirectOn: 'root'` when accessing an unprefixed (non existant in this case after this change) route like `/about`, should it return a 404? Right now this still redirects to the default language route, relevant issue: #2524.

Let me know if there are specific cases are missing tests for this change!

### 📝 Checklist

### DIFF
--- a/specs/routing_strategies/prefix.spec.ts
+++ b/specs/routing_strategies/prefix.spec.ts
@@ -1,6 +1,6 @@
 import { describe, test, expect } from 'vitest'
 import { fileURLToPath } from 'node:url'
-import { setup, url } from '../utils'
+import { setup, url, fetch } from '../utils'
 import { getText, getData, renderPage, waitForURL } from '../helper'
 
 import type { Response } from 'playwright'
@@ -18,17 +18,14 @@ await setup({
 })
 
 describe('strategy: prefix', async () => {
-  test.todo('cannot access to no prefix url: /', async () => {
-    const home = url('/')
-    const { page } = await renderPage(home)
-    let res: Response | (Error & { status: () => number }) | null = null
-    try {
-      res = await page.goto(home)
-    } catch (error: unknown) {
-      res = error as Error & { status: () => number }
-    }
-    // 404
-    expect(res!.status()).toBe(404) // eslint-disable-line @typescript-eslint/no-non-null-assertion
+  test.each([
+    ['/', '/en'],
+    ['/about', '/en/about'],
+    ['/category/foo', '/en/category/foo']
+  ])('cannot access unprefixed url: %s', async (pathUrl, destination) => {
+    const res = await fetch(pathUrl, { redirect: 'manual' })
+    expect(res.status).toBe(302)
+    expect(res.headers.get('location')).toBe(destination)
   })
 
   test('can access to prefix locale: /en', async () => {
--- a/src/pages.ts
+++ b/src/pages.ts
@@ -45,7 +45,7 @@ export function setupPages(
   let includeUprefixedFallback = nuxt.options.ssr === false
   nuxt.hook('nitro:init', () => {
     debug('enable includeUprefixedFallback')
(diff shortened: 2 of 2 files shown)


kazupon:
>What is the intended behavior for projects with redirectOn: 'root' when accessing an unprefixed (non existant in this case after this change) route like /about, should it return a 404? Right now this still redirects to the default language route

as your mention,redirects to the default language route. this befavior is from nuxt-i18n v7.
we should be fixed.

divine:
Hello,

Does it handle correct redirection from `/about` to `/fr/about` with `Accept Language` or `Cookies` already set?

Thanks!

kazupon:
> Does it handle correct redirection from /about to /fr/about with Accept Language or Cookies already set?

Thanks feedback!

@BobbieGoede 
From this feedback, we need a test with a pattern of cases for cookies and `navigator.languages`.

BobbieGoede:
@divine 
> Does it handle correct redirection from `/about` to `/fr/about` with `Accept Language` or `Cookies` already set?

You need to set `detectBrowserLanguage.redirectOn` to `'all'` for this to work, this should already work work without this PR. Relevant https://github.com/nuxt-modules/i18n/issues/2524#issuecomment-1801838427.

@kazupon 
> From this feedback, we need a test with a pattern of cases for cookies and navigator.languages.

As described above this seems to already work with the correct configuration. 

> as your mention,redirects to the default language route. this befavior is from nuxt-i18n v7.

After giving it more thought, I suppose it makes sense to redirect to a localized catch all route rather than responding with a plain 404 response. Hmm 🤔 not entirely sure what would be the best approach. I believe this PR should at least resolve #2131.

divine:
> You need to set detectBrowserLanguage.redirectOn to 'all' for this to work, this should already work work without this PR. Relevant https://github.com/nuxt-modules/i18n/issues/2524#issuecomment-1801838427.

I'll check but just make sure this won't create a regression later.

> After giving it more thought, I suppose it makes sense to redirect to a localized catch all route rather than responding with a plain 404 response.

Returning 404 doesn't make sense, it's like returning 404 error code on the root path `/` then redirect to `/en/`.

You might be surprised but the old version had this problem and a lot of pages were duplicated in search engines. The reason behind this is that i18n module for Nuxt didn't handle correctly SEO. So redirecting to the correct page according to the language header or `fallback` locale it's what should resolve the duplicate problem.

Only in 2020 SEO feature was integrated correctly, I'm not sure why my contribution silently lost but there are PRs with the reasoning and explanation behind it so I'd expect at least this should work like the previous version which logic was fixed.

Thanks!

rudolfbyker:
> I believe this PR should at least resolve https://github.com/nuxt-modules/i18n/issues/2131.

I tested this, and the unprefixed routes are still available. ~~In fact, I can't see any difference after applying the changes from this PR.~~ EDIT: The unprefixed routes are gone, but for some reason I still get redirected from `/foo` to `/en/foo` instead of the expected 404 on `/foo` (since `foo` is not a language).

BobbieGoede:
@rudolfbyker 
You probably have a catch all route set up, I think removing that would result in 404 response on these routes. Maybe as a workaround you could detect redirects to the catchall page, and ensure a 404 response if this is from a nonexistent page?

rudolfbyker:
I double checked, and I definitely do not have a catch-all route. See my updated example with explanation at https://github.com/nuxt-modules/i18n/issues/2524#issuecomment-1824120507 . Thanks for your time.

rudolfbyker:
This PR is good. It removes the unprefixed routes as intended. #2524 seems to be a separate issue.

kazupon (approved):

PR #2545, opened 2023-11-08
feat: add `skipDefaultLocale` options

s00d:
### ❓ Type of change

- [X] 📖 Documentation (updates to the documentation or readme)
- [X] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

~~removed the default locale from loadInitialMessages. Why is this needed? There's a FallbackLocale, if needed, one can add it themselves, but now it works weirdly, always loading the default locale, even if it's not needed.~~

I added the skipDefaultLocale option, which allows you to manually control the list of available locales.

### 📝 Checklist

- [X] I have updated the documentation accordingly.

### DIFF
--- a/docs/content/3.options/10.misc.md
+++ b/docs/content/3.options/10.misc.md
@@ -55,3 +55,10 @@ Don't enable this option for use in production. Performance will be decreased.
 - default: `false`
 
 Set the plugin as `parallel`. See [nuxt plugin loading strategy](https://nuxt.com/docs/guide/directory-structure/plugins#loading-strategy).
+
+## `skipDefaultLocale`
+
+- type: `boolean`
+- default: `false`
+
+If `true`, the `DefaultLocale` will be skipped. By default, the `DefaultLocale` is added and loaded every time.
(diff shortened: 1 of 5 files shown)


kazupon:
Thank you for contribution!

nuxt i18n's loadInitialMessages first dynamically loads the required locale. The `defaultLocale` is an option in nuxt i18n. It is necessary for correct translation.
Removing `defaultLocale` will not work correctly in nuxt, which has some rendering modes.

s00d:
> Thank you for contribution!
> 
> nuxt i18n's loadInitialMessages first dynamically loads the required locale. The `defaultLocale` is an option in nuxt i18n. It is necessary for correct translation. Removing `defaultLocale` will not work correctly in nuxt, which has some rendering modes.

Could you tell me for which modes is this required? I will improve the logic, it is bad now.

or I can suggest a new parameter that would allow manually controlling the loading of the default, so that it could be turned off.

s00d:
Can we do it like this? I added the option skipDefaultLocale and now, if needed, you can manually control  the list of locales.

PR #2566, opened 2023-11-22
refactor: remove aliases

BobbieGoede:
### ❓ Type of change

### 📚 Description
@kazupon 
It seems like this works as expected when using normal imports, was there a specific reason these were imported like this? 

### 📝 Checklist


kazupon:
@BobbieGoede 
When I started developing the Nuxt i18n module v8, importing from a template-generated module in the form of another relative path didn't seem to work out, so I have a feeling it was solved that way, but I forget the reason. 😅 

The test passes with the normal import style, so it looks fine. 
Thanks for refactoring as always!

kazupon (approved):

Issue #2429, opened 2023-09-20
Dynamic Routes in Nuxt 3


mawiswiss:
### Describe the feature

Hey there,

I don't know if for my issue there's already a working solution. At least I don't get it to work for now.
I'm now working with Nuxt 3 (@nuxtjs/i18n: 8.0.0-rc.4) and in the process of upgrading from Nuxt 2.

I have dynamic routes with a slug/url that is translated and changes per language.
Before I was able to dispatch those params like this:
```
await store.dispatch('i18n/setRouteParams', {
      en: { postId: 'my-post' },
      fr: { postId: 'mon-article' }
    })
```
In the new version, I don't see such an option. Also since VueX also isn't the default (I'm using Pinia now as well). But how is it possible to push the slug to the route somehow and the package is resolving it with `switchLocalePath`?

In the [documentation](https://i18n.nuxtjs.org/custom-paths/#in-component-options) there's something mentioned about "Custom route paths", but it still doesn't make sense to me how to add those dynamic values.

Something like that and then the second part the actual `name` param is missing to me.
```
export default {
  nuxtI18n: {
    paths: {
      en: '/articles/:name',
      es: '/artículo/:name'
    }
  }
}
```

Thanks and would appreciate it if somebody has experience with this feature.

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution) (The contribution guideline of nuxt-modules/i18n is compliant with Nuxt too).
- [X] Check existing [discussions](https://github.com/nuxt-modules/i18n/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

TomKremer:
Hi,
I got the same issue. The current doc explains using definePageMeta (https://v8.i18n.nuxtjs.org/guide/lang-switcher) but the issue is, this function cannot be used async with loading content which includes the language specific slugs or uids.

Also an Idea would be to change the copy of the doc from:

> Be sure you have not set the dynamicRouteParams option to false in Nuxt i18n module's options.

_to possible version:_
To make dynamic route parameters work, set the dynamicRouteParams option to true in Nuxt i18n module's options.
(because the default is false)

Best,
Tom

mawiswiss:
While your point is correct and the dynamicRouteParams needs to be set correctly (to true), this isn't the point of my ticket.

[...]

Issue #2006, opened 2023-04-18
Dynamic route parameters do not work with dynamic values


clem-zezuka:
### Environment

- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.4.1
- Nitro Version:    2.3.3
- Package Manager:  npm@7.17.0
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-njnv2g?file=pages/blog/[slug].vue

### Describe the bug

Hello!

According to this [page](https://v8.i18n.nuxtjs.org/guide/lang-switcher#dynamic-route-parameters) of the documentation, to set translations for route params we must use `definePageMeta`.

However, `definePageMeta` cannot reference the component or values defined on it.

To be crystal clear, variables cannot be injected inside this function and this makes the feature impossible to use at this stage.

It seems like a regression compare to the function `i18n/setRouteParams` of the version of the module on Nuxt 2 and the documentation is not clear about it.

Is it something that you plan to fix in the future?

Thanks for reading me!

### Logs

```
[nuxt] [request error] [unhandled] [500] routes is not defined
  at eval (./pages/blog/[slug].vue:3:5)  
  at ViteNodeRunner.directRequest (file://./node_modules/vite-node/dist/client.mjs:348:11)  
  at async ViteNodeRunner.cachedRequest (file://./node_modules/vite-node/dist/client.mjs:188:14)  
  at async ViteNodeRunner.dependencyRequest (file://./node_modules/vite-node/dist/client.mjs:237:14)  
  at async eval (./virtual:nuxt:/home/projects/github-njnv2g/.nuxt/routes.mjs:1:216)  
  at async ViteNodeRunner.directRequest (file://./node_modules/vite-node/dist/client.mjs:348:5)  
  at async ViteNodeRunner.cachedRequest (file://./node_modules/vite-node/dist/client.mjs:188:14)  
  at async ViteNodeRunner.dependencyRequest (file://./node_modules/vite-node/dist/client.mjs:237:14)  
  at async eval (./node_modules/nuxt/dist/pages/runtime/plugins/router.js:19:31)  
  at async ViteNodeRunner.directRequest (file://./node_modules/vite-node/dist/client.mjs:348:5)
```

vladyslavprosolupov:
Had the same issue lately. The temporary solution that I came up with is tricky/messy and doesn't work on server-side.

[...]

Issue #1736, opened 2022-12-15
Feature: Dynamic route parameters should not be set using definePageMeta


kazupon:
### Discussed in https://github.com/nuxt-modules/i18n/discussions/1730

<div type='discussions-op-text'>

<sup>Originally posted by **adesombergh** December 13, 2022</sup>
Hi, thanks for the great work with this module.

I stumbled into a problem with i18n module v8 and nuxt3 and I think it might be interesting to change this in the future, as it makes it difficult to use with dynamic content.

Please, correct me if i'm not doing in right.

### The Problem

According to the [Dynamic route parameters documentation](https://v8.i18n.nuxtjs.org/guide/lang-switcher#dynamic-route-parameters), to set translations for route params we should use `definePageMeta` like so:
```vue
<script setup>
definePageMeta({
  // ...
  nuxtI18n: {
    en: { id: 'my-post' },
    fr: { id: 'mon-article' }
  }
  // ...
})
</script>
```

However, the Nuxt 3 documentation about `definePageMeta` ([right here](https://nuxt.com/docs/guide/directory-structure/pages/#page-metadata:~:text=Much%20like%20defineEmits,reference%20imported%20bindings.)) says that **the page meta object cannot reference the component (or values defined on the component)**.

It means that the content inside `definePageMeta({})` is supposed to be **hard-coded and cannot be dynamic**, making it impossible to set the page route params dynamically (as they often are). The following code won't work:

```vue
<script setup>
// /pages/[slug].vue
const route = useRoute()
const { locale } = useI18n()
const { data } = await useFetch(
  `https://yourapi.com/${locale}/page/${route.params.slug}`
)

definePageMeta({ 
  nuxtI18n: data.value.other_languages // Throws an error because data is undefined
})
</script>
```

```json
{
  "other_languages": {
    "en": { "slug": "tech-specs"},
    "fr": { "slug": "specifications-techniques"},
  },
  ...more content ...
}
```

### A solution

Instead of using `definePageMeta`, maybe i18n module could give us a composable like `setParamsTranslations` or whatever,  that could be used like this:

```vue
<script setup>
// /pages/[slug].vue
const route = useRoute()
const { locale, setParamsTranslations } = useI18n()
const { data } = await useFetch(
  `https://yourapi.com/${locale}/page/${route.params.slug}`
)
setParamsTranslations(data.value.other_languages)
</script>
```

If anyone agrees I can create an issue (feature)</div>

kazupon:
[...]

PR #2580, opened 2023-11-29
feat!: add `useSetI18nParams` composable

BobbieGoede:
### 🔗 Linked issue
* #2429 
* #2006
* #1736

### ❓ Type of change

### 📚 Description
* Depends on https://github.com/intlify/routing/pull/82
* Still could add more tests.

This PR adds a composable `useSetI18nParams` which returns a function `setI18nParams` which can be used to set the translated parameters for the current route.

```ts
// `products/[slug].vue` (current page='/en/products/red-mug')
 
const setI18nParams = useSetI18nParams();
setI18nParams({
  en: {
    slug: 'red-mug'
  },
  nl: {
    slug: 'rode-mok'
  }
})
```

The composable function logic is basically a reimplementation of `useLocaleHead` from `vue-i18n-routing` except it uses Nuxt's built in unhead instance to manage and render the tags.

This PR only works combined with https://github.com/intlify/routing/pull/82, as this allows us to change where these dynamic translated parameters are sourced from. ~~In this implementation it uses `useState` to set and retrieve these parameters, which seems to work in my testing.~~

There is likely a better solution for this, but I hope to find a way to resolve the linked issues as I think this is/would be blocking users from moving to v8. If this works, we can always improve the functionality or performance later.

### Goal
I have removed references to using dynamic parameters with `definePageMeta`, the issue with using `definePageMeta` is that it's fixed at build time. 

Workarounds described https://github.com/nuxt-modules/i18n/issues/1736 bypass this by setting `route.meta.nuxtI18n` directly, this works except for updating the SEO tags during SSR, this PR works similar to the workarounds except SEO tags are updated and rendered on the server.

The issue described below also applies to those workarounds.

### Common issue
In the documentation I have added a warning about a common scenario as there is no reactivity during SSR. For example, setting dynamic parameters from within a page will not be reflected in a layout as it would have already been rendered (see [this comment](https://github.com/nuxt/nuxt/issues/18627#issuecomment-1408951142)). ~~This will cause a hydration mismatch if a language switcher is present in a layout or any page or component that renders before setting dynamic parameters.~~ This has been improved to match the workaround, hydration mismatch is avoided as links will be retroactively updated on the client side, this means not all links are rendered correctly on the server-side while the head tags are.

### Remarks
* I haven't removed the logic of dynamic parameters using `definePageMeta` as that would make this a breaking PR. ~~should this be fully removed and/or trigger a deprecation warning when still in use?~~ I have added deprecation warnings which log which pages are using the deprecated method.
* In the future I would like to improve and implement this in `vue-i18n-routing`, as `unhead` can be used there as well. It's not my intention to move code out of that repository, but the current implementation makes use of Nuxt's `unhead` instance.

### 📝 Checklist


docs/content/3.options/2.routing.md
```diff
@@ -126,47 +126,3 @@ Set to a path to which you want to redirect users accessing the root URL (`/`).
   path: 'about-us'
 }
 ```
-
-## `dynamicRouteParams`
```
kazupon: The `dynamicRouteParams` docs should be kept for those who have already implemented it. And a note and link to `useSetI18nParams` should be left for those who have already implemented it.

Let’s remove it in the next minor release, v8.1, after the v8 official release.

specs/fixtures/issues/2020/nuxt.config.ts
```diff
@@ -16,7 +16,6 @@ export default defineNuxtConfig({
         iso: 'en-US'
       }
     ],
-    dynamicRouteParams: true,
```
kazupon, line 19: If `dynamicRouteParams` option is used, nuxt i18n should warn to console.

specs/fixtures/switcher/pages/[...pathMatch].vue
```diff
@@ -3,11 +3,10 @@ import LangSwitcher from '../components/LangSwitcher.vue'
 
 const route = useRoute()
 
-definePageMeta({
```
Comment (kazupon) [inline specs/fixtures/switcher/pages/[...pathMatch].vue:6]: if we are using `nuxtI18n` in `definePageMeta`, nuxt i18n should warn the console.

The way to implement it is to implement it as a bundler plugin. The code is passed as a parameter to `transform`, so you can check it at that time.

kazupon (changes requested):
Great work!
I've just reviewed your PR!
Please check it!

BobbieGoede:
Thanks for the feedback! I agree that it's better to keep the docs and refer to the new composable, and add warnings when using the deprecated functionality. 

After more testing with this PR I ran into some SSR/hydration issues that I would like to fix as well. It will take some time to figure out the issues and to make the suggested changes so I'm marking this PR as WIP in the meantime!

BobbieGoede:
While trying to debug the hydration issues I found out I overcomplicated the solution a bit by using `useState`. This feature now internally uses the `nuxtI18n` approach, but improves its usage and ensures the meta tags are updated. I think there are a few ways to improve on the current way we handle head tags, but that will be something I can focus on aftwards.

@kazupon 
I have added the deprecation warnings in the documentation and also added the warning logs when usage is detected, should I add tests to confirm these logs are working as expected? And should it be possible to disable this logs?

kazupon (approved):
Great works!
LGTM!

kazupon:
@BobbieGoede 
> should I add tests to confirm these logs are working as expected? And should it be possible to disable this logs?

The logging test should be added if possible.
As far as I know, warning can be disabled by specifying it in the `useLogger` options.
https://nuxt.com/docs/api/kit/logging

`useLogger` uses consola, I think it may be possible to disable it by setting an ENV or log level.
https://github.com/unjs/consola

BobbieGoede:
@kazupon 
>The logging test should be added if possible.

Okay add those soon!

I think I will also change the arguments passed to `useSetI18nParams`, so that instead of `{ addDirAttribute?: boolean, addSeoAttribute?: boolean | SeoAttributesOptions }` it will accept `SeoAttributesOptions | undefined`. Reasoning behind it is that this function shouldn't change the `dir` attribute (will remove that), and right now if `addSeoAttribute` isn't passed the function won't do anything.

BobbieGoede:
@kazupon 
Actually I don't know how to test those logs as they happen at build time 😅

>useLogger uses consola, I think it may be possible to disable it by setting an ENV or log level.
[unjs/consola](https://github.com/unjs/consola)

Not sure if this is possible, as far as I can tell it is possible to set the log level, but changing it would affect all logs, not just for this module.

I did change the composable parameter as I described.

kazupon:
@BobbieGoede 

> Actually I don't know how to test those logs as they happen at build time 😅

I think that we can do test at unit test that work unplugin directly. However, that is really difficult as it requires mock.
we can test the `dynamicRouteParams` and `route.meta.nuxtI18n` warning logs at playground. 

I've pull the your branch in my local, and I've checked your branch with playground.
I saw the warning logs.
I can merge your PR, so it looks good for me. :)

We should release rc.9, because that PR might have still issues.
If there is the issue, just fix it!

BobbieGoede:
Okay good! I'll merge this and release `rc.9` in a moment 💪

kazupon:
@BobbieGoede 
I'll merge your PR!

PR #2599, opened 2023-12-06
feat: `defineI18nRoute` name parameter

BobbieGoede:
### 🔗 Linked issue
* #2581

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description
This PR is still missing documentation and tests. This also moves/integrates and modifies route localization code from `vue-i18n-routing to make this possible.

This adds a `name` property to the object passed to `defineI18nRoute` to use instead of passing `name` to `definePageMeta` as this causes issues with page generation, see #2581. Perhaps I can add warnings when `name` is passed to `definePageMeta` as well.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

Issue #2581, opened 2023-11-30
Using `definePageMeta({ name: ... })` breaks route generation


Anoesj:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v20.9.0
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.0
- Package Manager:  pnpm@8.11.0
- Builder:          -
- User Config:      ssr, experimental, modules, app, runtimeConfig, i18n, imports, typescript, devtools, vite, vue, image
- Runtime Modules:  @nuxtjs/stylelint-module@5.1.0, @pinia/nuxt@0.5.1, nuxt-schema-org@3.2.1, @nuxtjs/i18n@8.0.0-rc.5, @nuxt/image@1.1.0
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-jiqdtm?file=app.vue

Note that only the `/ja` route works here, which is exactly why I'm writing this bug report.

### Describe the bug

Let's assume the following base Nuxt i18n configuration with all examples below:
```ts
{
  defaultLocale: 'nl',
  lazy: true,
  langDir: 'locales',
  differentDomains: false,
  strategy: 'prefix',
  detectBrowserLanguage: false,
  locales: [
    {
      name: 'Nederlands',
      code: 'nl',
      iso: 'nl',
      file: 'nl.json',
    },
    {
      name: 'English',
      code: 'en',
      iso: 'en',
      file: 'en.json',
    },
    {
      name: 'Français',
      code: 'fr',
      iso: 'fr',
      file: 'fr.json',
    },
    {
      name: 'Deutsch',
      code: 'de',
      iso: 'de',
      file: 'de.json',
    },
  ],
}
```

A page can be named using:
```ts
// pages/products.vue
definePageMeta({
  name: 'products',
});
```

There are two situations where Nuxt i18n doesn't work nicely with named routes:
1. `customRoutes: 'config'` and route translations for `products` in `nuxt.config.ts`:
```ts
// nuxt.config.ts
customRoutes: 'config',
pages: {
  'products': {
    nl: '/producten',
    en: '/products',
    fr: '/produits',
    de: '/produkte',
  },
}
```
2. `customRoutes: 'page'` and route translations for `products` inside `products.vue` using `defineI18nRoute`:
```ts
definePageMeta({
  name: 'products',
});

defineI18nRoute({
[...]

PR #2603, opened 2023-12-06
fix: `definePageMeta` name overriding localized route name

BobbieGoede:
### 🔗 Linked issue
* #2581

### ❓ Type of change

### 📚 Description
While trying to figure out bundler plugins I figured there was an alternative way to tackle #2581. Maybe this is easier to break by Nuxt updates, but I feel this solution makes more sense than https://github.com/nuxt-modules/i18n/pull/2599. I still need to test this with disabled routes, and I may be missing other edge cases, please let me know.

What this PR does is transform the generated `routes.mjs` file, instead letting `name` defined in `definePageMeta` override the localized name (see [source code](https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/pages/utils.ts#L333)), we capture the locale route name suffix and combine it with the name if it exists, this makes named routes work as expected.

So in `routes.mjs` the generation would be changed like this example
```diff
-name: historyeTgp8UZZfFMeta?.name ?? "history___ja",
+name: (historyeTgp8UZZfFMeta?.name ? historyeTgp8UZZfFMeta?.name + "___ja" : undefined) ?? "history___ja",
```

What do you think @kazupon?

### 📝 Checklist

### DIFF
--- a/playground/pages/about/index.vue
+++ b/playground/pages/about/index.vue
@@ -32,6 +32,7 @@ export default defineComponent({
       }
     })
     definePageMeta({
+      name: 'named-about',
       title: 'pages.title.about'
     })
     return {}
--- a/playground/pages/index.vue
+++ b/playground/pages/index.vue
@@ -74,7 +74,8 @@ definePageMeta({
     <p>{{ $t('bar.buz', { name: 'buz' }) }}</p>
     <h2>Pages</h2>
     <nav>
-      <NuxtLink :to="localePath('/')">Home</NuxtLink> | <NuxtLink :to="localePath({ name: 'about' })">About</NuxtLink> |
+      <NuxtLink :to="localePath('/')">Home</NuxtLink> |
+      <NuxtLink :to="localePath({ name: 'named-about' })">About</NuxtLink> |
       <NuxtLink :to="localePath({ name: 'blog' })">Blog</NuxtLink> |
       <NuxtLink :to="localePath({ name: 'server' })">Server</NuxtLink> |
       <NuxtLink :to="localePath({ name: 'category-id', params: { id: 'foo' } })">Category</NuxtLink> |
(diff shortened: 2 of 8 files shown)


kazupon:
@BobbieGoede  Thanks!

> What this PR does is transform the generated `routes.mjs` file, instead letting `name` defined in `definePageMeta` override the localized name (see [source code](https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/pages/utils.ts#L333)), we capture the locale route name suffix and combine it with the name if it exists, this makes named routes work as expected.
> 
> So in `routes.mjs` the generation would be changed like this example
> 
> ```diff
> -name: historyeTgp8UZZfFMeta?.name ?? "history___ja",
> +name: (historyeTgp8UZZfFMeta?.name ? historyeTgp8UZZfFMeta?.name + "___ja" : undefined) ?? "history___ja",
> ```

Your PR looks good!

However, this approach is more of a workaround & hack.
If possible, we hope that work-around this issue if there is a correct way to extend the URL of the named page on the nuxt.

@danielroe 
Do you have any solution or idea?

BobbieGoede:
@kazupon 
>However, this approach is more of a workaround & hack.
If possible, we hope that work-around this issue if there is a correct way to extend the URL of the named page on the nuxt.

I agree it does feel hacky, my initial solution was #2599 but the drawback is that users would have to avoid using `definePageMeta({ name: '' })` and instead use `defineI18nRoute({ name: '' })`. 

The solution in this PR requires no changes in code or usage for users, but the drawback being that it's more likely to break. 

Personally I think the best solution would be to change the generated file from Nuxt's side, the way some data from `definePageMeta` overrides other values is inconsistent as you can see [here](https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/pages/utils.ts#L327-L338). For some reason it's just `name` and `path` that overwrite resolved page values, while `alias`, `meta`, and `redirect` overwrite values extracted from `definePageMeta`. 

~~I think I'll just open a PR to make this more consistent, but I can imagine changing this could be breaking for existing Nuxt projects. 🤔~~ This change is not as straightforward as I expected, curious to hear Daniel's perspective.

danielroe:
I think we might consider reworking Nuxt implementation so we actually expose some page metadata at build-time (via `pages:extend`).... That will solve some edge cases like this (and with typed pages as well, where a custom `name` matters).

BobbieGoede:
Closing this as it would break if changes are made upstream, will keep an eye on https://github.com/nuxt/nuxt/issues/24770 and if I have time later (maybe next year 😅) I'll see if I can work on resolving that.

Issue #2612, opened 2023-12-11
Version 8.0.0 rc 9 throws memory leak


agracia-foticos:
### Environment

- Operating System: `Linux`
- Node Version:     `v20.5.1`
- Nuxt Version:     `3.8.2`
- CLI Version:      `3.10.0`
- Nitro Version:    `2.8.1`
- Package Manager:  `npm@9.8.0`
- Builder:          `-`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

This graph its our application with i18n 8.0.0 rc 9, with one screaming frog crawl with 20 simultaneous threads and one Jmeter with 100 threads simultaneously.
The memory raises and the garbage doesnt recollect the memory (The drop that is seen is because we restart the server automatically when we reach a certain threshold)
![image](https://github.com/nuxt-modules/i18n/assets/77386602/a4c943d2-f7a4-4167-864b-6a4e9243a4f3)

The same situation (the same packages and config), but with i18n 8.0.0 rc 5 (only we change the version of this library)... the memory is stable.

![image](https://github.com/nuxt-modules/i18n/assets/77386602/6becfd21-0214-492a-a57a-1748dde8866d)

### Describe the bug

This graph its our application with i18n 8.0.0 rc 9, with one screaming frog crawl with 20 simultaneous threads and one Jmeter with 100 threads simultaneously.
The memory raises and the garbage doesnt recollect the memory (The drop that is seen is because we restart the server automatically when we reach a certain threshold)
![image](https://github.com/nuxt-modules/i18n/assets/77386602/a4c943d2-f7a4-4167-864b-6a4e9243a4f3)

The same situation (the same packages and config), but with i18n 8.0.0 rc 5 (only we change the version of this library)... the memory is stable.

![image](https://github.com/nuxt-modules/i18n/assets/77386602/6becfd21-0214-492a-a57a-1748dde8866d)

agracia-foticos:
It's impossible to replicate our environment...

We can provide us our package.json and nuxt.config.ts ??

BobbieGoede:
A minimal reproduction would of course help the most, but if you could provide package.json and nuxt config that would narrow down what features and dependencies could cause this 🙏.

kazupon:
If possible, I would like to find out from which version this is happening.
Can you reproduce the memory leak in each between rc.6 and rc.9? 🙏

agracia-foticos:
[...]

Issue #2034, opened 2023-05-02
Huge memory leak


cyperdark:
### Environment

- Operating System: Windows_NT
- Node Version:     v16.20.0
- Nuxt Version:     3.4.3
- Nitro Version:    2.3.3
- Package Manager:  yarn@1.22.19
- Builder:          vite
- User Config:      telemetry, ssr, modules, i18n
- Runtime Modules:  @nuxtjs/i18n@8.0.0-beta.11
- Build Modules:    -

### Reproduction

Source code: https://github.com/cyperdark/i18n

### Describe the bug

Memory leak video:
https://i.osuck.link/1683034408_ScxZCLgnzT.mp4

cyperdark:
This is how memory graph look like in production (when memory go down it means app crashed)
![image](https://user-images.githubusercontent.com/21735205/236648930-0e099189-885e-43b9-bec2-045afb82991d.png)

kazupon:
Thank you for your reporting!

I think that v8.0.0-beta.11 is buggy. I am fixing on it now, but that bug may have affected it.

we would like to see if memory leaks occur in the previous version v8.0.0-beta.10 to make sure this version is affected.
If you have the time we would appreciate your help. 🙏

cyperdark:
I tried on versions below, and until version 8.0.0-beta.7 situation is the same
BUT on this version i got errors

thats weird, need more testing
![image](https://user-images.githubusercontent.com/21735205/236970075-5ace3783-8932-4e3e-8c09-5080cdbbb422.png)

cyperdark:
Potentially i found what caused this leak


1. If i use **Head Title with $t**, memory went to **1gb** (1000 requests) then go to normal **~90**
![image](https://user-images.githubusercontent.com/21735205/236992899-0564ae82-d8c6-42fe-a406-05290e7b9f8c.png)


2. same situation with **useLang**, from **60** went to **1gb** then down to **85-90**
![image](https://user-images.githubusercontent.com/21735205/236992986-d2e338d9-4804-48e6-bf3e-f0d7fad195fe.png)
![image](https://user-images.githubusercontent.com/21735205/236993008-5464f17c-d3d3-4668-a4c3-094b8525085e.png)


3. but if i use **computed** title, from **60** to memory cap
![image](https://user-images.githubusercontent.com/21735205/236993451-3d2d8034-2f6b-4dec-b3ee-f8c8afa9a5bf.png)
[...]

PR #2616, opened 2023-12-12
fix: memory leak

BobbieGoede:
### 🔗 Linked issue
* #2612
* #2034 

### ❓ Type of change

### 📚 Description
It is unlikely that this fixes all present memory leaks in the module, but [this small change](https://github.com/nuxt-modules/i18n/pull/2616/commits/08958630588937cef18975d0f1a7a5bdc4233ffd) appears to have a measurable impact on the memory usage. 

I ran the reproductions with the latest release `rc.9`, and took memory snapshots with chrome. First without any requests, then running `siege` for 10 seconds with 10 concurrent users (`siege -t 10s -c 10 "http://localhost:3000/"`) after which I took another snapshot. After waiting for a short moment I took another snapshot in case any more memory was cleared 🤷‍♂️. Then I installed my local modified `@nuxtjs/i18n` and went through the same steps.

My testing isn't exactly thorough, but it was a reliable way to see the memory usage go up and stay up, it's possible memory will still increase after longer usage and uptime, it's something that takes time and patience.

Below a comparison of the snapshots before and after installing the modified package:
memory-usage-comparison

I also tested https://github.com/nuxt-modules/i18n/issues/2034#issuecomment-1659386170 (or https://github.com/danielroe/i18n-memory-leak) after making this comparison image, memory usage went from ~121mb to ~33mb.

While profiling I kept seeing `nuxtI18Options` and `baseUrl`, based on thismy theory is that the nuxt context is not being disposed of correctly as it is being set on `nuxtI18nOptions` in the `i18n.ts` plugin. 

For example here:
https://github.com/nuxt-modules/i18n/blob/4bf4f5013a68a9fa7dcf0d7035d367cd40710eb1/src/runtime/plugins/i18n.ts#L86-L91

`nuxtI18nOptions.baseUrl` is being set to a function returned from `extendBaseUrl`, the returned function makes use of the passed nuxt context (`options.nuxt`), `nuxtI18nOptions` is not defined in the scope of the plugin so I can imagine it would be hard to know when to dispose/free the memory? Based on this assumption I instead create a new object using `nuxtI18nOptions`, which since it's scoped to the plugin should be easier to safely dispose/free.

I don't know enough about memory management or garbage collection in Javascript, but I have experience optimizing memory usage for mobile games in C#, to provide a background for my reasoning. If someone has good resources on this please share them so I can form more informed conclusions 😅

#### Update
Based on the testing I did, I changed the way the Nuxt context is used/retrieved. Instead of passing it to functions/closures and potentially causing leaks the `useNuxtApp` composable is used where possible. 

There is still a small and slow increase in memory usage for every few 1000s of requests, will look into that as well.

### 📝 Checklist

### DIFF
--- a/src/runtime/internal.ts
+++ b/src/runtime/internal.ts
@@ -11,7 +11,7 @@ import {
 } from 'vue-i18n-routing'
 import { hasProtocol } from 'ufo'
 import isHTTPS from 'is-https'
-import { useRequestHeaders, useRequestEvent, useCookie as useNuxtCookie } from '#imports'
+import { useRequestHeaders, useRequestEvent, useCookie as useNuxtCookie, useRuntimeConfig, useNuxtApp } from '#imports'
 import { nuxtI18nOptionsDefault, NUXT_I18N_MODULE_ID, isSSG } from '#build/i18n.options.mjs'
 
 import type { NuxtApp } from '#app'
@@ -86,7 +86,7 @@ export function parseAcceptLanguage(input: string): string[] {
 }
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export function getBrowserLocale(options: Required<NuxtI18nInternalOptions>, context?: any): string | undefined {
+export function getBrowserLocale(options: Required<NuxtI18nInternalOptions>): string | undefined {
   let ret: string | undefined
 
   if (process.client) {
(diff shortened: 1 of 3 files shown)


dargmuesli:
Wow, I just found the same cause and wanted to report it! :joy: Good thing this PR was a bit quicker than me :wink: I'll try it in my project now.

dargmuesli (approved):
Tried and it seems to resolve the issue! :heart: 

There seems to be another leak in some head / pinia related files in my project though :sweat_smile: but that's unrelated to nuxt-i18n.

kazupon:
That's amazing work! ❤️ 

The causing of memory leaks is very difficult since the code must be checked.
I appreciate your efforts 🙇 


> I don't know enough about memory management or garbage collection in Javascript, but I have experience optimizing memory usage for mobile games in C#, to provide a background for my reasoning. If someone has good resources on this please share them so I can form more informed conclusions

As far as I know, I think you can control GC in Javascript by using `WeakRef` or `FinalizationRegistry` to get around it.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef

However, unlike C# and other languages, JavaScript does not currently provide a mechanism to release resources like the dispose / descructor pattern.
So, I think the only way to intentionally control memory release in JavaScript is to use those APIs.

Once again, thank you so much!

kazupon (approved):

BobbieGoede:
> There seems to be another leak in some head / pinia related files in my project though 😅 but that's unrelated to nuxt-i18n.

@dargmuesli 
Happy to hear it's working, thanks for testing! I do think there is a smaller memory leak related to the head tags, still need to check if it is present without `@nuxtjs/i18n`, it may be caused by nuxt itself.

>As far as I know, I think you can control GC in Javascript by using WeakRef or FinalizationRegistry to get around it.

@kazupon 
I still need to learn more about those APIs, as well as `WeakSet` and `WeakMap`, maybe it would be easier to prevent leaks with those 😅

PR #2624, opened 2023-12-16
fix: do not use import from `nuxt/dist/app`

danielroe:
### ❓ Type of change

### 📚 Description

[Found via ecosystem-ci](https://github.com/nuxt/ecosystem-ci/actions/runs/7233707958/job/19710329753). These type imports will only work with old (legacy) module resolution. 

Instead I would recommend using `nuxt/app` (for type imports only) or `#app`.

### 📝 Checklist

### DIFF
--- a/src/runtime/server/plugin.ts
+++ b/src/runtime/server/plugin.ts
@@ -7,7 +7,7 @@ import { loadVueI18nOptions, loadInitialMessages, makeFallbackLocaleCodes, loadL
 
 import type { NitroAppPlugin } from 'nitropack'
 import type { H3Event } from 'h3'
-import type { NuxtApp } from 'nuxt/dist/app'
+import type { NuxtApp } from 'nuxt/app'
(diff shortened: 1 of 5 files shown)


src/runtime/server/plugin.ts
```diff
@@ -7,7 +7,7 @@ import { loadVueI18nOptions, loadInitialMessages, makeFallbackLocaleCodes, loadL
 
 import type { NitroAppPlugin } from 'nitropack'
 import type { H3Event } from 'h3'
-import type { NuxtApp } from 'nuxt/dist/app'
+import type { NuxtApp } from 'nuxt/app'
```
s00d, line 10: It should also have #app here, right?

danielroe: In this case, `#app` is only meant to be imported in the Vue part of the app, not the Nitro part, so `#app` doesn't have any meaning there. But, since it's just a type import, it's okay to keep - but in that case we can use the subpath `nuxt/app`.

Issue #2629, opened 2023-12-19
Memory Leak when using top-level await before defining computed translations


azhirov:
### Environment

- Operating System: Linux
- Node Version:     v20.8.1
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@10.1.0
- Builder:          -
- User Config:      app, modules, devtools
- Runtime Modules:  @pinia/nuxt@0.5.1, @nuxtjs/i18n@8.0.0-rc.11
- Build Modules:    -

### Reproduction

I've created a minimal reproduction [repository](https://github.com/azhirov/nuxt-i18n-pinia-memory-leak) to illustrate the problem. To reproduce the issue:

1. Clone the repository: `git clone git@github.com:azhirov/nuxt-i18n-pinia-memory-leak.git`
2. Install dependencies: `npm install`
3. Build the project: `npm run build`
4. Run the server: `node .output/server/index.mjs`
5. Use a tool like `ab`, `siege`, or any other utility to send multiple requests. For example: `siege -c10 -t30s http://127.0.0.1:3000` or `ab -n 5000 -c 100 http://localhost:3000/`
6. Check the memory consumption of the process (for example, using the `top` command).

### Describe the bug

I've encountered a peculiar memory leak issue related to the simultaneous usage of the `@nuxtjs/i18n` and `@pinia/nuxt` modules in a Nuxt 3 project. The problem manifests when an asynchronous method from the Pinia store is called (using `await`) before a `computed` property that includes `useNuxtApp().$i18n.t()`. This results in a memory leak. Interestingly, moving the asynchronous method below the `computed` property prevents the leak. Additionally, if the `$i18n.t()` is not used in the computed property, the leak does not occur.

Code with memory leak:
```
import useFooStore from '~/store/foo';

const fooStore = useFooStore();
await fooStore.loadData()

const text = computed(() => {
  const i18n = useNuxtApp().$i18n;
  return i18n.t('seventeen');
})
```

After running the command `node .output/server/index.mjs`, the server consumes approximately 20 MB of RAM. However, after executing the command `siege -c10 -t30s http://127.0.0.1:3000`, the memory usage increases to 1.4 GB and remains at approximately this level even after stopping the load.

![node_1091683](https://github.com/nuxt-modules/i18n/assets/29061184/00714846-5f98-4612-b37f-af750da0f6b1)

[...]

PR #2647, opened 2024-01-03
fix: Memory leak

s00d:
### 🔗 Linked issue

https://github.com/nuxt-modules/i18n/issues/2629

This is just an example, no need to merge it. Because of this fix, everything works 10 times slower. We need a proper fix.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/runtime/plugins/i18n.ts
+++ b/src/runtime/plugins/i18n.ts
@@ -132,13 +132,13 @@ export default defineNuxtPlugin({
     __DEBUG__ && console.log('first detect initial locale', initialLocale)
 
     // load initial vue-i18n locale messages
-    vueI18nOptions.messages = await loadInitialMessages(vueI18nOptions.messages, localeMessages, {
+    vueI18nOptions.messages = JSON.parse(JSON.stringify(await loadInitialMessages(vueI18nOptions.messages, localeMessages, {
       ...nuxtI18nOptions,
       initialLocale,
       fallbackLocale: vueI18nOptions.fallbackLocale,
       localeCodes,
       cacheMessages
-    })
+    })))
 
     /**
      * NOTE:


rchl:
- the existing approach where `vueI18nOptions.messages` is passed to `loadInitialMessages` and then both modified inside and returned feels unnecessarily confusing
- if the problem is that `loadInitialMessages` sets something on `messages` that blocks it from being collected then maybe it should itself take care of not doing that (by assigning stringified/parsed data maybe).

s00d:
> * the existing approach where `vueI18nOptions.messages` is passed to `loadInitialMessages` and then both modified inside and returned feels unnecessarily confusing
> * if the problem is that `loadInitialMessages` sets something on `messages` that blocks it from being collected then maybe it should itself take care of not doing that (by assigning stringified/parsed data maybe).

The situation is more complicated here. I tried the approach you described first, but the issue is something else, which I haven't identified yet. For now, this is the only working method to detect the problem. I created the pull request just as an example to help solve the leak issue.

BobbieGoede:
Thanks for your continued investigation of this leak @s00d! First testing of this change seems promising, but I will have a more thorough look later. 

I agree with @rchl, but as it is, this PR is helpful in proving that this may be the source of the memory leak. I'll test it further and see if I can refactor/simplify the relevant functions.

s00d:
image
 seems the entire problem lies here: sometimes deepCopy malfunctions and returns true for objects and arrays. After that, the garbage collector can't clean up the data because it's being copied by reference into the objects. No wonder there were comments about deepCopy.
 
 
 ```js
const isObject = (val) => val !== null && typeof val === "object";
const isArray = Array.isArray;
function deepCopy(src, des) {
  if (Array.isArray(src) && Array.isArray(des)) {
    for (let i = 0; i < src.length; i++) {
      if (isObject(src[i])) {
        des[i] = deepCopy(src[i], des[i] || (Array.isArray(src[i]) ? [] : {}));
      } else {
        des[i] = src[i];
      }
    }
  } else if (isObject(src) && isObject(des)) {
    for (const key in src) {
      if (Object.prototype.hasOwnProperty.call(src, key)) {
        if (isObject(src[key])) {
          des[key] = deepCopy(src[key], des[key] || {});
        } else {
          des[key] = src[key];
        }
      }
    }
  } else {
    throw new Error("Invalid value");
  }
  return des;
}
 ```
 
 Here's a simple example of deep copying with array support. After this, the memory leak disappears, but due to the increased complexity of the function, there's a slowdown in performance.
 
 It seems that indeed it is necessary to create a translation cache to avoid processing arrays every time.
 
[Here's ](https://github.com/intlify/vue-i18n-next/pull/1675) an example, but something needs to be done about performance. If the array is large, it significantly slows down Nuxt.

PR #2655, opened 2024-01-05
docs: Update 1.setup.md

JRew1234:
Added a paragraph about installing @nuxtjs/i18n using installModule function

### ❓ Type of change

### 📚 Description

I have added a paragraph about installing @nuxtjs/i18n using installModule function. Currently there is no information in docs about passing config the correct way when using installModule. If there is no 'overrides' key in inlineOptions there will be no vueI18n config imported to i18n.options.mjs. I have modified this old stackblitz to show the issue, https://stackblitz.com/edit/nuxt-i18n-queistion-3v8ggj?file=.nuxt%2Fi18n.options.mjs
If you delete 'overrides' there will be no mergeVueI18NConfigs() with a path to the config file.

### 📝 Checklist

### DIFF
--- a/docs/content/1.getting-started/1.setup.md
+++ b/docs/content/1.getting-started/1.setup.md
@@ -54,3 +54,13 @@ export default defineNuxtConfig({
   ]
 })
 ```
+
+To install the `@nuxtjs/i18n` module using the `installModule` function, you need to include your settings inside an object labeled `overrides`, which is a part of the second argument of the function. Here's how you can do it:
+In your `module.ts` file, use the `installModule` function and specify `@nuxtjs/i18n` as the first argument. For the second argument, create an object and within this object, use the key `overrides` and specify your options.
+```ts {}[module.ts]
+await installModule('@nuxtjs/i18n', {
+  overrides: {
+    /* module options */
+  }
+})
+```
--- a/docs/content/1.getting-started/1.setup.md
+++ b/docs/content/1.getting-started/1.setup.md
@@ -55,8 +55,7 @@ export default defineNuxtConfig({
 })
 ```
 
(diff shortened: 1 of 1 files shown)


BobbieGoede:
Hmm, I'm honestly not entirely sure what the intended way is to support `installModule` for this module. The `overrides` property is used internally to make certain test configurations possible, this will interfere/overwrite options passed by layers and modules. 

I recommend using layers instead of using a module to fully configure this module. If this is a common pattern in projects I could look into expanding the `i18n:registerModule` to support more options, or ideally fully support `installModule` (I tried this before but ran into issues).

JRew1234:
I will look into layers, thanks:) 

To be honest it feels like a workaround but I think it's the only way to pass vuei18n config path into the nuxtjs/i18n, the issue is this function called resolveVueI18nConfigInfo which uses layers, when config is passed as an argument to installModule it won't be merged into i18n options.

If you check this originial stackblitz, https://stackblitz.com/edit/nuxt-i18n-queistion-ckrqt7?file=nuxt.config.ts,package.json,modules%2Fi18n-resource%2Findex.ts 
vueI18n path is passed to the i18n.options.mjs only by 'accident'. this path is the default in the resolveVueI18nConfigInfo function, if you move config file into any other folder it won't work because there are no options passed into this function

kazupon:
> Hmm, I'm honestly not entirely sure what the intended way is to support installModule for this module. The overrides property is used internally to make certain test configurations possible, this will interfere/overwrite options passed by layers and modules.

As @BobbieGoede already mentioned, `overrides` is not explicitly explained in the nuxt documentation either, and is primarily an option for internal use.
So we cannot put your PR in the documentation.
I recommend using layers instead of using a module to fully configure this module.

we appreciate your PR to improve the nuxt i18n documentation, but I will close it once.

Thanks!

PR #2714, opened 2024-01-20
perf: favor importing generated options over passing as argument

BobbieGoede:
### ❓ Type of change

### 📚 Description
Instead of passing the generated `nuxtI18nOptions` as arguments, I instead import `nuxtI18nOptions` when needed or only pass the relevant properties instead of the full object. 

Since most of the memory leaks so far have been related to the generated options, and we probably shouldn't mutate the generated options, this may reduce memory usage and reduce the odds of keeping a reference to the entire object in a closure. Maybe we could consider freezing the generated options to prevent any mutations, something to look into another time.

### 📝 Checklist


kazupon (approved):
Great work! ❤️ 
Thanks!

PR #2723, opened 2024-01-22
fix: rollup comment annotation warning

### LINKED ISSUE #2640: "/*#__PURE__*/"  contains an annotation that Rollup cannot interpret due to the position of the comment. The comment will be removed to avoid issues.
### Environment

- Operating System: Darwin
- Node Version:     v20.8.1
- Nuxt Version:     3.9.0
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@10.2.5
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://github.com/s00d/i18n-memory-pure

### Describe the bug

pnpm i
pnpm build

### Additional context

When using defineI18nRoute on the latest version of Nuxt, it now says

app.vue?vue&type=script&setup=true&lang.ts (11:16) A comment

"/*#__PURE__*/"

in "app.vue?vue&type=script&setup=true&lang.ts" contains an annotation that Rollup cannot interpret due to the position of the comment. The comment will be removed to avoid issues.

### Logs

_No response_

BobbieGoede:
### 🔗 Linked issue
* #2640 

### ❓ Type of change

### 📚 Description
Resolves #2640 

Not sure if it this is the correct way to solve this issue, I don't exactly understand how the comment works and should be handled. I looked at https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/pages/runtime/composables.ts#L55-L83 for reference, our `defineI18nRoute` macro works in a similar way to `definePageMeta` in Nuxt but that doesn't use a comment like this, while `defineRouteRules` does which is also similar to our macro. 

Let me know if this should be solved differently, this change simply removes the comment and adds a usage check just like `definePageMeta` has.

### 📝 Checklist

### DIFF
--- a/src/transform/macros.ts
+++ b/src/transform/macros.ts
@@ -61,7 +61,7 @@ export const TransformMacroPlugin = createUnplugin((options: TransformMacroPlugi
       const match = code.match(new RegExp(`\\b${NUXT_I18N_COMPOSABLE_DEFINE_ROUTE}\\s*\\(\\s*`))
       if (match?.[0]) {
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        s.overwrite(match.index!, match.index! + match[0].length, `/*#__PURE__*/ false && ${match[0]}`)
+        s.overwrite(match.index!, match.index! + match[0].length, `false && /*#__PURE__*/ ${match[0]}`)
       }
 
       if (!parseQuery(search).macro) {


src/transform/macros.ts
```diff
@@ -1,74 +0,0 @@
...
-          }
-        }
-      }
-
-      // tree-shake out any runtime references to the macro.
-      // we do this first as it applies to all files, not just those with the query
-      const match = code.match(new RegExp(`\\b${NUXT_I18N_COMPOSABLE_DEFINE_ROUTE}\\s*\\(\\s*`))
-      if (match?.[0]) {
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        s.overwrite(match.index!, match.index! + match[0].length, `/*#__PURE__*/ false && ${match[0]}`)
```
kazupon, line 64: I think #2640 issue is using `#__PURE__` here. vite 5 i.e. rollup v4, it is recommended to use `#__NO_SIDE_EFFECTS__` Notation. For more information, see this document organized by anthony
https://github.com/javascript-compiler-hints/compiler-notations-spec/blob/main/no-side-effects-notation-spec.md.

nuxt i18n need to support DCE (dead code elimination) in bundler for `defineI18nRoute`
we should not remove this file.

kazupon (changes requested):
Thanks!
I've just reviewed this PR!
You can check it! :)

danielroe:
I think if you update to `/*@__PURE__*/` it will prevent the warning.

src/transform/macros.ts
BobbieGoede: The warnings in #2640 are triggered because `defineI18nRoute` is used, and this code adds the `#__PURE__` comment. I changed the PR to switch around the comment to be after `false && `, this seems to prevent the rollup warning from being triggered as well.

Maybe I misunderstand the document describing `#__NO_SIDE_EFFECTS__`, but it seems it may achieve the same thing as `#__PURE__`, but instead of having to insert the comment at each function call it can be defined at function declaration instead. So maybe we can remove these transformations after all?

danielroe: if appropriate, yes, that works just as well, with fewer annotations.

kazupon: > if appropriate, yes, that works just as well, with fewer annotations.

Yeah!
As daniel said, even without the notation, bundler should recognize `false` and do tree-shake.
For `#__NO_SIDE_EFFECTS__`, we should not use it because it's incompatible with webpack as mentioned in #2725, since Nuxt provides an option to use webpack as a bundler as well

PR #2785, opened 2024-02-13
fix: `statusCode` in `rootRedirect` not working with strategy 'prefix'

### LINKED ISSUE #2758: statusCode in rootRedirect not working with strategy 'prefix'
### Environment

- Operating System: `Linux`
- Node Version:     `v20.11.0`
- Nuxt Version:     `3.9.1`
- CLI Version:      `3.10.0`
- Nitro Version:    `2.8.1`
- Package Manager:  `pnpm@8.14.1`
- Builder:          `-`
- User Config:      `extends`, `app`, `i18n`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-tdp4cc?file=nuxt.config.ts

To reproduce, open the preview in a new tab and watch Network Devtools while navigating to `/`. It will show a 302 redirect, even though 307 has been configured.

### Describe the bug

When using `rootRedirect`, the `statusCode` option is ignored (actually overwritten afterwards, I believe) and a `302 Found` is returned. 

### Additional context

_No response_

### Logs

_No response_

BobbieGoede:
### 🔗 Linked issue
* #2758 

### ❓ Type of change

### 📚 Description
Resolves #2758 

I think the `rootRedirect.statusCode` was ignored because the request was redirected twice, once from `/` to the configured `rootRedirect.path` (e.g. `about`) with the statusCode, then again because `/about` doesn't exist for strategy `prefix` so it will be redirected to `/en/about` while ignoring `rootRedirect` configuration, as it's no longer redirecting on root.

This PR changes `rootRedirect` to try and translate the redirect path, so that a single redirection happens with the configured status code.

### 📝 Checklist

### DIFF
--- a/src/runtime/utils.ts
+++ b/src/runtime/utils.ts
@@ -329,7 +329,6 @@ type NavigateArgs = {
 }
 
 function _navigate(redirectPath: string, status: number) {
-  console.log(redirectPath, status)
   return navigateTo(redirectPath, { redirectCode: status })
 }
 
--- a/src/runtime/utils.ts
+++ b/src/runtime/utils.ts
@@ -321,6 +321,7 @@ function isRootRedirectOptions(rootRedirect: unknown): rootRedirect is RootRedir
 const useRedirectState = () => useState<string>(NUXT_I18N_MODULE_ID + ':redirect', () => '')
 
 type NavigateArgs = {
+  nuxtApp: NuxtApp
   i18n: I18n
   redirectPath: string
   locale: string
@@ -328,6 +329,7 @@ type NavigateArgs = {
 }
 
 function _navigate(redirectPath: string, status: number) {
+  console.log(redirectPath, status)
   return navigateTo(redirectPath, { redirectCode: status })
 }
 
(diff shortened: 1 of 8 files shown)


src/runtime/utils.ts
```diff
@@ -321,13 +321,15 @@ function isRootRedirectOptions(rootRedirect: unknown): rootRedirect is RootRedir
...
 type NavigateArgs = {
+  nuxtApp: NuxtApp
   i18n: I18n
   redirectPath: string
   locale: string
   route: RouteLocationNormalized | RouteLocationNormalizedLoaded
 }
 
 function _navigate(redirectPath: string, status: number) {
+  console.log(redirectPath, status)
```
kazupon:
```suggestion
```

kazupon: Maybe you forgot to delete the debug log, we should remove the console log. :)
If we want to output debug logs, we should use them with the `__DEBUG__` flag.

kazupon (changes requested):
This PR has passed testing, so I think the logic is fine. Once you have fixed my review point, you can merge them!

src/runtime/utils.ts
BobbieGoede: Ah! You're right I forgot to remove it 😅

PR #2796, opened 2024-02-15
docs: add lang switching vue-i18n caveat

BobbieGoede:
### ❓ Type of change

### 📚 Description
This is a common pitfall and I'm guessing this mostly happens for users who first used `vue-i18n` only. Directly mutating `locale` and not using `setLocale` or `switchLocalePath` navigation when switching language is an easy mistake, hopefully this informs users before they make it or at least it can be referenced when users open an issue.

I'm thinking we should probably add a section with common pitfalls or best practices that includes bits of information like this. 

### 📝 Checklist


kazupon (approved):
Thanks!
That's good caveat doc!
LGTM ❤️

PR #2799, opened 2024-02-15
feat: add `loadLocaleMessages` to manually load locale messages

### LINKED ISSUE #740: Can't access non-loaded lazy locales using $t
Hello.

I was trying to put two different languages on the same page. 
In my example, I'm trying to pull messages for two different languages i.e. polish and french.
The idea is to get language from a browser (fr) and compare it to language from URL(pl) which will show the message about difference and choose between two of them. That message should be in those two languages that I've got from the browser and URL.  

I tried use v-t directive with the locale attribute, but it seems to not work as I expected.
Tried also $t with same effect.

After checking i18n object I had only current language messages (in this situation polish) and both strings were translated into polish only...

My nuxt-i18n config
```
  parsePages: false,
  locales: LOCALES,
  defaultLocale: process.env.NODE_ENV !== 'production' ? 'en' : 'en-gb',
  lazy: true,
  langDir: 'lang/',
  vueI18n: {
    fallbackLocale: 'en',
    silentFallbackWarn: true,
    formatFallbackMessages: true,
  },
```

Is there any possibility to use both languages on the same page?

BobbieGoede:
### 🔗 Linked issue
* #740 

### ❓ Type of change

### 📚 Description
Resolves #740 

Adds the function `loadLocaleMessages` which allows users to load locale messages for non-loaded locales when using lazy-load translations.

### 📝 Checklist

### DIFF
--- a/docs/content/docs/2.guide/7.lazy-load-translations.md
+++ b/docs/content/docs/2.guide/7.lazy-load-translations.md
@@ -192,3 +192,23 @@ export default defineNuxtConfig({
   }
 })
 ```
+
+
+## Using translations of non-loaded locale
+
+As only the current locale translations are loaded you have to manually load a locale to be able to use its translations. 
+
+Nuxt i18n extends Vue i18n to provide the `loadLocaleMessages` function to manually load locale messages, the example below demonstrates its usage.
+
+```ts
+const { loadLocaleMessages, t } = useI18n()
+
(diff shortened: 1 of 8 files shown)


kazupon (approved):
Thanks!
Oh, I was not aware of the v7 issue. 😅 
This feature is useful and we should be included! ✨

PR #2801, opened 2024-02-16
chore!: remove deprecated legacy dynamic route params

BobbieGoede:
### ❓ Type of change

### 📚 Description
The deprecation of using `nuxtI18n` with `definePageMeta` was a bit soft, as we kept using the same key internally for `setI18nParams` I can imagine there may have been users ignoring the warning. I changed the `nuxtI18n` key to `nuxtI18nInternal`, this will break the functionality for projects using the deprecated functionality and better communicate that this is an internal key. 

At some point we may or may not move away from using the `meta` key on routes, this is already the case when having `experimental.switchLocalePathLinkSSR` feature enabled.

### 📝 Checklist


kazupon:
Thank you!

I had planned to drop features in v8.1 that I had marked as deprecated prior to the v8.0 release.
However, nuxt i18n versioning is in semver, and according to semver policy that is equivalent to a major change.
And also, the recent OSS semver for automatic tracking tools like renovate and dependbot, as well as release tools.

So I believe that if we drop deprecation, we should release as v9.

There may be other deprecations in nuxt i18n.
I think we should pickup all of them once and drop them and release v9.

BobbieGoede:
>However, nuxt i18n versioning is in semver, and according to semver policy that is equivalent to a major change.

>There may be other deprecations in nuxt i18n.
I think we should pickup all of them once and drop them and release v9.

I agree, I'll label it v9 and set it to draft. It's good to be careful with breaking changes, I think with some planning it can be combined with some other changes.

Maybe I'll open an issue/discussion for v9 in the coming days to track which issues/changes could be included so users can give feedback or suggest changes.

kazupon:
> Maybe I'll open an issue/discussion for v9 in the coming days to track which issues/changes could be included so users can give feedback or suggest changes.

Good idea! 👍 
Let's open the issue/discussion!

BobbieGoede:
@kazupon 
I have updated this branch and changed the relevant docs in the v9 documentation, originally this PR also added logs whenever the old `nuxtI18n` meta key were set by users but since we have the deprecation warning in v8 and the added migration docs I don't think those logs are necessary.

Related to this feature, should we also remove `experimental.switchLocalePathLinkSSR` and have it enabled by default? It would clean up the implementation further and will be easier to document how to get proper dynamic route parameter for SSR projects. This change would be a separate PR.

PR #2734, opened 2024-01-26
[WIP] refactor: pass generated options through `runtimeConfig`

BobbieGoede:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description
This is a work in progress, I want to improve the type in `ModulePublicRuntimeConfig` so users know which values cannot be configured using `runtimeConfig`, for example those used during route generation.

By using `runtimeConfig` to pass the generated options we can reduce the complexity of our generated options templates (`#build/i18n.options.mjs` and `#internal/i18n/options.mjs`) and will make it easier for us to add `runtimeConfig` support for options used exclusively during runtime ([`detectBrowserLanguage`](https://github.com/nuxt-modules/i18n/pull/2576) for example).

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #2828, opened 2024-03-05
refactor: pass generated options through `runtimeConfig`

BobbieGoede:
### ❓ Type of change

### 📚 Description
Replaces #2734 

I would still like to improve the type in `ModulePublicRuntimeConfig` so users know which values cannot be configured using `runtimeConfig`, for example those used during route generation, should I mark these `@internal` using JSDocs?

By using `runtimeConfig` to pass the generated options we can reduce the complexity of our generated options templates (`#build/i18n.options.mjs` and `#internal/i18n/options.mjs`) and will make it easier for us to add `runtimeConfig` support for options used exclusively during runtime ([`detectBrowserLanguage`](https://github.com/nuxt-modules/i18n/pull/2824) for example).

I had to add `configLocales` to the runtime config as `locales` was already used for runtime domain support, we should probably merge these configurations for v9.

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -17,7 +17,7 @@ import { setupAlias } from './alias'
 import { setupPages } from './pages'
 import { setupNitro } from './nitro'
 import { extendBundler } from './bundler'
-import { generateI18nPageTypes, generateI18nTypes, generateLoaderOptions } from './gen'
+import { generateI18nPageTypes, generateI18nTypes, generateLoaderOptions, simplifyLocaleOptions } from './gen'
 import {
   NUXT_I18N_MODULE_ID,
   DEFAULT_OPTIONS,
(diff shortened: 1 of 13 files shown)


kazupon:
> I would still like to improve the type in `ModulePublicRuntimeConfig` so users know which values cannot be configured using `runtimeConfig`, for example those used during route generation, should I mark these `@internal` using JSDocs?

Yes!
We should mark as `@internal` and  summary note for developer.

src/module.ts
```diff
@@ -140,8 +140,20 @@ export default defineNuxtModule<NuxtI18nOptions>({
...
+      defaultDirection: options.defaultDirection,
+      strategy: options.strategy,
+      lazy: options.lazy,
+      rootRedirect: options.rootRedirect,
+      routesNameSeparator: options.routesNameSeparator,
+      defaultLocaleRouteNameSuffix: options.defaultLocaleRouteNameSuffix,
+      skipSettingLocaleOnNavigate: options.skipSettingLocaleOnNavigate,
+      differentDomains: options.differentDomains,
+      trailingSlash: options.trailingSlash,
+      configLocales: options.locales,
```
kazupon, line 156: Hmm 🤔 
I would also prefer `locales` instead of `configLocales` if possible.
It seems that `locales` have been introduced by this PR.
https://github.com/nuxt-modules/i18n/pull/2446

In v9, we should also be organized runtimeConfig.
`locales` should be `domainLocales` and `configLocales` should be `locales`.

@BobbieGoede 
What do you think?

BobbieGoede: >In v9, we should also be organized runtimeConfig.
locales should be domainLocales and configLocales should be locales.

Agreed, that sounds good! I have added a reference in the roadmap discussion https://github.com/nuxt-modules/i18n/discussions/2814

kazupon: Thanks! Okay, I'll merge this PR!

kazupon (approved):

PR #2838, opened 2024-03-14
feat: experimental feature `switchLocalePathLinkSSR` with `<SwitchLocalePathLink>` component

### LINKED ISSUE #2798: useLocaleHead prerenders properly. useSwitchLocalePath does not
### Environment
@nuxtjs/i18n@8.1.1

- Operating System: Linux
- Node Version:     v20.6.1
- Nuxt Version:     3.9.1
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@10.1.0
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

I'll be glad to make it if there's not a clue

### Describe the bug

Having a dinamic route page and a language switcher, while **useLocaleHead generates the right urls, useSwitchLocalePath does not**. It gets, what I assume, is the default slug before updating the slug with useSetI18nParams.

That leads into the crawler having wrong urls and forcing me to ignore prerender on nitro block.
While this do a proper build it does not escalate on my application and is just for debug purposes. I will be dealing with thousand urls soon.

Once the server is started, this urls are fixed, but they do appear wrong on the dist files

[Related](https://github.com/nuxt-modules/i18n/issues/2097)?

### Additional context

_No response_

### Logs

_No response_

BobbieGoede:
### 🔗 Linked issue
* #2798

### ❓ Type of change

### 📚 Description
Resolves #2798 

This adds a component `SwitchLocalePathLink` which is not much different from passing `switchLocalePath` to  `NuxtLink` except the resulting link is wrapped in comments `<!--nuxt-i18n-slp-[locale]-->...<!--/nuxt-i18n-slp-->`.

While `setI18nParams` currently works, it actually doesn't render the correct links during SSR depending on when and where it is used (as a warning describes in the docs), I have added an experimental feature that aims to resolve this.

The following functionality is enabled with the `experimental.switchLocalePathLinkSSR` property:

The `setI18nParams` is changed to use a `useState` ref instead of the `nuxtI18n` meta property on the route, this would normally cause a hydration mismatch (lang switcher rendered before `setI18nParams` is called, while the client will have the correct data at hydration) which is why we couldn't use this before. 

Using the wrapping comments to find the links with their locale in the rendered DOM at the end of SSR, the URL is replaced with an up to date localized path returned from `switchLocalePath`.

The implementation feels hacky to me but it seems to work fine, if there is a better way to do this please let me know! 

### 📝 Checklist

### DIFF
--- a/docs/content/docs/2.guide/8.lang-switcher.md
+++ b/docs/content/docs/2.guide/8.lang-switcher.md
@@ -77,10 +77,13 @@ Dealing with dynamic route parameters requires a bit more work because you need 
 During SSR it matters when and where you set i18n parameters, since there is no reactivity during SSR.
 
 :br :br
-Components which have already been rendered do not update by changes introduced by pages and components further down the tree. Instead, these links are updated on the client side, so this should not cause any issues.
+Components which have already been rendered do not update by changes introduced by pages and components further down the tree. Instead, these links are updated on the client side during hydration, in most cases this should not cause issues.
 
 :br :br
 This is not the case for SEO tags, these are updated correctly during SSR regardless of when and where i18n parameters are set.
+
+:br :br
+Check out the experimental [`switchLocalePathLinkSSR`](/docs/options/misc#experimental) feature, which combined with the [`<SwitchLocalePathLink>`](/docs/api/components#switchlocalepathlink) component, correctly renders links during SSR regardless of where and when it is used. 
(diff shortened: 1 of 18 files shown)


kazupon (approved):
That looks good for me.

PR #2847, opened 2024-03-18
feat: experimental auto import translation functions

### LINKED ISSUE #1644: [next] Auto-import of `t` in `<script setup>`

Using `$t` is common when e.g. building meta-tags. Always writing `{ t: $t } = useI18n()` feel a bit tedious after a while

Ideally: Full intellisense + "using" `$t` out of the box in `<script setup>` without writing above mentioned line.

* Writing it - works for now

BobbieGoede:
### 🔗 Linked issue
* #1644

### ❓ Type of change

### 📚 Description
Resolves #1644 

Adds `experimental.autoImportTranslationFunctions`, which when enabled adds `$t`, `$d` and `$n` to `globalThis`.

I did add the option to the docs but it's the bare minimum, not sure where and how to describe it in a clearer way 🤔, I suppose it's not that big of a deal while it's an experimental feature?

### 📝 Checklist


src/runtime/routing/extends/i18n.ts
```diff
@@ -119,6 +125,20 @@ export function extendI18n<Context = unknown, TI18n extends I18n = I18n>(
 
     const globalComposer = getComposer(i18n)
 
+    if (experimental.autoImportTranslationFunctions) {
+      globalThis.$t = function (...args: unknown[]) {
```
kazupon, line 129: I think memory leaks will be occure with exporting to `globalThis`.

Especially on the server side, `setup` of `defineNuxtPlugin` is executed on every request. This means that `extendI18n` will be executed multiple times.

we need to release those functions assigned to `globalThis`.

A good way to relase them is to use `onScopeDispose`. 
https://vuejs.org/api/reactivity-advanced.html#onscopedispose

`extendI18n` creates a scope with `effectScope`. You can detach the scope from `globalThis` when it is destroyed, so we need to check or verify when `onScopeDispose` is called on the server side.

kazupon:
Great work!
I am glad you have improved nuxt i18n dx!

Another implementation way for auto impot is to use AST in the bundler transform, which implements nuxt auto-import, or macros such as unplugin-auto-import, or like Vue 3 `defineProps` macro way, so we can  generate `useI18n` executable code.
- auto-import: https://github.com/unplugin/unplugin-auto-import
- vue-macro: https://github.com/vue-macros/vue-macros

This method is advanced and difficult to implement, but by injecting the transformed code into the script setup, we don't have worry about releases.

src/runtime/routing/extends/i18n.ts
BobbieGoede: Ah yes, I had a feeling this could cause leaks.

I'm also worried people will use these translation functions outside of setup in unexpected places which may cause further leaks or unexpected behavior.

BobbieGoede:
>Another implementation way for auto impot is to use AST in the bundler transform, which implements nuxt auto-import, or macros such as unplugin-auto-import, or like Vue 3 defineProps macro way, so we can generate useI18n executable code.

>This method is advanced and difficult to implement, but by injecting the transformed code into the script setup, we don't have worry about releases.

Ah yes! I never considered this approach, I'll close this PR for now and see if I can work out a solution using this method. This will give us more control where this functionality is injected, and I'm interested in learning more about this anyway 😄.

PR #2862, opened 2024-03-20
[WIP] feat: detect browser language from `localStorage`

BobbieGoede:
### ❓ Type of change

### 📚 Description
This is still a work in progress!

Adds `detectBrowserLanguage.useLocalStorage` which can be set to `string | undefined`, when set to a string this will be used to attempt to retrieve stored language in `localStorage` on the client-side.

This could potentially be useful for projects that want to store language settings but are not able (or don't want to) to use cookies, as is the case for https://github.com/elk-zone/elk.

I think we may want to consider a similar pattern as the `localeDetector` we currently use for the server side translations feature.

### 📝 Checklist

### DIFF
--- a/src/runtime/internal.ts
+++ b/src/runtime/internal.ts
@@ -66,19 +66,31 @@ export function parseAcceptLanguage(input: string): string[] {
   return input.split(',').map(tag => tag.split(';')[0])
 }
 
-export function getBrowserLocale(): string | undefined {
+export function getBrowserLocale(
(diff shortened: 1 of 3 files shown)


BobbieGoede:
Closing as I think the solution for this use case can and should be resolved differently, based on recommendations by Daniel and Joaquin I'll explore a different approach for SSG, inserting translations in the page while detecting and translating during hydration client-side, though I'm not sure yet when I will have time to work on it.

PR #2877, opened 2024-03-24
fix: sync cookie when setting locale

### LINKED ISSUE #2829: detectBrowserLanguage doesn't work
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.10.3
- CLI Version:      3.10.1
- Nitro Version:    2.9.1
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      devtools, modules, i18n
- Runtime Modules:  @nuxtjs/i18n@8.1.1
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-xkxgkk?file=nuxt.config.ts,pages%2Findex.vue

### Describe the bug

The cookie is never set when hit root, whether it's default `i18n_redirected` or custom `test_cookie` in my example

### Additional context

_No response_

### Logs

_No response_

BobbieGoede:
### 🔗 Linked issue
* #2829 

### ❓ Type of change

### 📚 Description
Resolves #2829 

Please note that this actually changes the cookie behavior slightly, this is why I set this RP to draft status.

These changes ensure that the cookie is always set and up to date (if configured of course), even when no redirection takes place necessarily. I feel like this would only ensure the correct locale is used even if the user has not been redirected (e.g. entering the website on a page in which redirection would occur anyways), are there any downsides to this change?

If we want to keep this change we may need to adjust the documentation a bit, I haven't looked into this yet.

### 📝 Checklist

### DIFF
--- a/docs/content/docs/2.guide/5.browser-language-detection.md
+++ b/docs/content/docs/2.guide/5.browser-language-detection.md
@@ -23,7 +23,7 @@ For better SEO, it's recommended to set `redirectOn` to `root` (which is the def
 
 Browser language is detected either from `navigator` when running on client-side, or from the `accept-language` HTTP header. Configured `locales` (or locales `iso` and/or `code` when locales are specified in object form) are matched against locales reported by the browser (for example `en-US,en;q=0.9,no;q=0.8`). If there is no exact match for the full locale, the language code (letters before `-`) are matched against configured locales.
 
-To prevent redirecting users every time they visit the app, **Nuxt i18n module** sets a cookie after the first redirection. You can change the cookie's name by setting `detectBrowserLanguage.cookieKey` option to whatever you'd like, the default is _i18n_redirected_.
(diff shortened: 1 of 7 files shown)


kazupon:
> These changes ensure that the cookie is always set and up to date (if configured of course), even when no redirection takes place necessarily. I feel like this would only ensure the correct locale is used even if the user has not been redirected (e.g. entering the website on a page in which redirection would occur anyways), are there any downsides to this change?

Thanks for your fixing!
I believe that your fix is no downside.
In fact, your change has passed testing and has not caused any regression.
Let’s merge this PR!

kazupon (approved):

makamekm:
It looks like this PR has broken the package with infinity redirects in some browsers like Opera or Yandex.

BobbieGoede:
> It looks like this PR has broken the package with infinity redirects in some browsers like Opera or Yandex.

Can you open an issue with a minimal reproduction? 🙏

PR #2882, opened 2024-03-26
fix: options passed with `installModule` are overwritten

### LINKED ISSUE #2809: @nuxtjs/i18n not picking up vueI18n config when installed via Nuxt Kits `installModule`
### Environment

- Operating System: `Darwin`
- Node Version:     `v20.10.0`
- Nuxt Version:     `3.9.1`
- CLI Version:      `3.10.0`
- Nitro Version:    `2.8.1`
- Package Manager:  `yarn@1.22.19`
- Builder:          `-`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-y1vnuv

### Describe the bug

We are trying to install `@nuxtjs/i18n` from a custom module via [Nuxt Kits `installModule` method](https://nuxt.com/docs/api/kit/modules#installmodule). So far @nuxtjs/i18n is not picking up the config file configured via the `vueI18n` option. You can see the empty `VueI18nConfigPaths` array in the reproduction.

### Additional context

_No response_

### Logs

```shell
@nuxtjs/i18n:module options {
  experimental: { localeDetector: '' },
  bundle: {
    compositionOnly: true,
    runtimeOnly: false,
    fullInstall: true,
    dropMessageCompiler: false
  },
  compilation: { jit: true, strictMessage: true, escapeHtml: false },
  customBlocks: { defaultSFCLang: 'json', globalSFCScope: false },
  vueI18n: '/home/projects/bobbiegoede-nuxt-i18n-starter-y1vnuv/modules/example/config/i18n/options.ts',
  locales: [],
  defaultLocale: '',
  defaultDirection: 'ltr',
  routesNameSeparator: '___',
  trailingSlash: false,
  defaultLocaleRouteNameSuffix: 'default',
  lazy: false,
  langDir: 'locales/',
  rootRedirect: null,
  detectBrowserLanguage: false,
  differentDomains: false,
  baseUrl: '',
  dynamicRouteParams: false,
  customRoutes: 'page',
  pages: {},
  skipSettingLocaleOnNavigate: false,
  types: 'composition',
  debug: true,
  parallelPlugin: false
} +0ms
[nuxt] i18n:registerModule: 0.02ms                                                                                                          07:35:34
  @nuxtjs/i18n:module localeInfo [] +1ms
  @nuxtjs/i18n:module VueI18nConfigPaths [] +40ms
  @nuxtjs/i18n:module vueI18nPath for auto-import /home/projects/bobbiegoede-nuxt-i18n-starter-y1vnuv/node_modules/vue-i18n/dist/vue-i18n.mjs
```

### LINKED ISSUE #2874: Configured locales get overridden when installing @nuxtjs/i18n using Nuxt Kit's installModule utility
### Local environment:
- Operating System: Linux
- Node Version: v20.11.1
- Nuxt Version: 3.11.1
- CLI Version: 3.11.1
- Nitro Version: 2.9.4
- Package Manager: npm@10.5.0
### StackBlitz reproduction environment:
- Operating System: Linux
- Node Version: v18.18.0
- Nuxt Version: 3.11.1
- CLI Version: 3.11.1
- Nitro Version: 2.9.4
- Package Manager: npm@10.2.3

### Reproduction

https://stackblitz.com/edit/github-8ok7gy-kkmqkq

### Describe the bug

I am adding new pages from my own module as documented [here](https://i18n.nuxtjs.org/docs/guide/extend-pages).
I am installing and configuring `@nuxtjs/i18n` from my own module after I added these new pages as documented [here](https://nuxt.com/docs/guide/going-further/modules#using-other-modules-in-your-module).

The locales that I configured in the `moduleOptions` passed to the `installModule` function are not present in the locales of the `i18n` instance.

### Additional context

Use case: I want to add and configure new pages in a headless CMS, I want to translate the routes and I want to configure these route translations in the headless CMS as well.

I debugged the issue to the [`applyLayerOptions` function](https://github.com/nuxt-modules/i18n/blob/eafe06bb561154b67ebbdb75acf8bcb00c8b1df2/src/layers.ts#L73). This is where `options.locales` is overridden by `mergedLocales`.

When generating `mergedLocales` in the [`mergeLayerLocales` function](https://github.com/nuxt-modules/i18n/blob/eafe06bb561154b67ebbdb75acf8bcb00c8b1df2/src/layers.ts#L93), I can see that `options.locales` is made sure not empty, but I don't see it being used after that since the [`mergeConfigLocales` function](https://github.com/nuxt-modules/i18n/blob/eafe06bb561154b67ebbdb75acf8bcb00c8b1df2/src/layers.ts#L110) does not consider `options`.

### Logs

_No response_

BobbieGoede:
### 🔗 Linked issue
* #2874 
* #2809

### ❓ Type of change

### 📚 Description
Resolves #2874 
Resolves #2809

As options passed to the module `setup` already contain the merged layer options, the only way we can retrieve the `locales` and `vueI18n` options are by looking for absolute paths and whether they do not match layer configurations.

And since we can't know which module installs our module, and we also don't know which layer has the installer module configured, we don't know the priority at which we should merge these options. So far it seems that the approach in other modules is to treat `installModule` as having a higher priority than the project layer.

I added support for this since it seems there is demand for this functionality, but I honestly think this will cause issues if a lot of modules end up doing this, extending projects with layers is a much more predictable and safe approach and as such I have recommended this in the documentation as well.

### 📝 Checklist


kazupon (approved):
Great work!
LGTM!

PR #2887, opened 2024-03-28
fix: `defineI18nRoute` macro transformed inside `<template>`

BobbieGoede:
### ❓ Type of change

### 📚 Description
This is very specific, I don't think anyone ran into this bug 😅 

While refactoring/merging tests I noticed that the text `defineI18nRoute(false)` inside of `<template>` in `specs/fixtures/basic/pages/index.vue` was being transformed and replaced to `false && /*#__PURE__*/ defineI18nRoute(false)`. This should only happen inside the script contents.

Maybe this PR changes the transform plugin more than it should, if anything is off let me know!

### 📝 Checklist


kazupon (approved):
Good catch!
Thanks!

Issue #2790, opened 2024-02-15
Strategy prefix with detectBrowserLanguage not redirecting correctly on root in Static Site Generation


FabianRzepka:
### Environment

- Operating System: Windows_NT
- Node Version:     v20.9.0
- Nuxt Version:     3.10.1
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  pnpm@8.9.2
- Builder:          -
- User Config:      app, vue, ssr, devtools, modules, image, googleFonts, i18n
- Runtime Modules:  @nuxtjs/i18n@8.1.0, @nuxtjs/tailwindcss@6.11.4, @nuxtjs/google-fonts@3.1.3, @nuxt/image@1.3.0, nuxt-icon@0.6.8, nuxt-headlessui@1.1.5
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-cs5xbf?file=nuxt.config.ts

On local work fine.

`site.com/` --> click ` /nl `to save cookie -->` site.com/` redirect to `site.com/nl`

npm run generate
npx serve .output/public

`site.com/` --> click ` /nl `to save cookie --> `site.com/` redirect to default lang `site.com/en`

### Describe the bug

After generating a static page with  `pnpm run generate` and uploading it to hosting the application does not detect the existing cookie or browser language, in the case of (root) site.com it always redirects to defaultLocale (en) site.com/en.
When running the site locally, `pnpm run dev` or run `build `works correctly. 

in addition, the "prefix" mode adds an error in the console after entering the root path. 

![image](https://github.com/nuxt-modules/i18n/assets/67225684/be24c061-0fb6-496f-a837-3d133556f156)

### Additional context

related? old task #2507 

_No response_

BobbieGoede:
We will need to reimplement detection + redirection specifically for statically generated projects, the reason why there are issues right now is that most of the detection relies on server-side behavior for things to work without hydration mismatches.

For SSG we will need to figure out a way to detect and redirect before hydration happens on page load, this also means we can't rely on functionalities provided by Nuxt for that to work 🤔

I'm sorry this has still not been resolved but it is a tough one and I am working on it over the long term.

PR #2894, opened 2024-04-02
fix: ignore root when prerendering using `strategy: 'prefix'`

### LINKED ISSUE #2855: No match found for location with path "/"
### Environment

"nuxt": "^3.11.0",
"@nuxtjs/i18n": "^8.2.0",

### Reproduction

this is the export default defineNuxtConfig({

```
    ['@nuxtjs/i18n',
      {
        strategy: "prefix",
        lazy: false,
        langDir: "lang/",
        fallbackLocale: ['en', 'nl','fr','fa'],
        detectBrowserLanguage: {
          fallbackLocale: ['en', 'nl','fr','fa'],
          useCookie: true,
          cookieKey: lang_cookie_name,
          cookieSecure: true,
          redirectOn: 'root',
          alwaysRedirect: true
        },

```

### Describe the bug

the bug is in npm run dev get the error [Vue Router warn]: No match found for location with path "/".

this is very annoying when nuxt goes to npm run generate.

### Additional context

this is very annoying when nuxt goes to npm run generate.

### Logs

```shell
[Vue Router warn]: No match found for location with path "/"
```

BobbieGoede:
### 🔗 Linked issue
* #2855
* Related #2790

### ❓ Type of change

### 📚 Description
Resolves #2855 

Not sure how to test this exactly 🤔 

These changes removes/ignores the `/` entry route added by nitro during prerendering and adds localized entry pages to use instead. This also excludes the `/` route from the generated build, which I think is/was the original behavior when running generate with strategy prefix (currently without these changes this throws an error). 

When not prerendering we re-add the index page (not localized) so that requests on root can be redirected. 

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -198,6 +198,25 @@ export default defineNuxtModule<NuxtI18nOptions>({
     }
 
     /**
+     * ignore `/` during prerender when using prefixed routing
+     */
+
+    if (options.strategy === 'prefix' && nuxt.options._generate) {
+      const localizedEntryPages = normalizedLocales.map(x => ['/', x.code].join(''))
+      nuxt.hook('nitro:config', config => {
+        config.prerender ??= {}
+
+        // ignore `/` which is added by nitro by default
+        config.prerender.ignore ??= []
+        config.prerender.ignore.push(/^\/$/)
+
+        // add localized routes as entry pages for prerendering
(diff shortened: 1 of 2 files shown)


kazupon (approved):
Thanks!
LGTM!

PR #2906, opened 2024-04-12
ignore this pull request; see #2914 instead

### LINKED ISSUE #1161: Different domains with multiple languages each

DarthGigi:
Check out #2914 instead.


ghost (approved):

src/runtime/internal.ts
```diff
@@ -319,15 +319,17 @@ export function detectBrowserLanguage(
...
+export function getHost(nuxtApp: NuxtApp) {
   let host: string | undefined
   if (import.meta.client) {
     host = window.location.host
   } else if (import.meta.server) {
     const header = useRequestHeaders(['x-forwarded-host', 'host'])
 
     let detectedHost: string | undefined
-    if ('x-forwarded-host' in header) {
+    if (nuxtApp?.ssrContext?.event?.context.storeHost) {
```
e-kravtsov, line 330: this will fire only if you add a server plugin like
export default defineNitroPlugin((nitro) => {
  nitro.hooks.hook('request', (event) => {
    event.context.storeHost = 'host-you-want-to-use.com' // or get the host from event manually;
  });
});
It gives more control and easier for local development, if you don't want to use headers, or want to use your own header for any reason
Maybe we should add this to the docs

DarthGigi: Where in the docs would be a good place to add this?

e-kravtsov: maybe here https://github.com/nuxt-modules/i18n/pull/2906/commits/5b50a5b5b8063f08020a80fca472d22b87f59959#diff-c8b1c956c008039ba7f8e88d7d78dfced016488cafd0f6076c7e7ed456f0b230R174

DarthGigi:
I think I screwed something up 😬...

DarthGigi:
Please check out the new pull request: #2914

PR #3017, opened 2024-07-05
refactor: restructure extend vue-i18n

BobbieGoede:
### ❓ Type of change

### 📚 Description
We still have some split logic between this module and what was originally in `vue-i18n-routing`, because of this we were extending `vue-i18n` in two places. 

After some reorganizing I noticed that `onExtendExportGlobal` and `onExtendVueI18n` were actually doing the same thing but with different variable names 🤔 possibly `__composerExtend` did the same as well, after merging and generalizing these extends it appears to still be working.

I feel like my changes should have broken something.. 😂 Either the refactored code was interchangeable or it contains untested logic 🤔

### 📝 Checklist

### DIFF
--- a/src/runtime/plugins/i18n.ts
+++ b/src/runtime/plugins/i18n.ts
@@ -1,15 +1,8 @@
-import { computed } from 'vue'
+import { ref } from 'vue'
 import { createI18n } from 'vue-i18n'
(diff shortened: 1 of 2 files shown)


kazupon (approved):
Good Refactoring!
Thank you always! ❤️

PR #3021, opened 2024-07-09
refactor: separate switchLocalePath SSR feature into plugin

BobbieGoede:
### ❓ Type of change

### 📚 Description
I feel like there is a lot going on in the i18n plugin, after looking at Nuxt's source code I noticed some features are separated into plugins, so I tried to do the same here 😅 

I also tried another way of reorganizing the plugin logic here https://github.com/nuxt-modules/i18n/pull/3017, but that's a bit more involved.

If refactors like these actually make things harder to read, do let me know, I'm too familiar with the codebase now to notice if that is the case 😂

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -217,6 +217,7 @@ export default defineNuxtModule<NuxtI18nOptions>({
 
     // for core plugin
     addPlugin(resolve(runtimeDir, 'plugins/i18n'))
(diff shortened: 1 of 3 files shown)


kazupon (approved):
Good refactoring!
Thanks!

Issue #2857, opened 2024-03-18
Narrow typed locale identifiers


warflash:
### Describe the feature

`const { locale, locales } = useI18n()`

Currently `locale` is typed as `string` and `locales` is `string[]`.
The suggestion is to use tighter types like `locale -> "en" | "de"` in the case of a config like:
```ts
	i18n: {
		locales: [
			{ code: "en", iso: "en-US", file: "en.json" },
			{ code: "de", iso: "de-DE", file: "de.json" },
		],
	},
```

That would make type safe access to typed records possible. e.g.

```ts
const MIN_SEARCH_LEN_BY_LOCALE = {
	en: 3,
	de: 3,
	ko: 1,
}

const { locale, locales } = useI18n()

const minLen = MIN_SEARCH_LEN_BY_LOCALE[locale] 
```

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution) (The contribution guideline of nuxt-modules/i18n is compliant with Nuxt too).
- [X] Check existing [discussions](https://github.com/nuxt-modules/i18n/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

BobbieGoede:
Should be resolved by #3025, which will be included in v9

warflash:
Thanks! 💚

PR #3025, opened 2024-07-12
feat: generate `Locale` type based on configuration

BobbieGoede:
### 🔗 Linked issue
* #2857 

### ❓ Type of change

### 📚 Description
This changes type generation to narrow the `Locale` type to the configured `locales`, this way functions that accept a `locale` parameter (such as our composables) will have proper autocompletion for all configured locales.

@kazupon 
There's still an issue/quirk with the type, I'm don't know how to narrow the `locale` property on the `Composer` type in the same way, so currently the following would cause a type error:

```typescript
const { locale } = useI18n()
const switchLocalePath = useSwitchLocalePath()

switchLocalePath(locale.value) // Argument of type 'string' is not assignable to parameter of type 'Locale'.
```

Maybe you know how to resolve this? I don't think this should be merged while this issue is present 😅

Also working on this https://github.com/BobbieGoede/i18n/pull/49.

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "sucrase": "^3.35.0",
     "ufo": "^1.3.1",
     "unplugin": "^1.10.1",
-    "vue-i18n": "^10.0.0-beta.1",
+    "vue-i18n": "^10.0.0-beta.4",
     "vue-router": "^4.4.0"
   },
   "devDependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -21,7 +21,7 @@ importers:
         version: 10.0.0-beta.1
       '@intlify/unplugin-vue-i18n':
         specifier: ^5.0.0-beta.3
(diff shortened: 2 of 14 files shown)


kazupon:
@BobbieGoede 
Hmm, 🤔 
The `locale` of the current `Compose` is resolved by the type parameter as `locale: WritableComputedRef<Locales>`.

https://github.com/intlify/vue-i18n/blob/c03fec09f9c3f9f16f8773295523a8d693e9ce0b/packages/vue-i18n-core/src/composer.ts#L1268-L1296

To support nuxt-i18n use cases, this could be extended more on the Vue I18n side.

e.g. `locale: IsNever<CustomLocales> extends true ? WritableComputedRef<Locales> : WritableComputedRef<CustomLocale>` .

It may be possible to dynamically extend the locale.

BobbieGoede:
@kazupon 
I have a possible solution here https://github.com/intlify/vue-i18n/pull/1890, I have updated this PR to make use of that solution. Let me know if this works for you locally or if a different approach is preferred.

kazupon (approved):
Okay, I could work it with vue-i18n v10 latest version!

PR #3034, opened 2024-07-18
fix: generated module augmentation should use `@intlify/core-base`

BobbieGoede:
### ❓ Type of change

### 📚 Description
After some more testing it seems necessary to augment `@intlify/core-base`, as the types will not be inferred when using `@intlify/core`. I suppose this makes as this is where the other packages import from? 🤔 I will open a PR to correct the JSDoc/comment in `vue-i18n` as well (https://github.com/intlify/vue-i18n/pull/1898).

This behavior differs from the types in the playground within this repository, in which case augmenting `@intlify/core` works as well, I'm not entirely sure why that is.

### 📝 Checklist

### DIFF
--- a/src/gen.ts
+++ b/src/gen.ts
@@ -199,7 +199,7 @@ declare module 'vue-i18n' {
   interface VueI18n extends NuxtI18nRoutingCustomProperties<${resolvedLocaleType}> {}
 }
 
-declare module '@intlify/core' {
+declare module '@intlify/core-base' {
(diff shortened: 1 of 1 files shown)


kazupon:
> I suppose this makes as this is where the other packages import from?

Yes!
`GeneratedTypeConfig` is declared at `@intlify/core-base`, so we need to argument from it.

kazupon (approved):

Issue #2983, opened 2024-06-20
[BUG?]: Locale resolver scans `srcDir` instead of `rootDir`


Gerbuuun:
In Nuxt 4 the `srcDir` will be `app` by default instead of being the same as the `rootDir`.
I want to put my `lang` directory outside of the `app` directory because I also use translations server side.

In the module setup it will scan the `srcDir` for the translation files but they are located outside of it.
https://github.com/nuxt-modules/i18n/blob/cddb8b31825405f08f569f882b6d80284c263ca7/src/module.ts#L184-L188

(The `experimental.localeDetector`, in contrast, does scan the `rootDir`)
https://github.com/nuxt-modules/i18n/blob/cddb8b31825405f08f569f882b6d80284c263ca7/src/nitro.ts#L138

I tested the module with only changing the `nuxt.options.srcDir` to `nuxt.options.rootDir` and it works fine.
I hope this is the easy fix I think it is.

cc @danielroe

Gerbuuun:
Just realized I can use a relative path in the config.
```ts
i18n: {
  langDir: '../lang',
}
```

A point on consistency could still be made though.

kazupon:
Thank you for your reporting!
We are now going to work on nuxt-i18n v9. At that time we will also test it with nuxt4. Please wait!

bernhardberger:
This is also an issue when using local modules with Nuxt4 directory structure, I haven't found a way around this yet tbh:

```
ERROR  Pre-transform error: Failed to resolve import "../app/modules/mylocalmodule/src/runtime/lang/de-AT.json" from ".nuxt/i18n.options.mjs". Does the file exist? (x7)
```

It tries to look up the module under `/app/modules` which should actually be `/modules`.


I'd even argue that this is **already** broken vor v3 for people that change their `srcDir` manually to be different from `rootDir`. I just seems to happen that nobody actually does this in practice right now.

danielroe:
@bernhardberger Could you open an issue with a reproduction on https://github.com/nuxt/nuxt ? 🙏

BobbieGoede:
@Gerbuuun 
[...]

PR #3054, opened 2024-08-06
feat: configurable directory restructure

BobbieGoede:
### 🔗 Linked issue
* https://github.com/nuxt-modules/i18n/issues/2983

### ❓ Type of change

### 📚 Description
This is still a work in progress!

This changes the way files are resolved and can be enabled by setting `restructureDir: 'i18n'`. This is no longer a breaking change as it is disabled by default allowing users a path to migrate before updating to v9, we will have to make sure to set the default values for v9 before release!

With `restructureDir` set to `i18n` the directory structure would look like so:

```sh
app/
server/
i18n/
  locales/
    en.json
    ja.json
  i18n.config.ts
  localeDetector.ts
nuxt.config.ts
```

And `langDir` will have a default value of `locales`, perhaps we may want to rename this to `messages` instead?

This kind of tackles https://github.com/nuxt-modules/i18n/issues/2983, as with this feature enabled we would resolve from `<rootDir>/i18n/` for all i18n files.

### 📝 Checklist


kazupon:
Thanks!
Great works! 
This looks good to me because this PR supports Nuxt4 and i18n-specific directories.

It might be good to be able to change the top directory name, as this might cause conflicts with users and 3rd vendors. `restructure` may accept string instead of boolean.
The option name `restructureDir` might be better.

BobbieGoede:
@kazupon 
Thanks for the feedback! I have changed the option to `restructureDir` which accepts a string so users can configure the directory themselves.

We should set the default value in a separate PR as that would be a breaking change, but these changes should be fully compatible with v8, so users can opt into the restructure.

I've tried to update the v9 docs to already reflect the new default values, hopefully I haven't missed anything 😅

kazupon (approved):
Well done!
Thank you so much always 💚 !

Issue #2449, opened 2023-09-28
Language tags used in `locales`


xfq:
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.15.0`
- Nuxt Version:     `-`
- CLI Version:      `3.9.0`
- Nitro Version:    `-`
- Package Manager:  `unknown`
- Builder:          `-`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/nuxt-modules/i18n/blob/6394a48c0d349c24d4005d490df7ce1b6d3a17c3/docs/content/3.options/2.routing.md?plain=1#L40-L42

### Describe the bug

The name `iso` here is not quite accurate and I think it can be changed to something like `language`. If it is too late to update it due to backward compatibility concerns, at least the description in the docs here can be updated.

Both [navigator.language(s)](https://html.spec.whatwg.org/multipage/system-state.html#language-preferences) and [Accept-Language](https://www.rfc-editor.org/rfc/rfc9110#field.accept-language) refer to [BCP 47](https://www.rfc-editor.org/info/bcp47) and the [IANA Language Subtag Registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry), which is based on underlying standards, including the various parts of ISO 639 (639-1, 639-2, 639-3, etc.), ISO 15924 script codes, ISO 3166, and UN M.49 region codes.

### Additional context

Related information:

* https://www.w3.org/International/questions/qa-choosing-language-tags.en.html
* https://www.w3.org/International/articles/language-tags/index
* https://www.w3.org/International/questions/qa-accept-lang-locales.en.html
* https://www.w3.org/TR/international-specs/#lang_values

kazupon:
I think `iso` certainly is a mistake from a specification standard. but, nuxt i18n v8 is RC, we cannot do breaking changes can be made.
I think it would be better to put the details in the documentation, and then add `language` or `lang` while keeping `iso` compatibility.

BobbieGoede:
We can probably add `language` as a sort of 'alias' for `iso` after stable `v8` release and fully deprecate/replace `iso` at some point for `v9`. 

I remember having a hard time finding (or reading) docs describing the standards, using correct/common terminology and referencing the relevant information will definitely help me and other devs out!

PR #3055, opened 2024-08-08
fix: rename locale `iso` property to `language`

BobbieGoede:
### 🔗 Linked issue
* https://github.com/nuxt-modules/i18n/issues/2449

### ❓ Type of change

### 📚 Description
As discussed in https://github.com/nuxt-modules/i18n/issues/2449, this renames the `iso` key to `language` to better fit the usage of the property.

This PR should be backwards compatible but will log a warning when `iso` is being used instead of `language`. We should again remember to remove the backwards compatibility before releasing v9.

### 📝 Checklist


specs/basic_usage.spec.ts
```diff
@@ -141,7 +141,7 @@ describe('basic usage', async () => {
...
...JP", "rel": "alternate", "href": "/ja/nuxt-context-extension", "hreflang": "ja-JP" }, { "hid": "i18n-alt-nl", "rel": "alternate", "href": "/nl/nuxt-context-extension", "hreflang": "nl" }, { "hid": "i18n-alt-nl-NL", "rel": "alternate", "href": "/nl/nuxt-context-extension", "hreflang": "nl-NL" }, { "hid": "i18n-alt-kr", "rel": "alternate", "href": "/kr/nuxt-context-extension", "hreflang": "kr" }, { "hid": "i18n-alt-kr-KO", "rel": "alternate", "href": "/kr/nuxt-context-extension", "hreflang": "kr-KO" }, { "hid": "i18n-xd", "rel": "alternate", "href": "/nuxt-context-extension", "hreflang": "x-default" }, { "hid": "i18n-can", "rel": "canonical", "href": "/nuxt-context-extension" } ], "meta": [ { "hid": "i18n-og-url", "property": "og:url", "content": "/nuxt-context-extension" }, { "hid": "i18n-og", "property": "og:locale", "content": "en" }, { "hid": "i18n-og-alt-fr-FR", "property": "og:locale:alternate", "content": "fr_FR" }, { "hid": "i18n-og-alt-ja-JP", "property": "og:locale:alternate", "content": "ja_JP" }, { "hid": "i18n-og-alt-nl-NL", "property": "og:locale:alternate", "content": "nl_NL" }, { "hid": "i18n-og-alt-kr-KO", "property": "og:locale:alternate", "content": "kr_KO" } ] }"`
```
BobbieGoede: I'm not sure why this changed, I think the `i18n-alt-fr-FR` tag was added, shouldn't this have been there originally? 🤔

docs/content/docs/5.v9/2.guide/18.breaking-changes-in-v9.md
```diff
@@ -46,4 +46,8 @@ nuxt.config.ts
 
 Reasons for change
   1. Context - i18n files are used both server-side and client-side, using a dedicated `i18n/` folder in the root directory outside `app/` and `server/` makes more sense.
-  2. Clean - less clutter/fragmentation of i18n files, and should make resolving and loading files easier for us.
\ No newline at end of file
+  2. Clean - less clutter/fragmentation of i18n files, and should make resolving and loading files easier for us.
+
+## Locale `iso` renamed to `language`
+
+The `iso` property on a locale object has been renamed to `language` as the original name refers to ISO standards which describe valid language tags.
```
kazupon: We should add more explanation why we made this change, include PR link.
This would make it easier for users to understand.
https://github.com/nuxt-modules/i18n/issues/2449

specs/basic_usage.spec.ts
kazupon: `#locale-head` value is injected by `useHead`. `useHead` parameters is specified by resolved with `useLocaleHead`
https://github.com/nuxt-modules/i18n/blob/08638d703dbd11467b4159a86c8c7dfab36e70aa/specs/fixtures/basic_usage/pages/index.vue#L65-L76

This chaning is affected by this PR, or  maybe `useLocaleHead` has the issue. 🤔

kazupon (changes requested):
Thanks!
I've just reviewed your PR.
Please check it! 🙏

specs/basic_usage.spec.ts
BobbieGoede: I figured out why this happened and fixed it by moving the check/rename of the `iso` property (https://github.com/nuxt-modules/i18n/pull/3055/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R518-R524) earlier in the code. 

This is related to the way locales are normalized and merged if both string and object locales are configured in multiple layers. So the project layer had `fr` configured which was normalized to `{ code: 'fr', language: 'fr' }` while a layer had configured `{ code: 'fr', iso: 'fr-FR' }`, which is why both language tags ended up in the build.

BobbieGoede:
@kazupon 
I made changes based on your feedback, let me know if anything needs to be changed!

kazupon (approved):
Good job!
Thank you! 💚

PR #3063, opened 2024-08-18
refactor: detect browser language internals

BobbieGoede:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [x] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description
This refactors `detectBrowserLanguage` and `detectLocale`, I managed to remove some unused/dead logic that made the function look more complex than it really is and simplified the logging for legibility, it can likely be simplified even further.

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have added tests (if possible).
- [ ] I have updated the documentation accordingly.

PR #3067, opened 2024-08-19
feat: runtime logger formatting and debug verbose option

BobbieGoede:
### ❓ Type of change

### 📚 Description
This changes the runtime debug logging to use a simple logger which prefixes each log with a label: `[i18n:${label}]`, and changes the logging slightly to make it less verbose.

I have also changed the `debug` option to accept a string `'verbose'`, now when enabling debug logging the loaded messages objects will not be logged by default but instead only fully log for verbose logging.

The main purpose of this change is to make the debug logs a bit more consistent and allow for easier filtering (especially in the browser) but we could also explore adding basic filtering by label/namespace ourselves. 

### 📝 Checklist


kazupon:
Thank you!

we should switch from debug to consola.
https://github.com/unjs/consola

there are log level outputing in cosola. :)

BobbieGoede:
@kazupon 
You're right! I've replaced it with `consola` now but I'm not sure if this is the correct way to use it, for some reason `consola` won't get treeshaken out even if a project has `debug: false` 🤔

kazupon:
@BobbieGoede 
I think @danielroe or @pi0  know more about it. (sorry for metion 🙇 )

src/runtime/messages.ts
```diff
@@ -14,6 +15,18 @@ export type LocaleLoader<T = LocaleMessages<DefineLocaleMessage>> = {
...
 }
+
+/**
+ * Used for runtime debug logs only, cannot be imported due to
+ * this being a virtual file in both nuxt and nitro contexts
+ *
+ * NOTE: We could export the logger from a virtual file instead
+ */
+const debugLogger = /*#__PURE__*/ createConsola({ level: __DEBUG_VERBOSE__ ? 999 : 4 }).withTag('i18n')
+function createLogger(label: string) {
```
danielroe: Does this resolve it?

```suggestion
/* @__NO_SIDE_EFFECTS__ */
function createLogger(label: string) {
```

src/runtime/utils.ts
```diff
@@ -475,3 +469,11 @@ export function getNormalizedLocales(locales: string[] | LocaleObject[]): Locale
   }
   return normalized
 }
+
+/**
+ * Used for runtime debug logs only
+ */
+const debugLogger = /*#__PURE__*/ createConsola({ level: __DEBUG_VERBOSE__ ? 999 : 4 }).withTag('i18n')
+export function createLogger(label: string) {
```
danielroe:
```suggestion
/* @__NO_SIDE_EFFECTS__ */
export function createLogger(label: string) {
```

src/runtime/messages.ts
BobbieGoede: That doesn't seem to work either 🤔 

I'm using `/*#__PURE__*/` wherever the `createLogger` function is called to achieve the same though, from what I understand `/* @__NO_SIDE_EFFECTS__ */` would not work for projects using webpack right?

BobbieGoede:
Using the plugins used in https://github.com/vuetifyjs/nuxt-module as a reference, I think I have found a solution by using a virtual file that only imports consola if necessary, it comes with the added benefit of being able to import the logic in both the Nuxt and Nitro context.

I barely understand how this works so I honestly don't know if there are any drawbacks or if this could have been made with a simpler approach.

docs/content/docs/3.options/10.misc.md
```diff
@@ -47,10 +47,10 @@ If it can not detect Nuxt configuration changing, you may need to run `nuxi prep
 
 ## `debug`
 
-- type: `boolean`
+- type: `boolean | 'verbose'`
```
kazupon: if we are extending for `debug` for v8, we need to back-port for `v8` branch.

kazupon: I think Nuxt3 will still be used, so we should be back-ported to v8 as well.

src/runtime/internal.ts
```diff
@@ -49,20 +50,21 @@ export function parseAcceptLanguage(input: string): string[] {
 
 export function getBrowserLocale(): string | undefined {
   let ret: string | undefined
+  const logger = /*#__PURE__*/ createLogger('getBrowserLocale')
```
kazupon, line 53: We tag the logs with `createLogger` for each function, do we need to go that far? 😅 

I think it would be better to have a logger tagged per js/ts module.
I believe we can make debugging more efficient from currently by filtering the logs by module.

BobbieGoede: I agree it seems like a bit much, but most of the logs originally already had the function name in the log itself in each log statement (for example https://github.com/nuxt-modules/i18n/pull/3067/files/a44ce5dbcd40696e2cfd3db47b18464476a5cb45#diff-7755b40274cbd4ca5d224c6ed6953280e589110e9289595aeb246ff4a1c07448L57).

The primary reason for tagging them is so the actual code/logic is easier to read and understand (fewer multiline debug log statements, less word repetition). With these refactors I'm hoping to find patterns for simplification like in https://github.com/nuxt-modules/i18n/pull/3063.

Logging per module/file would definitely make more sense, and I think the fact that we had the function name in the logs probably means that `internal.ts` and utils.ts` are not descriptive enough, which we could solve by splitting these up further.

docs/content/docs/3.options/10.misc.md
BobbieGoede: I think we can make `v9` fully compatible with both Nuxt 3 and 4, and keeping the `v8` branch up to date is starting to get difficult 😅 We may need to decide to only do maintenance/hotfixes for `v8` soon..

kazupon: > keeping the v8 branch up to date is starting to get difficult 😅 We may need to decide to only do maintenance/hotfixes for v8 soon..

Yeah, you're right, the v8 branch should be hotfix only.
This `debug` options extending should only be included in the `main` branch, i.e. v9.
So, Let's remove it from the docs for v8.

src/runtime/internal.ts
kazupon: > The primary reason for tagging them is so the actual code/logic is easier to read and understand (fewer multiline debug log statements, less word repetition). With these refactors I'm hoping to find patterns for simplification like in #3063.

Okay, I could understand what you are trying to improve for logging. Let's continue to refactor and improve! :)

docs/content/docs/3.options/10.misc.md
BobbieGoede: >This debug options extending should only be included in the main branch, i.e. v9.
So, Let's remove it from the docs for v8.

Done!

kazupon (approved):

PR #3079, opened 2024-08-29
refactor: reduce console noise about `iso` deprecation

reslear:
CleanShot 2024-08-30 at 02 35 15@2x

also maybe need add:

https://github.com/nuxt-modules/i18n/blob/ebbc7f4779ec6eb3aa2ffa556e797688ea0b846f/src/utils.ts#L17-L19


kazupon (approved):
Welcome contribution!
Thanks!

kazupon:
Ah, Can you fix the CI errors! 🙏

BobbieGoede:
Thank you for your contribution! I've fixed the CI errors so we can include this in the next release 💪

Issue #3002, opened 2024-06-28
Roadmap for v9


kazupon:
### Discussed in https://github.com/nuxt-modules/i18n/discussions/2814

<div type='discussions-op-text'>

<sup>Originally posted by **BobbieGoede** February 20, 2024</sup>
> [!NOTE]  
> This list is incomplete and a work in progress, feel free to suggest features, topics or issues that you think should be included in v9.

There's still plenty of issues open for v8 and we'll continue to work on fixing those, this roadmap is simply to plan and discuss changes that would be breaking. There is currently no estimate or deadline for the next major version, you don't have to worry about migrating (again) anytime soon.

### Features
- [x] support vue-i18n v10
- [x] support nuxt 4 

### Bug Fixes
- [x] support `v-t` fully SSR 

### Breaking changes
- [x] https://github.com/nuxt-modules/i18n/pull/2801
- [x] Rename locales property `iso` to `language`
  - [x] Add alias as described in https://github.com/nuxt-modules/i18n/issues/2449#issuecomment-1740517847 to allow for easier transition
- [x] https://github.com/nuxt-modules/i18n/issues/2712
- [x] Restructure `runtimeConfig` keys
  - [x] Rename `locales` to `domainLocales` and `configLocales` to `locales` (https://github.com/nuxt-modules/i18n/pull/2828#discussion_r1513740467)
- [x] enable jit compilation as default (remove `jit` option) https://i18n.nuxtjs.org/docs/options/compilation#jit
- [x] Fix prefixed injected functions (https://github.com/nuxt-modules/i18n/issues/2168), this is a breaking change as currently the unprefixed inject functions do work but should not and prefixed functions don't.
- [x] Changing/aligning `useLocaleHead` and `useSetI18nParams` argument defaults

### Postponed
- [ ] Change route localization to omit parent from path in `defineI18nRoute` as described in https://github.com/nuxt-modules/i18n/pull/2743#issue-2108024664
  - This can be a non-breaking change if we implement this using relative paths as described in https://github.com/nuxt/nuxt/discussions/28826#localizing-routes
- [ ] https://github.com/nuxt-modules/i18n/issues/1985
  - This is a non trivial change, options used at build time should be taken into consideration, as well as layer support and more. Feel free to open issues, PRs or open a discussion to brainstorm solutions.

### Considering
* Change `locales` configuration
  * Remove `file`, only accepting `files` array instead, simplifies options merging.

</div>

[...]

PR #3097, opened 2024-09-13
feat!: configurable `lang` attribute for `useLocaleHead`

### LINKED ISSUE #2712: Add option to add lang attribute to useLocaleHead composable
### Describe the feature

`useLocaleHead` composable should return the current language which can be added to the `<html lang='…'>`. However the value is present only if `useLocaleHead` is called with `addSeoAttributes: true`, see [/src/runtime/routing/compatibles/head.ts#L52](https://github.com/nuxt-modules/i18n/blob/cd05cea877136abf71b9ab5636b34e536970c7fc/src/runtime/routing/compatibles/head.ts#L52).

Yes, the `lang` attribute is probably related to SEO, but in my opinion that relation is not as strong as its relation to the locale and basic functionality of `useLocaleHead`. You can also notice that the `useLocaleHead` composable is returning an empty state if it is called as `useLocaleHead({})`, which might look more like a bug than words “language is related to the SEO”. Maybe the function may return lang attribute by default, but probably not right now as it might be considered as a breaking change.

I am thinking about new `addLangAttribute` which sets `metaObject.htmlAttrs.lang` regardless of `addSeoAttributes` value. The current behavior should be probably untouched until some major version changes (which might be quite long because the current version is 8.0.0, but it may not be a serious problem).

### Additional information

- [X] Would you be willing to help implement this feature?

### Final checks

BobbieGoede:
### 🔗 Linked issue
* https://github.com/nuxt-modules/i18n/issues/3002
* https://github.com/nuxt-modules/i18n/issues/2712

### ❓ Type of change

### 📚 Description
Resolves https://github.com/nuxt-modules/i18n/issues/2712

This adds the `addLangAttribute` option to the `useLocaleHead` and `$localeHead` functions, after this is merged I will explore changing the default parameters for the functions as suggested in https://github.com/nuxt-modules/i18n/issues/2712 so that the functions do not return an empty object by default.

### 📝 Checklist

### DIFF
--- a/docs/content/docs/5.v9/2.guide/19.breaking-changes-in-v9.md
+++ b/docs/content/docs/5.v9/2.guide/19.breaking-changes-in-v9.md
@@ -60,3 +60,7 @@ Some properties have changed or swapped names to better fit their functionality,
 | --- | --- | --- |
 | `locales` | `domainLocales` | This also changes the environment variable key to `NUXT_PUBLIC_I18N_DOMAIN_LOCALES_{code}_DOMAIN`, see [`runtimeConfig`](/docs/options/runtime-config#domainLocales) |
 | `configLocales` | `locales` | | 
+
+## SEO - `useLocaleHead`
+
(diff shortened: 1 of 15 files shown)


kazupon (approved):
LGTM!

PR #3112, opened 2024-09-19
fix: use generated tsconfig and virtual file augmentations

BobbieGoede:
### ❓ Type of change

### 📚 Description
This cleans up the virtual file augmentations as well as the tsconfig file to resemble the module started more.

I had to add a bunch of type assertions for both `I18n` and runtime config types, will need to look into these,

Thanks to @userquin for giving examples on how to deal with the virtual file types!

### 📝 Checklist


src/module.ts
```diff
@@ -180,8 +185,9 @@ export default defineNuxtModule<NuxtI18nOptions>({
     addPlugin(resolve(runtimeDir, 'plugins/i18n'))
     addPlugin(resolve(runtimeDir, 'plugins/switch-locale-path-ssr'))
 
+    const resolver = createResolver(import.meta.url)
     // for composables
-    nuxt.options.alias['#i18n'] = resolve(distDir, 'runtime/composables/index.js')
+    nuxt.options.alias['#i18n'] = resolver.resolve('./runtime/composables/index')
```
userquin, line 190: add `.js` extension here

package.json
```diff
@@ -47,17 +50,18 @@
...
   "files": [
-    "dist"
+    "dist",
+    "internals.d.ts"
   ],
   "scripts": {
     "prepare": "git config --local core.hooksPath .githooks",
-    "build": "nuxi build-module .",
-    "stub": "nuxi build-module --stub .",
+    "build": "nuxt-module-build build .",
```
userquin: This script should be `nuxt-module-build prepare && nuxt-module-build build .`, you need to create `.nuxt` stuff before running other scripts

kazupon (commented):
Great work!
This approach makes sense!

I have always found it strange that the `src` directory contains a mixture of `.ts` implementation files and `.d.ts` type definition files, and I have wanted to refactor them at some point.

Virtual modules solve this issue, but it depends on the bundler. Since the nuxt module is a framework module system and Nuxt provides a bundling framework system, I believe the bundler dependency is not an issue.

Let's keep it up!

@userquin 
Thank you for supporting!

src/module.ts
BobbieGoede: Looks like adding the extension breaks explicit imports (when having auto imports disabled) https://github.com/nuxt-modules/i18n/actions/runs/10947564325/job/30397002095?pr=3112

BobbieGoede:
Seems to be working as expected now 🤔 It's unfortunate that we now have two identical virtual file augmentations, and will have to look into the added type assertions in a future refactor. But overall this should make the module types more predictable and in line with established (but undocumented 😅) conventions among Nuxt modules.

PR #3117, opened 2024-09-20
chore: refactor module

userquin:
### 🔗 Linked issue

This PR just moves logic from `module.ts` to the corresponding module: for example, when configuring nitro, the current `module.ts` will build the `AdditionalSetupNitroParams` (being used or not), with the new approach, the `nitro.ts` module will build the `AdditionalSetupNitroParams` when required.

The idea is to have the `module.ts` to orchestrate the I18n configuration: any shared data should be on the context, and the modules should do their work. The `module.ts` will call the initialization sequence correctly, and modules preparing shared data should add their stuff to the context (check `prepare/locale-info.ts` and `prepare/runtime.ts`).

Any module should only accept the context and nuxt as parameters: using destructuring on the context to get used data.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/prepare/runtime.ts
+++ b/src/prepare/runtime.ts
@@ -1,11 +1,11 @@
 import type { Nuxt } from '@nuxt/schema'
-import type { I18nNuxtContext } from './context'
-import { i18nVirtualLoggerPlugin, RESOLVED_VIRTUAL_NUXT_I18N_LOGGER, VIRTUAL_NUXT_I18N_LOGGER } from './virtual-logger'
+import type { I18nNuxtContext } from '../context'
+import { i18nVirtualLoggerPlugin, RESOLVED_VIRTUAL_NUXT_I18N_LOGGER, VIRTUAL_NUXT_I18N_LOGGER } from '../virtual-logger'
 import { defu } from 'defu'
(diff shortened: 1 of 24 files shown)


BobbieGoede:
approving to run ci

src/prepare/runtime.ts
```diff
@@ -0,0 +1,71 @@
...
+      normalizedLocales,
+      dev,
+      isSSG,
+      parallelPlugin: options.parallelPlugin
+    })
+  }
+
+  ctx.genTemplate = genTemplate
+
+  nuxt.options.runtimeConfig.public.i18n.locales = simplifyLocaleOptions(nuxt, defu({}, options))
```
BobbieGoede, line 44: I'm guessing we can't move this to `prepare/runtime-config.ts` right? 🤔

userquin: oh, we can do what we want here

BobbieGoede (approved):

src/prepare/runtime.ts
userquin: we can move runtime config after runtime and then we can move it: do you want me to move it?

kazupon (approved):
This PR is amazing!
Thanks!

Issue #2976, opened 2024-06-13
useRoute('login') does not work, it expects useRoute('login__LOCALE')


martinszeltins:
### Environment

N/A

### Reproduction

N/A

### Describe the bug

Nuxt i18n module has support for Nuxt typedPages but I can't understand how to use it.

Nuxt docs says:

> You can get typed params within a page by using `const route = useRoute('route-name')`

But I don't have one `route-name` when using i18n. I don't have a `login` route. Instead I have `login__en`, `login__de`, `login__fr` etc.

Passing in simply _"login"_ does not work `useRoute('login')`, it expects a locale which is very strange. How do I use this? I couldn't find anything in the docs.

martijndewit:
```
definePageMeta({
      name: 'login'
    })
```
on the login page

So you can use 'name' as `localeRoute({ name: 'login' }, locale.value)` for example.

martinszeltins:
@martijndewit That didn't work. I still see `index__en` and `user__de` in route names.

![image](https://github.com/nuxt-modules/i18n/assets/34019878/490d56d7-acdc-4b92-a297-28b8aa36f765)

![image](https://github.com/nuxt-modules/i18n/assets/34019878/35d3d3f7-cf58-470d-8c11-7bfb78a332f5)

martijndewit:
How did you set your [Routing Strategy](https://i18n.nuxtjs.org/docs/guide) ?

martinszeltins:
Sorry, that still doesn't work. `useRoute()` expects a union type to be passed in. I don't think the routing strategy has anything to do with it.

Error:
````
Argument of type '(RouteLocation & { href: string; }) | undefined' is not assignable to parameter of type 'keyof RouteNamedMap | undefined'.
````

![image](https://github.com/nuxt-modules/i18n/assets/34019878/cf607c9c-2057-4dad-a382-ae5188544d88)

MinecraftStorage1:
Same issue, why route names are localized by default? It makes code extremely unreadable and quite doesn't make sense. Every time I need to resolve a route path, I need to figure out the current locale and then suffix it with that locale. I think you should be able to just reference the route name without a locale, and it resolves into a path with the current locale.

letoast:
How do you reproduce this with stackblitz? As far as I've tried there are no type hints on stackblitz with nuxt(vue). 
Not only that, I can't even get i18n to redirect to /en on stackblitz

[...]

Issue #2813, opened 2024-02-20
Support typed routes inside localePath()


AndreyYolkin:
### Describe the feature

According to the code here, nuxt i18n imports types from `vue-router` package:
https://github.com/nuxt-modules/i18n/blob/c43842f11027253c07c45b7a9e8e6ab81bae7d54/src/runtime/composables/index.ts#L24

However, nuxt itself registers `#vue-router` alias to handle both situations with enabled/disabled typed routes. Example of usage: https://github.com/nuxt/nuxt/blob/d326e054d372bd2eb5bf75f3feca6a291169ff76/packages/nuxt/src/pages/runtime/utils.ts#L2

Since nuxt/i18n is a nuxt module, I suppose we can rely on this alias too and enable type imports by replacing `vue-router` with `#vue-router` for types imports

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution) (The contribution guideline of nuxt-modules/i18n is compliant with Nuxt too).
- [X] Check existing [discussions](https://github.com/nuxt-modules/i18n/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

AndreyYolkin:
As workaround, I added this snippet inside d.ts file and it works:

```ts
/// <reference types="unplugin-vue-router/client" />
import type { RouteLocation, RouteLocationNormalizedLoaded, RouteLocationRaw, Router } from '#vue-router';

declare module '#i18n' {
  export type StubbedLocalePathFunction = (route: RouteLocation | RouteLocationRaw, locale?: Locale) => string;

  export declare function useLocalePath(): StubbedLocalePathFunction;
}
```

Anoesj:
Thanks @AndreyYolkin! I modified it a bit, because lately `vue-router` started providing typed routes by itself. Your stub wasn't working as is for me, because `nuxt-i18n` suffixes route names with `___<locale>`. Hopefully this is useful to anyone needing this.

`utils/locales.ts`:
```ts
export const locales = [
  'nl',
  'en',
  'fr',
  'de',
] as const;

export type AppLocale = typeof locales[number];
```

`/typings/i18n.d.ts`:
```ts
import type { RouteNamedMap } from 'vue-router/auto-routes';
import type { RouteLocation, RouteLocationRaw } from 'vue-router';

type RouteName = keyof RouteNamedMap;
[...]

PR #3142, opened 2024-09-27
feat: experimental typed routes

BobbieGoede:
### 🔗 Linked issue
* #2813
* #2976

### ❓ Type of change

I'm not very happy with the implementation right now, but it does work... I think the demand/convenience of this feature somewhat justifies adding this albeit under an experimental flag.

We should communicate very clearly that this relies on internals of Nuxt, Vue Router and `unplugin-vue-router`, perhaps we can make the types fallback to original behavior on failure (maybe it already does). A proper implementation will likely require collaboration with the maintainers of those packages, we should not expect this to keep working as is.

This will need some documentation and basic tests so we know things fail if internals have changed. 

Demo stackblitz project here: https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-srohiz?file=nuxt.config.ts&file=composables%2Ftest.ts

### Demo of `localePath` autocompletion
https://github.com/user-attachments/assets/29d00cb6-1cdc-409f-bb23-fa08153d4a4e

### Demo of custom routes configuration autocompletion and basic validation (prefix `'/'`)
https://github.com/user-attachments/assets/38c41d9d-deb7-4688-be02-1ce0c18a9628

### 📝 Checklist


src/gen.ts
kazupon, line 1: wow! This typing is too amazing!

kazupon (commented):
I've just reviewed your PR!
It's an excellent for me!

kazupon:
Since it uses the vue-router type, it would be good if posva is able to review it.

@posva
Can you review it ?, if you have the time 🙏

src/gen.ts
BobbieGoede: I should reference the original source types 😅 I had to redefine/reimplement a bunch of internal types of `vue-router` and uvr, so the credits for these go to @posva

BobbieGoede:
@userquin 
If you see some obvious structural improvements, feel free to suggest them, the must be some way to leave out the augmentations in `internal-global-types.d.ts` 🤔

src/internal-global-types.d.ts
```diff
@@ -14,6 +14,78 @@ declare module '#app' {
...
+      ? RouteLocationAsRelativeGeneric
+      : RouteLocationAsRelativeTypedList<RouteMapI18n>[Name]
+
+  export type RouteLocationAsPathI18n<Name extends keyof RouteMapI18n = keyof RouteMapI18n> =
+    RouteMapGeneric extends RouteMapI18n ? RouteLocationAsPathGeneric : RouteLocationAsPathTypedList<RouteMapI18n>[Name]
+
+  /**
+   * Helper to generate a type safe version of the {@link RouteLocationAsRelative} type.
+   */
+  export interface RouteLocationAsRelativeTyped<
```
posva: Isn't this the same type exported by the router?

BobbieGoede: Hmm most of the types should be the same except suffixed with `I18n`, which this type seems to be missing. This might conflict with existing router types 🫣 going to look into it.

Reasoning behind redefining the same types but suffixed is to use the `RouteMapI18n` type, perhaps there's a better way to go about this that I'm unaware of.

posva: Doesn’t extending the same route map from Vue router works? Or do you need to have two separate route maps?

BobbieGoede: Since we're generating routes for each locale (depending on the routing strategy) and custom localized route the generated types are localized, this PR generates `RouteMapI18n` (which ironically contains the routes without localization) to be used for the i18n routing utility functions.

Some examples to illustrate the problem we're trying to solve:

Named route completion
image


Path completion (not desirable when using custom routes)
image

I guess this is specific to i18n, I doubt there are other use cases where generating a separate map is necessary 🤔

posva: I would need to try it out to see the needed final types. I was hoping the types were flexible enough to at least keep things typer for the router methods. I18n methods could use extra types for localized route names (that seems to be the main difference)

docs/content/docs/5.v9/3.options/10.misc.md
```diff
@@ -18,6 +18,9 @@ About how to define the locale detector, see the [`defineI18nLocaleDetector` API
 - `autoImportTranslationFunctions` (default: `false`) - Automatically imports/initializes `$t`, `$rt`, `$d`, `$n`, `$tm` and `$te` functions in `<script setup>` when used.
 ::callout{icon="i-heroicons-exclamation-triangle" color="amber"}
 This feature relies on [Nuxt's Auto-imports](https://nuxt.com/docs/guide/concepts/auto-imports) and will not work if this has been disabled.
+- `typedPages` (default: `false`) - Generates route types used in composables and configuration.
+::callout{icon="i-heroicons-exclamation-triangle" color="amber"}
+This feature relies on [Nuxt's `experimental.typedRoutes`](https://nuxt.com/docs/guide/going-further/experimental-features#typedpages) and will not work if this is not enabled.
```
darthf1, line 23: Maybe its possible to enable the `experimental.typedRoutes` for users, when they set `typedPages` to `true`? :)

BobbieGoede: Sounds good to me, since it's an experimental feature in Nuxt too we might as well automatically enable this when Nuxt's `experimental.typedPages` is enabled and when i18n's `experimental.typedPages` is left unset. Explicitly setting the property to false should still disable it.

Issue #2591, opened 2023-12-04
Enable type checking from vue-i18n


dreitzner:
### Describe the feature

vue-i18n supports type safe resource Schemas. ([docs](https://vue-i18n.intlify.dev/guide/advanced/typescript.html))

It would be nice to have a way to do the same in nuxt-i18n as well.

From what I can gleen from the source code it might only be needed to add a possible type type to `defineNuxtPlugin` that is passed to `createI18n`.

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution) (The contribution guideline of nuxt-modules/i18n is compliant with Nuxt too).
- [X] Check existing [discussions](https://github.com/nuxt-modules/i18n/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

danwithabox:
For people who want type-safety out-of-the-box without waiting for this, I have successfully used [global resource schema type definitions](https://vue-i18n.intlify.dev/guide/advanced/typescript#global-resource-schema-type-definition) as a workaround.

adamdehaven:
I'm willing to contribute, but might need an assist: https://github.com/nuxt-modules/i18n/issues/1753#issuecomment-2146330404

thomaslecoeur:
> For people who want type-safety out-of-the-box without waiting for this, I have successfully used [global resource schema type definitions](https://vue-i18n.intlify.dev/guide/advanced/typescript#global-resource-schema-type-definition) as a workaround.

Any details on the implementation? Can we reproduce on a vanilla nuxt project?

BobbieGoede:
The latest v9 release candidate (https://github.com/nuxt-modules/i18n/releases/tag/v9.0.0-rc.2) includes experimental `vue-i18n` options and messages generation, this provides autocompletion for configuring options and translation functions, please try it out and let me know if this works as expected 🙏

More on the option that enables this functionality: https://i18n.nuxtjs.org/docs/v9/options/misc#experimentaltypedoptionsandmessages, the 
PR implementing this has a demo video of it in action https://github.com/nuxt-modules/i18n/pull/3151

BobbieGoede:
Closing this as we have experimental support for this in the stable release of v9, please try it out and let me know your thoughts!

Issue #1753, opened 2022-12-20
$t() autocomplete support


EternalC0der:
### Describe the feature

Hello,
It would be really nice to have auto completion feature while using `t('welcome')` in template, especially when you want to reuse keys in different sections.

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution) (The contribution guideline of nuxt-modules/i18n is compliant with Nuxt too).
- [X] Check existing [discussions](https://github.com/nuxt-modules/i18n/discussions) and [issues](https://github.com/nuxt/framework/issues).

kazupon:
Hi!
I'll support about auto complete.
vue-i18n v9 has type-safe resources feature.
https://vue-i18n.intlify.dev/guide/advanced/typescript.html

we need to setup manually in vue-i18n. 
In nuxt, I will support auto complete by respecting the `defaultLocale` and dynamically generating types with that resource as master schema.

I would also like to support `$t`, but we need to refactor the `$t` type definition on the vue-i18n side.
related issue on vue-i18n:
https://github.com/intlify/vue-i18n-next/issues/1124

jenseo:
@kazupon Any news on when we can expect this to be shipped? It would be a real step forward for developer experience in my opinion :)

natemate90:
Any update on this?

EternalC0der:
I have been waiting for this feature for so long 😂
Any updates?

Here is a quick solution to get the list of locale keys, I wrote this to manually add type-safety for my code.
I hope it helps others:

```ts
import english from '~/i18n/en-US.json'

type StringKey<T> = Extract<keyof T, string>
type GenerateKeyPaths<T, Prefix extends string = ''> = T extends object
  ? { [K in StringKey<T>]: T[K] extends object ? GenerateKeyPaths<T[K], `${Prefix}${K}.`> : `${Prefix}${K}` }[StringKey<T>]
  : Prefix

type Locale = typeof english
export type I18nKeys = GenerateKeyPaths<Locale>
```

Bobakanoosh:
@EternalC0der how are you using that? Are you making a custom translation function that then uses `I18nKeys`?

This works if we use `t` from `useI18n`
```ts
import "vue-i18n";
import en from "./i18n/website/en.json";

type MainTranslations = typeof en;
declare module "vue-i18n" {
[...]

PR #3151, opened 2024-10-03
feat: experimental `vue-i18n` and messages type generation

BobbieGoede:
### 🔗 Linked issue
* #1753
* #2591 

### ❓ Type of change

### 📚 Description
This is a different, improved approach compared to https://github.com/BobbieGoede/i18n/pull/50, that PR reimplemented the loading of configs and messages while this PR reuses the existing logic.

(Same demo video as https://github.com/BobbieGoede/i18n/pull/50 as this should be functionally the same)

https://github.com/user-attachments/assets/f135dd5f-3a43-4420-9e5d-246deea7f8bc

### 📝 Checklist


src/module.ts
```diff
@@ -148,6 +154,12 @@ declare module '@nuxt/schema' {
   }
   interface NuxtOptions {
     ['i18n']: UserNuxtI18nOptions
+    /**
+     * @internal
+     */
+    _i18n: {
```
kazupon: sorry, where is this option used? I couldn't find it. 😅

src/runtime/server/api/merged-options.get.ts
kazupon, line 1: I will leave a comment on the points that I am concerned about.

I integrated all the locale messages when I implemented i18n support first time for h3.
I noticed that all the locale messages are unnecessary on the server side, affecting performance.
Since this is experimental, I think this implementation is fine, but at some point, I think a mechanism that allows locale messages to be loaded only on the server side will be necessary.

kazupon:
@BobbieGoede 
Excellent work! 💯 

I've wanted to implement a type generation like your PR, and I'm so happy you're supporting it!
Let's keep going!

src/module.ts
BobbieGoede: Ah looks like I forgot to remove that, it's part of the original/previous implementation in https://github.com/BobbieGoede/i18n/pull/50, will clean it up

src/runtime/server/api/merged-options.get.ts
BobbieGoede: Hmm I'm not entirely sure what you mean 🤔 Do you mean a mechanism to load a part/subset of messages?

I haven't checked if this does anything to a project production build, this should only be used in development so I should probably disable this functionality if not in dev mode.

BobbieGoede: After looking into it, it looked like it was being included in production too 😅 Now this functionality should only be enabled (if configured) during development, this still has performance impact but not after building.

BobbieGoede: I have added some options, the configuration now accepts `false | 'default' | 'all'`:

- `false` - disables type generation
- `'default'` - generate types based on configured `defaultLocale`
- `'all'` - generate types based on all configured locales

This way users can decide to only generate types using the `defaultLocale` resources, which prevents loading all locales unnecessarily.

PR #3243, opened 2024-11-24
[DRAFT] feat: merge static locale file chains

BobbieGoede:
### ❓ Type of change

### 📚 Description

Proof of concept, needs to be configurable and documentation.

This merges static locale files to improve performance and memory usage, currently merging happens at runtime which is wasteful as the merged result is always the same for subsequent static files.

### 📝 Checklist

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -21,7 +21,7 @@ export async function extendBundler({ options: nuxtOptions }: I18nNuxtContext, n
   const i18nModulePaths =
     nuxtOptions?.i18nModules?.map(module => resolve(nuxt.options._layers[0].config.rootDir, module.langDir ?? '')) ?? []
   debug('i18nModulePaths -', i18nModulePaths)
-  const localePaths = [...langPaths, ...i18nModulePaths]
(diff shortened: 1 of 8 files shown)


src/bundler.ts
```diff
@@ -42,7 +42,7 @@ export async function extendBundler({ options: nuxtOptions }: I18nNuxtContext, n
       compositionOnly: nuxtOptions.bundle.compositionOnly,
       onlyLocales: nuxtOptions.bundle.onlyLocales,
       dropMessageCompiler: nuxtOptions.bundle.dropMessageCompiler,
-      optimizeTranslationDirective: true,
+      optimizeTranslationDirective: false,
```
userquin: shouldn't we add a new module option (disabled by default)? Maybe we should add this in a new PR...

BobbieGoede: Oops I didn't mean to push that, that's a remnant of testing other things. 

Ideally we can fix the issue that's forcing projects to have this disabled, if I recall correctly we need this enabled for `v-t` to work in SSR (https://github.com/nuxt-modules/i18n/pull/3014), so disabling it is not desirable.

BobbieGoede:
(still draft, just triggering ci)

BobbieGoede:
Superseded by #3584 and #3579

PR #3272, opened 2024-12-20
refactor: provide context functions from plugin

BobbieGoede:
### ❓ Type of change

### 📚 Description

This changes the way functions are added to the context, relying on Nuxt plugin returning these via the `provide` property. Ideally we can remove the type assertions, these were not required previously as we indirectly added these to the `defineNuxtPlugin` statement previously to achieve the same types.

### 📝 Checklist


kazupon (approved):
great refactoring!
thanks!

PR #3273, opened 2024-12-20
refactor: simplify bundler plugin registration

BobbieGoede:
### ❓ Type of change

### 📚 Description
Reduces duplicate options and uses `addBuildPlugin` where possible, inspired from Nuxt's source code.

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,8 @@
       "@nuxtjs/i18n": "link:.",
       "consola": "^3",
       "unimport": "3.13.4",
-      "@intlify/core": "^10.0.5"
+      "@intlify/core": "^10.0.5",
+      "rollup": "4.28.1"
     }
   },
   "dependencies": {
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -1,8 +1,7 @@
 import createDebug from 'debug'
 import { resolve } from 'pathe'
-import { extendViteConfig, addWebpackPlugin, addVitePlugin, addBuildPlugin } from '@nuxt/kit'
-import VueI18nWebpackPlugin from '@intlify/unplugin-vue-i18n/webpack'
-import VueI18nVitePlugin from '@intlify/unplugin-vue-i18n/vite'
+import { extendViteConfig, addWebpackPlugin, addBuildPlugin } from '@nuxt/kit'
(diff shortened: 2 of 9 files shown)


src/bundler.ts
```diff
@@ -67,35 +61,29 @@ export async function extendBundler({ options: nuxtOptions }: I18nNuxtContext, n
...
-  addVitePlugin(TransformMacroPlugin.vite(sourceMapOptions))
-  addVitePlugin(ResourcePlugin.vite(sourceMapOptions))
-  if (nuxtOptions.experimental.autoImportTranslationFunctions) {
-    addVitePlugin(TransformI18nFunctionPlugin.vite(sourceMapOptions))
-  }
-
+  addVitePlugin(
+    VueI18nVitePlugin({
+      ...vueI18nPluginOptions,
+      fullInstall: nuxtOptions.bundle.fullInstall,
```
BobbieGoede: @kazupon 
These options were not set for the webpack plugin, was this intentional? 🤔

kazupon: @BobbieGoede 
Ah, I had forgotten these options for webpack. 😅 
you can specify these options for webpack.

package.json
```diff
@@ -81,7 +81,8 @@
     "overrides": {
       "@nuxtjs/i18n": "link:.",
       "consola": "^3",
-      "unimport": "3.13.4"
+      "unimport": "3.13.4",
+      "rollup": "4.28.1"
```
kazupon, line 86: why do we need to override? 🤔

BobbieGoede: This is to resolve the type mismatch for the nitro plugins like [`acfc08f` (#3273)](https://github.com/nuxt-modules/i18n/pull/3273/commits/acfc08f1296b5f55cc8424786160f1131c6f7f9f#diff-126f9d42f327404ca57cb5a71d7c101fe1b0ee34dc04757b61cf37d09b21f257L64), I was wondering how Nuxt isn't running into the same issue and found they also override the resolution for rollup https://github.com/nuxt/nuxt/blob/main/package.json#L59

src/bundler.ts
BobbieGoede: I see! changed it mow so all options are the same for vite and webpack 💪

Issue #2698, opened 2024-01-13
Inconsistent path encoding


BobbieGoede:
### Environment

```
- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.8.0
- CLI Version:      3.9.1
- Nitro Version:    2.7.2
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      modules, i18n
- Runtime Modules:  @nuxtjs/i18n@8.0.0-28410629.a2a883f
- Build Modules:    -
```

### Reproduction

https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-vuphy7?file=pages%2Findex.vue

Scripts to compare behavior:
* `npm run dev`
  * strategy = no_prefix
* `npm run dev:alternate-1` 
  * strategy = prefix_and_default
* `npm run dev:alternate-2`
  * strategy = prefix_and_default
  * defaultLocale = en

### Describe the bug

The paths returned by `localePath` are escaped inconsistently as shown in the scenarios below. It seems to be related to when locale prefixes are used, hence it changes when `defaultLocale` is set or not when used with strategy `prefix_and_default`. Found during https://github.com/nuxt-modules/i18n/pull/2686.

This issue is likely related or the same as previously reported issues related to path encoding .

```ts
strategy: "no_prefix"
defaultLocale: "undefined"
localePath: "/path/hello%20world?foo=hello+world"
```

```ts
strategy: "prefix_and_default"
defaultLocale: "undefined"
localePath: "/en/path/hello world?foo=hello+world"
```

```ts
strategy: "prefix_and_default"
defaultLocale: "en"
localePath: "/path/hello%20world?foo=hello+world"
```

mcombes-ew:
I am having a similar issue with Nuxt 3.

As an example: if I navigate to a page like the link below everything works as intended. 

             mysite.net/somePath/some%20%26%20path

But if there is any query string at the end it decodes the %26 to an & like example below.

             mysite.net/somePath/some%20%26%20path?queryString=true

gets decoded to 
             mysite.net/somePath/some%20&%20path?queryString=true
             
Can not figure out why this is happening.

Red-Asuka:
I have the same issue, if the URL contains a query, and the value of the query is encoded, after using localePath, the value of the query will be encoded multiple times.

```js
localePath({ path: '/signin', query: { continue: `https://www.example.com/` } }) 
// /signin?continue=https://www.example.com/

[...]

PR #3276, opened 2024-12-22
refactor: simplifications and cleanup

BobbieGoede:
### ❓ Type of change

### 📚 Description
This touches a lot of code, maybe I should split it up for clarity?

Mostly refactors runtime code, the goal is to reduce code surface for maintainability while also reducing the build size.

* Removes injections type, use composables to provide context functions where possible
* Remove duplicate function `adjustRoutePathForTrailingSlash`, we did test its behavior but we are not using it so I removed the tests.. May look into keeping the tests, but the functions are not similar enough to use as-is. 
* Extracts shared properties from `...CustomProperties` interfaces for reuse

I have simplified most of the composable JSDoc and removed most of the JSDoc from their internal functions, these were originally exported in `vue-i18n-routing` which is why they were well documented, but we are documenting the same functions multiple times due to the wrapping of composables. The simplification and removals should make it easier to maintain and keep the user-facing descriptions accurate.

### 📝 Checklist


BobbieGoede:
I think I accidentally fixed #2698 with these refactors... 🤔 reproduction using the preview release https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-dkqwmcqu?file=package.json

BobbieGoede:
I'm going to split this PR up in smaller pieces and merge the trivial changes, I don't think it's realistic to check all the changes combined like this 😅

PR #3280, opened 2024-12-24
chore: remove unused duplicate `adjustRoutePathForTrailingSlash`

BobbieGoede:
### ❓ Type of change

### 📚 Description
Remove duplicate function `adjustRoutePathForTrailingSlash`, this was copied from `vue-i18n-routing` but was not actually being used. We have a function with the same but its implementation has changed which means the tests cannot be reused as-is.

while we did test its behavior but we are not using it so I removed the tests.. May look into keeping the tests, but the functions are not similar enough to use as-is.

### 📝 Checklist


kazupon (approved):
LGTM!

PR #3379, opened 2025-02-26
fix: Unhead v2 compatibility

harlan-zw:
### 🔗 Linked issue

https://github.com/nuxt/ecosystem-ci/actions/runs/13519021085/job/37843739041#step:7:1498

### ❓ Type of change

### 📚 Description

In Unhead v2 `getActiveHead` is no longer exported. Instead, end users should use `injectHead()`, however in this implementation we should just use `useHead()` to avoid adding extra entries for each modification (will use less memory).

### 📝 Checklist


BobbieGoede (approved):
Thanks! 🙏

Issue #3407, opened 2025-03-12
Issue with SSG and detectBrowserLanguage: locale messages are not being loaded


ap-arto:
### Environment

- Operating System: Darwin
- Node Version:     v22.13.0
- Nuxt Version:     3.16.0
- CLI Version:      3.22.5
- Nitro Version:    2.11.6
- Package Manager:  npm@10.9.2
- Builder:          -
- User Config:      compatibilityDate, devtools, modules, i18n, runtimeConfig
- Runtime Modules:  @nuxtjs/i18n@9.3.1
- Build Modules:    -

### Reproduction

https://github.com/ap-arto/nuxt-i18n-ssg-issue

`npm i`
`npm run generate`
`PORT=3001 npm run preview`

### Describe the bug

When we use `detectBrowserLanguage: false` and `differentDomains: true`, the translations will not load properly in SSG mode.
The `main` branch contains version 9.3.1, while the `older-version` branch contains version 9.1.1, which works as expected

### Additional context

This issue has been present since 9.1.2

### Logs

```shell
i18n:plugin:i18n isSSG true DTTkGQjM.js:36:75
i18n:plugin:i18n useCookie on setup false DTTkGQjM.js:36:75
i18n:plugin:i18n defaultLocale on setup es DTTkGQjM.js:36:75
i18n:extendBaseUrl using differentDomains - 
Object { domain: "http://localhost:3001" }
DTTkGQjM.js:36:75
i18n:getLocaleCookie:client 
Object { useCookie: false, cookieKey: false, localeCodes: (2) […] }
DTTkGQjM.js:36:124
i18n:detectLocale 
Object { detectBrowserLanguage: {…} }
DTTkGQjM.js:36:124
i18n:detectLocale 1/3 
Object { detected: "", strategy: "no_prefix" }
DTTkGQjM.js:36:75
i18n:getLocaleDomain locating domain for host 
Object { host: "localhost:3001", strategy: "no_prefix", path: "/" }
DTTkGQjM.js:36:75
i18n:getLocaleDomain found one matching domain 
Object { host: "localhost:3001", matchedLocale: "es" }
DTTkGQjM.js:36:75
i18n:detectLocale 2/3 
Object { detected: "es", detectBrowserLanguage: false }
DTTkGQjM.js:36:75
i18n:detectLocale 3/3 
Object { detected: "es", cookieLocale: undefined, defaultLocale: "es", localeCookie: undefined }
DTTkGQjM.js:36:75
[...]

PR #3410, opened 2025-03-13
fix: respect `detectBrowserLanguage: false` in SSG plugin

ap-arto:
Do not call `detectBrowserLanguage` when `options.detectBrowserLanguage` is set to `false`

### 🔗 Linked issue
https://github.com/nuxt-modules/i18n/issues/3407

### ❓ Type of change

### 📚 Description

`i18n:plugin:ssg-detect` doesn't respect `nuxtI18nOptions.detectBrowserLanguage` and calls `detectBrowserLanguage()` regardless if that option is set to `false`.
And it causes an issue: when we use `strategy: 'no_prefix'` and `differentDomains: true` alongside with `detectBrowserLanguage: false`, locale messages are not being loaded.

### 📝 Checklist

### DIFF
--- a/specs/fixtures/issues/3407/app.vue
+++ b/specs/fixtures/issues/3407/app.vue
@@ -1,0 +1,6 @@
+<template>
+  <div>
+    <NuxtRouteAnnouncer />
+    <NuxtPage />
+  </div>
+</template>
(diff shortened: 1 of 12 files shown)


BobbieGoede (approved):

BobbieGoede:
Thank you for contributing! 🙏

PR #3416, opened 2025-03-17
feat: add handling if config file fails to load

gbyesiltas:
### 🔗 Linked issue

I don't think there is a specific issue related to this but we have been seeing a number of sentry errors where we see this line failing every once in a while, causing some other errors to appear down the line as well.

### ❓ Type of change

### 📚 Description

If the config file fails to load, we now reload the nuxt app. The `reloadNuxtApp` function has built-in reload-loop protection. There is also a `persistState` flag that we could use to maybe improve performance but I didn't do that since I don't know enough about this module to know if having left-over data from one reload to another could cause other issues

### 📝 Checklist

### DIFF
--- a/src/runtime/messages.ts
+++ b/src/runtime/messages.ts
@@ -2,7 +2,7 @@ import { deepCopy, isFunction } from '@intlify/shared'
 import { createLogger } from '#nuxt-i18n/logger'
 
 import type { I18nOptions, Locale, FallbackLocale, LocaleMessages, DefineLocaleMessage } from 'vue-i18n'
-import type { NuxtApp } from '#app'
+import { type NuxtApp, reloadNuxtApp } from '#app'
 import type { DeepRequired } from 'ts-essentials'
 import type { VueI18nConfig, NuxtI18nOptions } from '#internal-i18n-types'
 import type { CoreContext } from '@intlify/h3'
@@ -17,18 +17,22 @@ export type LocaleLoader<T = LocaleMessages<DefineLocaleMessage>> = {
 }
 
 const cacheMessages = new Map<string, LocaleMessages<DefineLocaleMessage>>()
-
 export async function loadVueI18nOptions(
   vueI18nConfigs: VueI18nConfig[],
   nuxt: Pick<NuxtApp, 'runWithContext'>
 ): Promise<I18nOptions> {
   const vueI18nOptions: I18nOptions = { messages: {} }
-  for (const configFile of vueI18nConfigs) {
-    const { default: resolver } = await configFile()
+  try {
+    for (const configFile of vueI18nConfigs) {
+      const { default: resolver } = await configFile()
 
-    const resolved = isFunction(resolver) ? await nuxt.runWithContext(() => resolver()) : resolver
+      const resolved = isFunction(resolver) ? await nuxt.runWithContext(() => resolver()) : resolver
 
-    deepCopy(resolved, vueI18nOptions)
+      deepCopy(resolved, vueI18nOptions)
+    }
+  } catch {
+    // something went wrong, reload the app (`reloadNuxtApp` has built-in reload-loop protection)
+    reloadNuxtApp()
   }
 
   return vueI18nOptions


BobbieGoede:
Thank you for your interest in contributing 🙏

>I don't think there is a specific issue related to this but we have been seeing a number of sentry errors where we see this line failing every once in a while, causing some other errors to appear down the line as well.

It would be helpful to have a reproduction of this, are you using async locale resources that could be throwing during a request or something like this?

I'm hesitant to implement reloading the app as a solution for this, I can imagine it will trigger constant reloading during development for faulty locale files and there may be a root cause to this that we could fix properly.

gbyesiltas:
Hey, thanks for the reply :) @BobbieGoede 

> It would be helpful to have a reproduction of this, are you using async locale resources that could be throwing during a request or something like this?

Sorry for the sparse PR description, we ourselves aren't able to create a reproduction locally but we do see it happen relatively often on our sentry errors. The specific error is `[nuxt] error caught during app initialization TypeError: Right side of assignment cannot be destructured` and according to the source maps, it is coming from here. I'll link our specific configuration in this comment, but I think in any case, it may make sense to include some sort of error-handling here as it could be caused by any sort of network error even if the whole configuration is right. 

> I'm hesitant to implement reloading the app as a solution for this, I can imagine it will trigger constant reloading during development for faulty locale files and there may be a root cause to this that we could fix properly.

Yeah I get it, though he `reloadNuxtApp` method [does have internal reloop protection](https://nuxt.com/docs/api/utils/reload-nuxt-app). I'm happy to adjust it if you have other concerns though :) 

Our specific nuxt-i18n configuration is:
```
import type { LocaleObject } from "@nuxtjs/i18n/dist/module";
import type { NuxtConfig } from "nuxt/schema";

export const availableLocales = [
  {
    code: "en",
    language: "en-GB",
    file: "en_GB.json",
    abbreviation: "en",
    name: "English",
  },

  {
    code: "nl",
    language: "nl-NL",
    file: "nl_NL.json",
    abbreviation: "nl",
    name: "Nederlands",
  },

  {
    code: "fr",
    language: "fr-FR",
    file: "fr_FR.json",
    abbreviation: "fr",
    name: "Français",
  },

  {
    code: "de",
    language: "de-DE",
    file: "de_DE.json",
    abbreviation: "de",
    name: "Deutsch",
  },

  {
    code: "it",
    language: "it-IT",
    file: "it_IT.json",
    abbreviation: "it",
    name: "Italiano",
  },

  {
    code: "id",
    language: "id-ID",
    file: "id_ID.json",
    abbreviation: "id",
    name: "Bahasa Indonesia",
  },

  {
    code: "es",
    language: "es-ES",
    file: "es_ES.json",
    abbreviation: "es",
    name: "Español",
  },

  {
    code: "pt",
    language: "pt-PT",
    file: "pt_PT.json",
    abbreviation: "pt",
    name: "Português",
  },

  {
    code: "br",
    language: "pt-BR",
    file: "pt_BR.json",
    abbreviation: "pt",
    name: "Português do Brasil",
  },

  {
    code: "ja",
    language: "ja-JP",
    file: "ja_JP.json",
    abbreviation: "ja",
    name: "日本語",
  },

  {
    code: "zh",
    language: "zh-CN",
    file: "zh_CN.json",
    abbreviation: "zh",
    name: "简体中文",
  },
  {
    code: "tr",
    language: "tr-TR",
    file: "tr_TR.json",
    abbreviation: "tr",
    name: "Türkçe",
  },
  {
    code: "ko",
    language: "ko-KR",
    file: "ko_KR.json",
    abbreviation: "ko",
    name: "한국어",
  },
] satisfies LocaleObject[];

/**
 * See: https://i18n.nuxtjs.org/getting-started/setup
 */
const nuxtI18nModuleConfig = {
  strategy: "prefix_and_default",
  compilation: {
    strictMessage: false, // so that the html messages in the translations don't break the build
  },
  lazy: true,
  langDir: "transifex/", // adds the `i18n` in the beginning by default
  locales: availableLocales,
  defaultLocale: "en",
  // We handle browser language detection ourselves on the client,
  // as it intervenes with our initial doc caching.
  detectBrowserLanguage: false,
  vueI18n: "./vueI18nConfig.ts",
  experimental: {
    typedOptionsAndMessages: "all",
  },
} satisfies NuxtConfig["i18n"];

export default nuxtI18nModuleConfig;
```

vue-i18n configuration is:
```
import type { I18nOptions } from "#i18n";

import { availableLocales } from "./i18nConfig";

export const vueI18nConfigOptions = {
  legacy: false,
  fallbackLocale: "en",
  warnHtmlInMessage: "off",
  warnHtmlMessage: false,
  availableLocales: availableLocales.map((locale) => locale.code),
} satisfies I18nOptions;

/**
 * vue-i18n configuration. this file is pointed to by the @nuxtjs/i18n module config in `i18nConfig.ts`
 */
export default defineI18nConfig(() => vueI18nConfigOptions);
```

BobbieGoede:
I'm sorry for the lack of response! I got reminded of this again since you opened #3503.

Have you tried this solution with a patch on your setup? If it does resolve your issue I am open to implementing this but with an opt-in flag, and maybe only enable it for production builds 🤔

It's unfortunate we can't really reproduce this issue, it would allow for properly debugging and fixing this issue rather than handling on the symptoms of the issue.

gbyesiltas:
@BobbieGoede 
Hey, no worries :) I only tried patching the package to log some more info about the error but I'll give it a go to to fix the actual error. I didn't do it because I didn't want to stay away from the actual package for possible version update mismatches, but if we can add it as an optIn feature later-on, that would work me, so I'll keep you posted

PR #3501, opened 2025-03-31
fix: delay module setup until `modules:done`

BobbieGoede:
### 📚 Description
This delays most of the module setup until `modules:done`, this ensures any modules that would like to hook into i18n build-time hooks such as `i18n:registerModule` have had the chance to do so before all options are combined in the final config.

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -39,82 +39,83 @@ export default defineNuxtModule<NuxtI18nOptions>({
     await initParser()
 
     /**
-     * Prepare options
+     * prepare options
      */
     prepareOptions(ctx, nuxt)
 
     /**
-     * nuxt layers handling ...
+     * allow other modules to setup first in case these use i18n hooks
      */
-    await prepareLayers(ctx, nuxt)
+    nuxt.hook('modules:done', async () => {
+      /**
+       * nuxt layers handling ...
+       */
+      await prepareLayers(ctx, nuxt)
+      /**
+       * setup runtime config
+       */
+      prepareRuntimeConfig(ctx, nuxt)
 
-    /**
-     * setup runtime config
-     */
-    // for public
-    prepareRuntimeConfig(ctx, nuxt)
+      /**
+       * resolve locale info and vue-i18n config path
+       */
+      await resolveLocaleInfo(ctx, nuxt)
 
-    /**
-     * resolve locale info and vue-i18n config path
-     */
-    await resolveLocaleInfo(ctx, nuxt)
+      /**
+       * setup nuxt/pages
+       */
+      await setupPages(ctx, nuxt)
 
-    /**
-     * setup nuxt/pages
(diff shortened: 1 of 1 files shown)


Tofandel:
This change causes warning to popup for any routes like this in route rules

```js
  routeRules: {
    '/hu/webaruhaz/**': {
      redirect: {
        to: '/hu/termek-kategoria/**',
        statusCode: 301,
      },
    },
 }
 ```
 
 
`WARN  [@nuxtjs/i18n]: Couldn't find AnalyzedNuxtPageMeta by NuxtPage (/hu/webaruhaz/:pathMatch(.*)), so no custom route for it`

In fact they appear to be double translated now
image


This change also seems to have the side effect that if you specifically registered a module after nuxt-i18n with 'pages:extend' hook to modify already translated routes, this will not work anymore. You now need to hook into `modules:done` first and THEN hook again

Before:

```js

    '@nuxtjs/i18n',
    defineNuxtModule({
      setup(_options, nuxt) {
        nuxt.hook('pages:extend', (pages) => {
          // Edit translated pages
        });
      }
    })
```

After:

```js
    '@nuxtjs/i18n',
    defineNuxtModule({
      setup(_options, nuxt) {
        nuxt.hook('modules:done', () => {
          nuxt.hook('pages:extend', (pages) => {
            // Edit translated pages
          });
        });
      }
    })
```

Might be something worth documenting because this could be a breaking change, ideally to avoid a breaking change there, the hooks the plugin defines should still be registered not within the modules:done hook but outside of it, to be registered in the correct order but the config part still resolved inside modules:done

BobbieGoede:
Sorry for the inconvenience, but I recently learned modules should call their hooks in `modules:done`, see https://github.com/nuxt/nuxt/issues/31674 for context.

I'm not sure what your use case is for further modifying the routes after i18n, the module assumes it is among the last ones to modify the routes to be able to localize all registered routes.

Tofandel:
My use case is because I'm duplicating some routes with path modification to handle a limitation of vue-router, but those routes are also translated with `i18n.pages` like so

```js
    pages: {
      'shop/[...category]/page/[page]': {
        en: '/product-category/[...category]/page/[page]',
        hu: '/termek-kategoria/[...category]/page/[page]',
      },
```

But this means that if I run the route duplication before nuxt-i18n, this translation will not apply to the duplicated route, because the path differs, so I have to run it after and simply duplicate the already translated routes and modify them there (Now thinking I guess I could just edit the nuxt config to also duplicate the i18n pages rules as well during the pages:extend and run it before i18n)

Could you look into the routeRules issue? As I'm sure this is not uncommon to add raw redirects for already translated content (seeing this was before this release the only way to do that). It seems the module now just tries to translate all the route rules, so maybe we just need a way to opt out in the rule, or check if the path already looks translated

Issue #3435, opened 2025-03-21
Roadmap to v10.0.0


BobbieGoede:
With [vue-i18n v10 soon entering its maintenance/hotfix-only status](https://vue-i18n.intlify.dev/guide/maintenance.html), and vue-i18n v12 entering alpha, we have some catching up to do.

Updating vue-i18n to its next major means breaking changes, albeit small ones in vue-i18n v11. We will take this as an opportunity to make some breaking changes in the form of cleanup and reorganization of config options.

Feel free to suggest any changes and please let me know if some change needs reconsidering.

- [x] Drop CJS, go ESM only
- [x] Raise minimum Node version to v20
- [x] Split up v9 (and earlier) documentation, see if we can move it to v9.i18n.nuxtjs.org
- [ ] ~~Move `typescript` to `devDependencies` - this should cause no issues after updating `@intlify/unplugin-vue-i18n` but needs to be tested.~~ requires release, may not be possible since next release will be vue-i18n v12 compatible
- [x] Update to Vue I18n v11 (see [breaking changes](https://vue-i18n.intlify.dev/guide/migration/breaking11.html))
  - [x] #3526 

- [x] Remove migration ease for `restructureDir` - require a directory name (keep default value `'i18n'`)
- [x] Drop `bundle.optimizeTranslationDirective` (#3238)
- [x] Drop `experimental.generatedLocaleFilePathFormat` (we use hashed virtual files now)
- [x] Strip `file`/`files` properties from `i18n.options.mjs` generated file
- [x] Remove deprecated functionality
  - [x] Drop `$resolveRoute()`
  - [x] Drop `$localeLocation()`
  - [x] Drop `useLocaleLocation`
- [x] Deprecate (remove in v11)
  - [x] `$localeHead` - reason: use-case unclear, `useLocaleHead` is more flexible and easier to maintain
  - [x] `$getRouteBaseName` - rename to `$routeBaseName` to be consistent with `useRouteBaseName` (still under consideration)
- [x] Promoting experimental options
  - [x] `experimental.switchLocalePathLinkSSR` - enable by default
  - [x] `experimental.hmr` - move to `hmr`
  - [x] `experimental.autoImportTranslationFunctions` - enable by default, move out of experimental (where?)
  - [x] `experimental.alternateLinkCanonicalQueries` - enable by default, consider deprecating (dropping in v11)

productdevbook:
Thank you. how can i add i18n support in nuxt app.config.ts?

[...]

PR #3527, opened 2025-04-05
docs: remove legacy documentation

BobbieGoede:
### 🔗 Linked issue
* #3435 

### 📚 Description
Split up v9 (and earlier) documentation, see if we can move it to v9.i18n.nuxtjs.org


BobbieGoede:
@danielroe @atinux

Not sure who to tag, I need some help with setting up old and new documentation 🙏 

I would like to move the docs to a url structure similar to Nuxt Content, by moving the current docs which are deployed from the `main` branch (eventually `v9`) to `v9.i18n.nuxtjs.org`, and use the docs on the `next` branch (eventually `main`) to deploy to `i18n.nuxtjs.org`.

I did try and see what I could change by myself but I think adding an additional sub-domain requires more permissions, and I can't deploy the `next` branch docs without having the current docs up somewhere.

Also, less important but, I would like the match the font and size of the logo/text of the other Nuxt projects, is there some workfile or design doc for this?

PR #3535, opened 2025-04-07
refactor: encapsulate i18n routing

BobbieGoede:
### 🔗 Linked issue
* https://github.com/nuxt/nuxt/discussions/28826

### 📚 Description
I think I am starting to understand why https://github.com/intlify/routing/tree/main/packages/vue-i18n-routing was structured in the way it was, while trying to encapsulate the logic again it is starting to look similar to the repo it came from.


BobbieGoede:
This is not done, my abstraction/encapsulation refactors are still messy, but it is functioning and going  in the right direction, so I'm merging it to prevent this from becoming a blocker.

PR #3555, opened 2025-04-16
docs: update localeDetector example cookie name with nuxt-i18n default

cwandev:
### 📚 Description

Align the localeDetector example cookie name in the docs with the @nuxtjs/i18n default '`i18n_redirected`' for consistency.


BobbieGoede (approved):
Thanks!

PR #3556, opened 2025-04-16
fix: use right way to get `bundle.optimizeTranslationDirective` option

rx-837:
### 📚 Description

As know nuxt have two way to pass options for modules.

The first way is dedicated key at the first level of config.

```
modules: [
  "@nuxtjs/i18n",
],

i18n: {
  // place options here
},
```

The second way is to pass the options as the second element of the array.

```
{
  modules: [
    [
      "@nuxtjs/i18n",
      {
        // place options here
      },
    ],
  ],
}
```

This PR used correct way to get `optimizeTranslationDirective` option to suppress warning in stdout.

### DIFF
--- a/src/prepare/options.ts
+++ b/src/prepare/options.ts
@@ -20,11 +20,7 @@ export function prepareOptions({ debug, logger, options }: I18nNuxtContext, nuxt
     )
   }
 
-  if (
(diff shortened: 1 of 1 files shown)


BobbieGoede (approved):
Thanks!

PR #3561, opened 2025-04-21
fix!: change head tag default `key` to match Unhead v2

rx-837:
### 📚 Description

Unhead v2 library have changes in API. The key `hid` now is removed and couldn't be used to dedupe head tags e.g link or meta. Current behavior to dedupe tags is usage key with name `key`. See https://nuxt.com/docs/getting-started/upgrade#unhead-v2 and https://unhead.unjs.io/docs/typescript/head/guides/core-concepts/handling-duplicates#using-custom-keys-for-precise-control

I change default key option from `hid` to `key` to avoid tags duplications. Also i fixed tests to pass it.


BobbieGoede:
Thank you for the PR! 

Since this is technically a breaking change I changed the target branch to `next`, it is still possible users are on earlier Nuxt versions that do not use Unhead v2 (or have opted to use v1 using `unhead: { legacy: true }`).

BobbieGoede (approved):

PR #3572, opened 2025-04-25
test: set runtime config using IPC signal

BobbieGoede:
### 📚 Description
Tried this about two years ago with https://github.com/nuxt/test-utils/pull/670, now with a bit more experience I think I managed to have something working.

This changes the way runtime config is set during tests, previously we restarted the server with new environment variables, this PR changes this to use IPC signals to update the process environment variables (and restore it after the test) while the server is running.


alexandre-hallaine:
nice!

PR #3576, opened 2025-04-25
test: tweak server startup

BobbieGoede:
### 📚 Description
This should speed up startup by not using `pnpx` but directly using `serve`, but ran into issues before when adjusting `waitForPort` timing and `pnpx` usage, will see if CI succeeds..


BobbieGoede:
The small test speed gains are not worth the flakiness

PR #3586, opened 2025-05-04
perf: strip unused messages from rendered HTML

BobbieGoede:
### 📚 Description
This optimization could make a big difference for larger projects, reduces the size of the initial page load and should be more performant client-side.

This might cause issues:
* Conditionally rendered translations may be missing, needs testing.
* An additional request is needed to retrieve the stripped messages client-side, but only on/before the first navigation, perhaps this could be done when the page is idle.

## Summary by CodeRabbit

- **New Features**
  - Added an experimental option to strip unused translation messages from server-rendered HTML, reducing page payload size.
  - Translation keys used during server-side rendering are now tracked per locale, enabling selective inclusion of only the required messages.

- **Bug Fixes**
  - Updated test expectations to reflect the correct initial count of loaded locale message keys.

- **Documentation**
  - Added type definitions and comments for the new experimental option and environment variable.

- **Chores**
  - Introduced utility functions for handling nested object properties related to translation messages.

### DIFF
--- a/specs/lazy_load/basic_lazy_load.spec.ts
+++ b/specs/lazy_load/basic_lazy_load.spec.ts
@@ -36,7 +36,7 @@ describe('basic lazy loading', async () => {
     // `en` present on initial load
     expect(await getLocalesMessageKeyCount(page)).toMatchInlineSnapshot(`
       {
-        "en": 4,
+        "en": 7,
       }
     `)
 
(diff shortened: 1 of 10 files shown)


BobbieGoede:
While these changes reduce the response time/size, it is disabled by default due to conditional translations being stripped out if they are not used during SSR.

The changes from both #3584 and #3579 also have negative performance implications due to the optimizations being done at runtime. 

More work needs to be done to keep perf on par with the latest beta release (which is faster than v9 stable currently), before publishing a new beta release.

Issue #3659, opened 2025-05-30
Builds with rolldown-vite + nuxt i18n fail


warflash:
### Environment

- Operating System: Windows_NT
- Node Version:     v23.10.0
- Nuxt Version:     3.17.4
- CLI Version:      3.25.1
- Nitro Version:    2.11.12
- Package Manager:  pnpm@8.15.9
- Builder:          -
- User Config:      devtools, modules, i18n, compatibilityDate
- Runtime Modules:  @nuxtjs/i18n@9.5.5
- Build Modules:    -

### Reproduction

https://github.com/warflash/nuxt-i18n-rolldown

### Describe the bug

Builds are failing

### Additional context

[Rolldown vite was just announced as technical preview,](https://voidzero.dev/posts/announcing-rolldown-vite) so I wanted to jump in and test against our project. Dev works well, builds seem to fail, also on the i18n starter with the override applied.

Given that vite-rolldown aims to be a drop in replacement it might just need time for the vite team to iron out the cause of this bug. Just wanted to raise the issue in case it's some internal tooling in the i18n project that won't be compatible with vite in the future.

### Logs

```shell
> nuxt build

[nuxi 21:06:47] Nuxt 3.17.4 with Nitro 2.11.12

[@nuxtjs/i18n 21:06:48]  WARN  bundle.optimizeTranslationDirective is enabled by default, we recommend disabling this feature as it causes issues and will be deprecated in v10.
Explicitly setting this option to true or false disables this warning, for more details see: https://github.com/nuxt-modules/i18n/issues/3238#issuecomment-2672492536

[nuxi 21:06:48] ℹ Building for Nitro preset: node-server
[21:06:51] ℹ Building client...

[21:06:51]  WARN  You have set optimizeDeps.esbuildOptions but this options is now deprecated. Vite now uses Rolldown to optimize the dependencies. Please use optimizeDeps.rollupOptions instead.

[21:06:51] ℹ rolldown-vite v6.3.14 building for production...
[21:06:52] ℹ ✓ 181 modules transformed.

[21:06:52]  ERROR  ✗ Build failed in 531ms


[nuxi 21:06:52]  ERROR  Nuxt Build Error: Build failed with 3 errors:

[PARSE_ERROR] Error: Unexpected token
   ╭─[ i18n/locales/en.json:1:1 ]
   │
 1 │ const resource = {
   │ ──┬──  
   │   ╰──── 
───╯

[PARSE_ERROR] Error: Unexpected token
[...]

PR #3666, opened 2025-05-31
feat: `vite-rolldown` support

BobbieGoede:
### 🔗 Linked issue
* #3659 

### 📚 Description
This PR is to check compatibility with `vite-rolldown` and which dep changes are required.

## Summary by CodeRabbit

- **Chores**
  - Upgraded the i18n tooling plugin dependency to a newer major version.
  - Adjusted a dev dependency version spec for build tooling.
  - Added a development dependency to support documentation tooling.
- **Refactor**
  - Improved i18n configuration handling so empty locale lists are treated as explicit empty sets rather than undefined.


src/bundler.ts
```diff
@@ -49,7 +49,7 @@ export async function extendBundler(ctx: I18nNuxtContext, nuxt: Nuxt) {
     ...options.customBlocks,
     allowDynamic: true,
     optimizeTranslationDirective: false,
-    include: localePaths.length ? localePaths : undefined
+    include: localePaths.length ? localePaths : []
```
BobbieGoede, line 52: @kazupon the `include` option has been inverted in `unplugin-vue-i18n` v11, is this intentional?

current:
```ts
// if (include == null)
include = [RE_RESOURCE_FORMAT, customBlockInclude]
// no changes to exclude
```

https://github.com/intlify/bundle-tools/pull/494/files#diff-065c42fc94971ffb5ea477bf1852c408c638f1fc567a356c8d4f814028d1b302R89

before:
```ts
// if (include == null)
// no changes to include - but I think `createFilter` will include everything by default for `[] | undefined`
exclude = '**/**'
```
https://github.com/intlify/bundle-tools/blob/93eab2bc2bf5244028ac9dd42f2a743ef66c2a76/packages/unplugin-vue-i18n/src/core/options.ts#L28

kazupon: @BobbieGoede 
Yeah!

With the module type of vite (rolldown) and the new filter mechanism called plugin hook filter, the filters of unplugin-vue-i18n are no longer constrained by `vite:json` plugin.

- plugin hook filter: https://rolldown.rs/guide/plugin-development#plugin-hook-filters
- module type: https://rolldown.rs/guide/in-depth/module-types

Previously, unplugin-vue-i18n handled the target json before the `vite:json` plugin due to the transform pipeline order enforced by `enforce: 'pre'`, but the plugin's transform was always handled. As a result, `include` and `exclude` options in unplugin-vue-i18n lacked the flexibility of the general filters in the rollup plugin.

roldown-vite and later, user plugins can prevent their transforms from being processed by the vite (roldown) side by returning `moduelType: 'js'`.
As a result, the filters in unplugin-vue-i18n v11 can now control which resource files are subject to transform using the `exclude` option.

BobbieGoede:
### @nuxtjs/i18n e2e tests

| OS      | `vite`         | `vite-rolldown` | time           | %       |
| ------- | -------------- | --------------- | -------------- | ------- |
| linux   | 370s (6m 10s)  | 325s (5m 25s)   | -45s (0m 45s)  | -12.16% |
| windows | 794s (13m 14s) | 604s (10m 4s)   | -190s (3m 10s) | -23.93% |

In case you're interested @TheAlexLichter

My guess is the bottleneck for the e2e tests is building each fixture (with various different build configs), and read/write fs related to that. But I'm curious whether the difference in test speed/duration is within the expected range 👀

package.json
```diff
@@ -87,7 +87,7 @@
     "@intlify/core": "^11.1.5",
     "@intlify/h3": "^0.6.1",
     "@intlify/shared": "^11.1.5",
-    "@intlify/unplugin-vue-i18n": "11.0.0-beta.3",
+    "@intlify/unplugin-vue-i18n": "https://pkg.pr.new/@intlify/bundle-utils@4c8824b",
```
kazupon: oh, not `@intlify/bundle-utils`, `@intlify/unplugin-vue-i18n` 😅 
You can copy the below and update with it
```
"@intlify/unplugin-vue-i18n": "https://pkg.pr.new/@intlify/unplugin-vue-i18n@4c8824b",
```

BobbieGoede: Ah! I must have been sleeping 🤦😂

BobbieGoede: @kazupon 
Looks like it is working with the correct dep installed 😅

kazupon: Thanks for your checking!
I'll merge and release unplugin-vue-i18n v11 latest version later

kazupon: I've just released out it
https://github.com/intlify/bundle-tools/releases/tag/v11.0.0-beta.4

src/bundler.ts
BobbieGoede: I was wondering, should nuxt-i18n conditionally pass `[]` or `undefined` by detecting if `rolldown-vite` is used?

PR #3667, opened 2025-05-31
[DRAFT] feat: upgrade `@intlify/unplugin-vue-i18n`

BobbieGoede:
### 📚 Description
This PR is to demonstrate the test failures when using vite rather than overriding with rolldown vite when using `@intlify/unplugin-vue-i18n` v11 beta.

## Summary by CodeRabbit

- **Chores**
  - Updated the "@intlify/unplugin-vue-i18n" dependency to a newer version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "@intlify/core": "^11.1.5",
     "@intlify/h3": "^0.6.1",
     "@intlify/shared": "^11.1.5",
-    "@intlify/unplugin-vue-i18n": "^6.0.8",
+    "@intlify/unplugin-vue-i18n": "^11.0.0-beta.4",
     "@intlify/utils": "^0.13.0",
     "@miyaneee/rollup-plugin-json5": "^1.2.0",
     "@nuxt/kit": "^3.17.4",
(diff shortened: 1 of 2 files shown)


BobbieGoede:
@kazupon 
This PR has the same changes as #3666 (tests succeed 🚀) but without the `rolldown-vite` override (test failures).

Is `@intlify/unplugin-vue-i18n` v11 incompatible with `vite`? It seems like the removal of the `vite:json` wrapping/transformation is what causes the test failures.

kazupon:
@BobbieGoede 

> Is `@intlify/unplugin-vue-i18n` v11 incompatible with `vite`? It seems like the removal of the `vite:json` wrapping/transformation is what causes the test failures.

Ah, I don't support still rollup-vite.
sorry, I forgot for nuxt vite both compatible cases 😅 

Okay, I'll try to support it

kazupon:
@BobbieGoede 
Could you test the nightly build version ?
https://github.com/intlify/bundle-tools/pull/497#issuecomment-2934656843

BobbieGoede:
@kazupon 
I updated to the preview release, in this PR it works! 🚀

I did the same in https://github.com/nuxt-modules/i18n/pull/3666, which fails to build and throws some strange node error 🤔 Maybe the setup in that branch needs to be different?

BobbieGoede:
Update was done in a separate PR

PR #3693, opened 2025-06-13
fix: getDefaultLocaleForDomain should respect runtime config

### LINKED ISSUE #2931: Overridden locale domains don't switch locale
### Environment

- Operating System: `Darwin`
- Node Version:     `v20.9.0`
- Nuxt Version:     `3.11.2`
- CLI Version:      `3.11.1`
- Nitro Version:    `2.9.6`
- Package Manager:  `pnpm@9.0.4`
- Builder:          `-`
- User Config:      `devtools`, `modules`, `i18n`
- Runtime Modules:  `@nuxtjs/i18n@8.3.0`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/nuxt-starter-dx8kkg?file=nuxt.config.ts

The reproduction will not work on stackblitz due to using the custom domains. Please clone it & test locally.

### Describe the bug

The documentation [says](https://i18n.nuxtjs.org/docs/guide/different-domains#runtime-environment-variables):
> Alternatively, to avoid the need for multiple builds, the locale domains can be overridden via runtime environment variables. The variable name should follow the format NUXT_PUBLIC_I18N_LOCALES_{locale code}_DOMAIN

However, based on my testing, this doesn't seem to work. The domains provided through `.env` **do** appear in the lang switcher (result of `switchLocalePath()` call), however, they don't switch the locale.

I'm not sure if I'm just doing something wrong or if this is a bug, but regardless, this issue either needs fixing or the documentation needs to be updated. I'd be happy to update the docs if necessary.

### Additional context

_No response_

### Logs

_No response_

cjpearson:
### 🔗 Linked issue

Resolves #2931 

### 📚 Description

`domainLocales` was being checked in `getDomainFromLocale` which ensured the URLs for switching locales are correct, but when resolving the current locale `domainLocales` was ignored which resulted in the fallback locale being used for translations.

This changes adds similar logic to `getDefaultLocaleForDomain` to respect `domainLocales` and runtime-configured domains.

### DIFF
--- a/specs/different_domains/different_domains.spec.ts
+++ b/specs/different_domains/different_domains.spec.ts
@@ -151,6 +151,16 @@ test.each([
   expect(dom.querySelector('#welcome-text').textContent).toEqual(header)
 })
 
+test.only('(#2931) detect using runtimeConfig domain', async () => {
+  const res = await undiciRequest('/', {
+    headers: {
+      host: 'kr.staging.nuxt-app.localhost'
+    }
+  })
+  const dom = getDom(await res.body.text())
+  expect(dom.querySelector('#welcome-text').textContent).toEqual('환영하다')
+})
+
 test('(#2374) detect with x-forwarded-host on server', async () => {
   const html = await $fetch('/', {
     headers: {
--- a/specs/different_domains/different_domains.spec.ts
+++ b/specs/different_domains/different_domains.spec.ts
@@ -151,7 +151,7 @@ test.each([
   expect(dom.querySelector('#welcome-text').textContent).toEqual(header)
 })
 
-test.only('(#2931) detect using runtimeConfig domain', async () => {
+test('(#2931) detect using runtimeConfig domain', async () => {
   const res = await undiciRequest('/', {
     headers: {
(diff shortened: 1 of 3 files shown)


specs/different_domains/different_domains.spec.ts
```diff
@@ -151,6 +151,16 @@ test.each([
   expect(dom.querySelector('#welcome-text').textContent).toEqual(header)
 })
 
+test.only('(#2931) detect using runtimeConfig domain', async () => {
```
BobbieGoede:
```suggestion
test('(#2931) detect using runtimeConfig domain', async () => {
```

BobbieGoede:
Thank you for your contribution! ❤️

We will need to port these changes to the `next` branch as well, PR welcome (on holiday atm)

specs/different_domains/different_domains.spec.ts
cjpearson: Oops 🙈 Thanks for catching that

cjpearson:
Thanks for the review! I can take a look at v10

PR #3726, opened 2025-07-13
feat(i18n): add hook to filter pages for i18n

### LINKED ISSUE #3716: Allow modules to modify route localization
### Environment

* Nuxt: `^3.x`
* @nuxtjs/i18n: `^9.x`
* Package manager: `pnpm`
* Node: `>=18`
* OS: macOS

#### 🧩 What I'm trying to do

I want to **dynamically disable translation for some pages**, such as `/admin` or any route matching a certain pattern, by injecting into the `i18n.pages` config like this:

```ts
i18n: {
  customRoutes: 'config',
  pages: {
    'admin': false
  }
}
```

But I want to do this **dynamically from another Nuxt module**, based on logic like:

* Whether route path starts with `/admin`
* Whether the file name includes certain keywords
* Optional user-defined exclude rule function

---

#### 📌 Problem

The `pages:extend` hook gives access to all generated Nuxt pages, and I can detect which pages I want to exclude, then inject:

```ts
nuxt.options.i18n.pages = {
  ...nuxt.options.i18n.pages,
  ...detectedPagesToExclude
}
```

However, **`@nuxtjs/i18n` appears to snapshot the config too early**, likely before `pages:extend` runs, and generates `i18n-options.mjs` using an internal `ctx.options`, so my dynamic changes to `nuxt.options.i18n.pages` do **not reflect** in the generated options.

---

#### ❓Question

Is there an **official or recommended way** to dynamically inject/disable page translation settings (`customRoutes: 'config'` + `pages`) at runtime or module time?

More specifically:

* Can you expose a Nuxt hook like `i18n:extendPagesConfig(pages)` before generating `i18n-options.mjs`?
* Or is there a safe way to hook into `ctx.options.pages` in a custom module?
* Or should we write a wrapper module that overrides the `i18n` module setup?

---

#### 🧪 What I tried

```ts
nuxt.hook('pages:extend', pages => {
  for (const page of pages) {
    if (page.path.startsWith('/admin') && page.name) {
      nuxt.options.i18n.pages[page.name] = false
    }
  }
})
```

But this doesn't reflect in the generated `.nuxt/i18n-options.mjs` file.

---

#### 🙏 Feature request?

Could you consider adding a hook like:

```ts
nuxt.hook('i18n:extendPagesConfig', (pages: NuxtPage[], currentConfig: Record<string, any>) => {
  // modify and return updated pages config
})
```

So that other modules (like `i18n-exclude`) can safely hook in and modify translation strategy for specific routes?

---

Let me know if there's already a supported way to do this.

Thanks a lot for the amazing work on `@nuxtjs/i18n` 🙌

### Additional context

_No response_

### Logs

```shell

```

icai:
### 🔗 Linked issue

resolves https://github.com/nuxt-modules/i18n/issues/3716

### 📚 Description

Allow modules to modify route localization

```
nuxt.hook('i18n:filterPages', (pages: NuxtPage[]) => {
    for (const page of pages) {
      if (page.path.startsWith('/admin')) {
        page.meta = page.meta || {}
        page.meta.i18n = false
      }
    }
})
```

## Summary by CodeRabbit

* **New Features**
  * Added a public hook to filter pages before localization, enabling per-page i18n control.
  * Included an example configuration and sample pages demonstrating disabling i18n for routes under /admin.

* **Bug Fixes**
  * Pages explicitly marked to disable i18n are now skipped during localization (early-exit).

* **Tests**
  * Added fixtures and example pages to validate i18n-disabled page behavior.


BobbieGoede:
Thank you for your PR!

After taking a look at the changes in this PR, I'm not sure about adding the `i18n:filterPages` hook. Replacing the hook usage in the `i18n-disable-pages` fixture with the `pages:extend` hook works as well 🤔

Effectively this change https://github.com/nuxt-modules/i18n/pull/3726/files#diff-eb1a20dccd87e92592b66458a2c4d10f5309b3b774fa41bcdc9925f3ad76c1f2R102 would partially enable the `customRoutes: 'meta'` option, which would be a breaking change and possibly unexpected behavior, so this will need some reconsidering.

icai:
@BobbieGoede This feature skips microcode compilation, so there is no conflict. Is there a priority issue? This priority must be relatively high, right？

icai:
@BobbieGoede This feature is very necessary. Can it be merged, or can related features be provided?

BobbieGoede:
I can't consider these changes while tests/CI fail, I'm pretty sure this PR contains breaking changes in its current state.

BobbieGoede (changes requested):
Please try to keep formatting changes to a minimum and ensure the tests still succeed with your changes.

PR #3736, opened 2025-07-20
docs: clarify redirectOn behavior change in v10 migration guide

### LINKED ISSUE #3735: Routes without locale prefix show 404 when using strategy: 'prefix' and redirectOn: 'root' in v10
### Environment

- **Operating System**: macOS
- **Node Version**: v20+ (bun runtime)
- **Nuxt Version**: 4.0.0
- **Package Manager**: bun@1.2.18
- **@nuxtjs/i18n Version**: 10.0.0
- **Strategy**: `prefix`
- **Deployment**: Cloudflare Workers

### Reproduction

**Steps to reproduce:**

1. Configure Nuxt i18n with `strategy: 'prefix'` and `detectBrowserLanguage.redirectOn: 'root'`
2. Create a page at `/app/pages/search.vue`
3. Access a route without locale prefix: `http://example.com/search`
4. Observe that the page shows "Page not found: /search"
5. Try accessing the same route with locale prefix: `http://example.com/zh/search` - this works correctly

**Current Configuration:**

```typescript
// nuxt.config.ts
i18n: {
  locales: [
    { code: 'en', name: 'English', language: 'en-US', file: 'en.ts' },
    { code: 'zh', name: '简体中文', language: 'zh-Hans-CN', file: 'zh.ts' },
    { code: 'zh-hant', name: '繁體中文', language: 'zh-Hant-CN', file: 'zh-hant.ts' },
    { code: 'ja', name: '日本語', language: 'ja-JP', file: 'ja.ts' },
    { code: 'ko', name: '한국어', language: 'ko-KR', file: 'ko.ts' },
  ],
  defaultLocale: 'zh',
  strategy: 'prefix',
  detectBrowserLanguage: {
    useCookie: true,
    cookieKey: 'panews-language',
    redirectOn: 'root',
  },
  baseUrl: 'https://www.panewslab.com',
}
```

### Describe the bug

After upgrading from `@nuxtjs/i18n` v9 to v10, routes without locale prefixes (e.g., `/search`) are returning 404 errors instead of being properly redirected to the localized version (e.g., `/zh/search`).

**Expected behavior (worked in v9):**
- Accessing `/search` should redirect to `/zh/search` (or user's preferred locale)
- The redirect should happen seamlessly without showing a 404

**Actual behavior (broken in v10):**
- Accessing `/search` shows "Page not found: /search"
- Direct access to `/zh/search` works perfectly
- The browser network tab shows that a redirect does occur, but the page still displays the 404 error

### Analysis

The issue appears to be related to the combination of:
1. `strategy: 'prefix'` - which requires all routes to have locale prefixes
2. `redirectOn: 'root'` - which only handles redirects for the root path (`/`)

In v10, it seems the redirect logic for non-root paths has changed, causing routes like `/search` to return 404 instead of being redirected to their localized equivalents.

### Additional context

This regression affects production websites that rely on SEO-friendly URLs and proper redirect behavior. The breaking change from v9 to v10 is causing 404 errors for users accessing non-prefixed routes that previously worked correctly.

The issue specifically impacts:
- Direct URL access (bookmarks, external links)
- Search engine crawlers
- Users sharing non-prefixed URLs

Our specific use case requires maintaining the exact configuration:
- `strategy: 'prefix'` (all routes must have locale prefixes)
- `redirectOn: 'root'` (only redirect on root access for performance reasons)

This configuration worked perfectly in v9 and should continue to work in v10.

### Logs

No specific error logs, but the behavior can be observed in browser developer tools showing:
1. Initial request to `/search`
2. 302 redirect to `/zh/search`
3. Page displays "Page not found: /search" despite the redirect

medz:
## Summary

This PR clarifies the `redirectOn` behavior change in the v10 migration guide to help developers understand and properly configure their language detection settings.

## Background

In v9, some combinations of `strategy` and `redirectOn` options worked unexpectedly. Specifically, when using `strategy: 'prefix'` with `redirectOn: 'root'`, non-root paths (e.g., `/search`) would automatically redirect to their localized versions (e.g., `/zh/search`), even though this contradicted the documented behavior.

v10 corrected this behavior to strictly follow the documentation, but this change wasn't explicitly documented in the migration guide, leading to confusion for developers upgrading from v9.

## Changes

- Added a new "Behavior Changes" section to the migration guide
- Documented the `redirectOn` behavior correction 
- Provided clear migration instructions for affected configurations
- Explained the impact on existing projects

## Migration Path

For users who relied on the previous unintended behavior:

**Before (v9 - worked unexpectedly):**
```typescript
detectBrowserLanguage: {
  redirectOn: 'root',  // Only should redirect root path
}
```

**After (v10 - correct configuration):**
```typescript
detectBrowserLanguage: {
  redirectOn: 'all',   // Redirects all paths as intended
}
```

## Related Issues

Fixes #3735

## Testing

- [x] Documentation builds successfully
- [x] Migration guide section is clear and actionable
- [x] Examples demonstrate the correct configuration

## Impact

This documentation update will help developers:
1. Understand why their v9 configuration no longer works in v10
2. Know exactly how to fix their configuration
3. Avoid confusion about the intended behavior of `redirectOn` options

The change affects projects using `strategy: 'prefix'` that relied on the previous unintended behavior where `redirectOn: 'root'` also handled non-root paths.

## Summary by CodeRabbit

* **Documentation**
  * Updated the migration guide to clarify changes in language detection and redirection behavior for v10, including stricter handling of root path redirection and explicit migration instructions for configuration updates.

### DIFF
--- a/docs/content/docs/02.guide/90.migrating.md
+++ b/docs/content/docs/02.guide/90.migrating.md
@@ -33,7 +33,7 @@ The following [Configuration options](/docs/api/options) have been changed, depr
 
 ### Browser Language Detection
 
-Language detection and redirection have been improved to strictly follow documented behavior. In v9, some combinations of `strategy` and `redirectOn` options worked unexpectedly, which has been corrected in v10.
+Language detection and redirection have been improved to follow documented behavior strictly. In v9, some combinations of `strategy` and `redirectOn` options worked unexpectedly, which has been corrected in v10.
 
 **Key Change**: When using `strategy: 'prefix'` with `redirectOn: 'root'`, non-root paths (e.g., `/search`) will no longer automatically redirect to their localized versions (e.g., `/zh/search`).
 
@@ -41,14 +41,24 @@ Language detection and redirection have been improved to strictly follow documen
 
 ```ts
 // v9 configuration (worked unexpectedly)
-detectBrowserLanguage: {
-  redirectOn: 'root',  // ⚠️ In v10 this will only redirect the root path
-}
+export default defineNuxtConfig({
+  i18n: {
+    strategy: 'prefix',
+    detectBrowserLanguage: {
+      redirectOn: 'root', // ⚠️ In v10 this will only redirect the root path
+    }
+  }
+})
 
 // v10 configuration (correct behavior)
-detectBrowserLanguage: {
-  redirectOn: 'all',   // Redirects all paths as documented
-}
+export default defineNuxtConfig({
+  i18n: {
+    strategy: 'prefix',
+    detectBrowserLanguage: {
+      redirectOn: 'all',  // Redirects all paths as documented
+    }
+  }
+})
 ```
 
 **Impact**: This affects projects using `strategy: 'prefix'` that relied on the previous unintended behavior where `redirectOn: 'root'` also handled non-root paths.
--- a/docs/content/docs/02.guide/90.migrating.md
+++ b/docs/content/docs/02.guide/90.migrating.md
@@ -102,8 +102,7 @@ The following options have been removed from runtime config:
 
 The generated options files in your projects are meant for internal use by this module at runtime and should never be used, more properties may be removed in the future.
 
--Future changes to these internal options will not be documented in the migration guide. If you have use cases for these options, please open an issue describing your use case so we can evaluate if we can support it in a different way.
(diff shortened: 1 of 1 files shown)


BobbieGoede (approved):
Thank you for contributing!

PR #3802, opened 2025-09-02
docs: provide correct navigation to search bar

cernymatej:
### 📚 Description
I noticed that search isn't working in the docs. After investigating, I found that incorrect navigation was being passed into the `<LazyUContentSearch />` component.

It looks like leftover code from a previous refactor 🤔

---

I'm not entirely sure, but I think we can safely remove the following code too. I couldn't find any places where `'navigation'` is being injected.

https://github.com/nuxt-modules/i18n/blob/28e60ac34610fb335e444756f75ae76f871ec7d7/docs/app.vue#L23-L26

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved docs search reliability and consistency across pages by standardizing the navigation data used for searching.
* **Refactor**
  * Simplified how navigation is sourced for the docs search to provide more consistent behavior and smoother navigation.
* **Chores**
  * Removed the "V9" entry from the documentation navigation so it no longer appears in menus.


BobbieGoede (approved):
Thanks! These are remnants from the v9 branch, which has documentation for versions 7, 8 and 9 and filters the search based on which version docs are viewed.

Issue #3377, opened 2025-02-25
useSetI18nParams should also support translation of query args


Tofandel:
### Describe the feature

Query args and parameters are pretty much equivalent when it comes to routing

I believe we should be able to translate query args as well the same way we can translate route params

```js

const setI18nParams = useSetI18nParams({
  translatedQuery: ['minPrice', 'maxPrice'],
});

// /en/page?minPrice=10

setI18nParams({
  'hu': { minPrice: 300 }
});

// /hu/page?minPrice=300
```

It probably goes the same for the hash but that one is a bit more problematic than query args because the hash is not transmitted to the server, causing hydration issues as only the frontend is able to access it

### Additional information

- [x] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [x] Read the [contribution guide](https://nuxt.com/docs/community/contribution) (The contribution guideline of nuxt-modules/i18n is compliant with Nuxt too).
- [x] Check existing [discussions](https://github.com/nuxt-modules/i18n/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

vad1ym:
Up

PR #3869, opened 2025-11-03
Allow localization of query parameters via useSetI18nParams

Tofandel:
### 🔗 Linked issue

https://github.com/nuxt-modules/i18n/issues/3377

### 📚 Description

Since we know which params and query args are available on a route we are already on, we can split them correctly between params and query without additional changes to setI18nParams

Also prevents the hash being included in switchLocalePath if the app is being hydrated from SSR (because the hash is never sent to the server and would be missing from the SSR, causing hydration mismatch)

## Summary by CodeRabbit

* **Bug Fixes**
  * Locale-switch routing now merges localized dynamic params and query keys correctly.
  * URL hash handling during SSR hydration fixed so hashes aren’t incorrectly preserved.

* **New Features**
  * Added SSR-hydration state detection to refine locale-switch path generation.
  * UI: added a "Translate query" link on product pages to generate translated query parameters.

* **Tests**
  * Added tests validating translated query params, locale-switch behavior during hydration, and suppression of hydration warnings.

### DIFF
--- a/specs/routing/routing-tests.ts
+++ b/specs/routing/routing-tests.ts
@@ -103,12 +103,28 @@ export async function switchLocalePathTests() {
   await page.waitForURL(url('/ja/about?foo=b%C3%A4r&four=%E5%9B%9B'))
   expect(await page.locator('#switch-locale-path .ja').innerText()).toEqual('/ja/about?foo=b%C3%A4r&four=%E5%9B%9B')
   expect(await page.locator('#switch-locale-path .en').innerText()).toEqual('/en/about?foo=b%C3%A4r&four=%E5%9B%9B')
+  const forbidden = [
+    'Hydration completed but contains mismatches.',
+    'Note: this mismatch is check-only'
+  ]
 
+  const logs: string[] = []
+
+  page.on('console', msg => {
+    const text = msg.text()
+    if (forbidden.some(f => text.includes(f))) {
+      logs.push(text)
+    }
+  })
   await gotoPath(page, '/ja/about#foo=bar')
   // path hash is not sent to the server, so we need to wait for the client to hydrate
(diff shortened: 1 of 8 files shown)


specs/routing/routing-tests.ts
```diff
@@ -103,12 +103,28 @@ export async function switchLocalePathTests() {
...
+    }
+  })
   await gotoPath(page, '/ja/about#foo=bar')
   // path hash is not sent to the server, so we need to wait for the client to hydrate
   await page.waitForFunction(() => !window.useNuxtApp?.().isHydrating)
-  expect(await page.locator('#switch-locale-path .ja').innerText()).toEqual('/ja/about#foo=bar')
-  expect(await page.locator('#switch-locale-path .en').innerText()).toEqual('/en/about#foo=bar')
+
+  // We don't force a rerender on mounted, so the hash will not be present for switchLocale
+  expect(await page.locator('#switch-locale-path .ja').innerText()).toEqual('/ja/about')
```
Tofandel, line 124: I modified this test because the current behavior would have logged this in the console

```
[Vue warn]: Hydration attribute mismatch on <a href=​"/​hu/​termek-kategoria/​nappali/​" class title=​"Magyar">​HU​</a>​ 
  - rendered on server: href="/hu/termek-kategoria/nappali/"
  - expected on client: href="/hu/termek-kategoria/nappali/#quote"
  Note: this mismatch is check-only. The DOM will not be rectified in production due to performance overhead.
  You should fix the source of the mismatch. 

Hydration completed but contains mismatches.
```

Meaning in a production build, the dom would not have been corrected like on the playwright test and an error about hydration mismatch would have been logged.

There is sadly not much that we can do about this unless we change the return of `switchLocalePath` to a computed so that we can update it once the hydration is done. (Which would be a breaking change)

PR #3871, opened 2025-11-06
fix: skip localized redirection on nuxt error routes

### LINKED ISSUE #3870: Nuxt Error Page Incorrectly Recognized as Dynamic Route
### Environment

- **Nuxt version**: 4.1.3
- **Nitro version**: 2.12.7
- **@nuxtjs/i18n**: 10.2.0

### Reproduction

https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-lto6a9ap?file=app%2Fpages%2F[depth]%2Findex.vue

### Describe the bug
https://bobbiegoedenuxti18nstarterlto6-x3te--3000--cf284e50.local-credentialless.webcontainer.io/en/__nuxt_error
Even directly accessing `/en/__nuxt_error` triggers the redirect/routing logic instead of showing an error page.

Image

## Bug Description

When using `strategy: 'prefix'` with a root-level dynamic route (e.g., `pages/[param]/`), the `__nuxt_error` path is incorrectly matched as a dynamic route parameter instead of being recognized as Nuxt's error page, causing infinite redirects or double redirects.

### Problem Details

**Location**: `src/runtime/shared/matching.ts:19-24`

**Root Cause**: Nuxt's error page uses the special path `__nuxt_error` (defined as `NUXT_ERROR_SIGNATURE`).

When there's a dynamic route at the root level (`pages/[param]/`):
1. The router matcher treats `__nuxt_error` as a valid value for `[param]`
2. `isExistingNuxtRoute('__nuxt_error')` returns `true` (matches the dynamic route)
3. **`__nuxt_error` is incorrectly treated as a dynamic route parameter instead of an error page**
4. i18n redirect logic is executed on the error page
5. This causes infinite redirects

### Solution

Add an explicit check for `NUXT_ERROR_SIGNATURE` inside the `isExistingNuxtRoute` function to prevent the error page from being matched as a dynamic route.

**Modified code**:

```typescript
export function isExistingNuxtRoute(path: string) {
  if (path === '') { return }
  // Skip if this is a Nuxt error page (even if it matches a dynamic route)
  if (path === NUXT_ERROR_SIGNATURE) { return }
  
  const resolvedMatch = matcher.resolve({ path }, { path: '/', name: '', matched: [], params: {}, meta: {} })

  return resolvedMatch.matched.length > 0 ? resolvedMatch : undefined
}
```

### Logs

```shell

```

ziyoonee:
### 🔗 Linked issue
resolves #3870

### 📚 Description

Fixes an issue where `__nuxt_error` is incorrectly matched as a dynamic route parameter when root-level dynamic routes exist (e.g., `pages/[param]/`), causing infinite redirects with `strategy: 'prefix'`.

**Problem:**
- When a project has root-level dynamic routes, the router matcher treats `__nuxt_error` as a valid parameter value
- `isExistingNuxtRoute('__nuxt_error')` returns `true`, triggering i18n redirect logic on error pages

**Solution:**
- Added explicit check for `NUXT_ERROR_SIGNATURE` in `isExistingNuxtRoute()` 
- Error pages are now excluded from route matching regardless of dynamic route patterns

**Changes:**
- `src/runtime/shared/matching.ts`: Added early return when path matches `NUXT_ERROR_SIGNATURE`

## Summary by CodeRabbit

* **Bug Fixes**
  * Error pages that end with "/__nuxt_error" are now excluded from route resolution, preventing them from being treated as valid dynamic routes and improving navigation consistency when errors occur.

### DIFF
--- a/src/runtime/shared/matching.ts
+++ b/src/runtime/shared/matching.ts
@@ -1,6 +1,7 @@
 import { joinURL, parsePath, withLeadingSlash } from 'ufo'
 import { createRouterMatcher } from 'vue-router'
 import { i18nPathToPath, pathToI18nConfig } from '#build/i18n-route-resources.mjs'
+import { NUXT_ERROR_SIGNATURE } from '#app/composables/error'
 
 const matcher = createRouterMatcher([], {})
 for (const path of Object.keys(i18nPathToPath)) {
@@ -18,6 +19,9 @@ const getI18nPathToI18nPath = (path: string, locale: string) => {
 
 export function isExistingNuxtRoute(path: string) {
   if (path === '') { return }
+  // Skip if this is a Nuxt error page (even if it matches a dynamic route)
+  if (path === NUXT_ERROR_SIGNATURE) { return }
+
   const resolvedMatch = matcher.resolve({ path }, { path: '/', name: '', matched: [], params: {}, meta: {} })
 
   return resolvedMatch.matched.length > 0 ? resolvedMatch : undefined
--- a/src/runtime/shared/matching.ts
+++ b/src/runtime/shared/matching.ts
@@ -1,7 +1,6 @@
 import { joinURL, parsePath, withLeadingSlash } from 'ufo'
 import { createRouterMatcher } from 'vue-router'
 import { i18nPathToPath, pathToI18nConfig } from '#build/i18n-route-resources.mjs'
-import { NUXT_ERROR_SIGNATURE } from '#app/composables/error'
 
 const matcher = createRouterMatcher([], {})
 for (const path of Object.keys(i18nPathToPath)) {
@@ -19,8 +18,9 @@ const getI18nPathToI18nPath = (path: string, locale: string) => {
 
 export function isExistingNuxtRoute(path: string) {
   if (path === '') { return }
-  // Skip if this is a Nuxt error page (even if it matches a dynamic route)
-  if (path === NUXT_ERROR_SIGNATURE) { return }
+  // TODO: path should not have base url - check if base url is stripped earlier
+  // skip nuxt error route - this path is hardcoded within nitro context code in nuxt
+  if (path.endsWith('/__nuxt_error')) { return }
 
   const resolvedMatch = matcher.resolve({ path }, { path: '/', name: '', matched: [], params: {}, meta: {} })
 


src/runtime/shared/matching.ts
```diff
@@ -1,6 +1,7 @@
 import { joinURL, parsePath, withLeadingSlash } from 'ufo'
 import { createRouterMatcher } from 'vue-router'
 import { i18nPathToPath, pathToI18nConfig } from '#build/i18n-route-resources.mjs'
+import { NUXT_ERROR_SIGNATURE } from '#app/composables/error'
```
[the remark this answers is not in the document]
ziyoonee: The current import path is correct. `NUXT_ERROR_SIGNATURE` is exported from `#app/composables/error`, not from `#app` directly.

BobbieGoede:
```suggestion
```

src/runtime/shared/matching.ts
```diff
@@ -18,6 +19,9 @@ const getI18nPathToI18nPath = (path: string, locale: string) => {
 
 export function isExistingNuxtRoute(path: string) {
   if (path === '') { return }
+  // Skip if this is a Nuxt error page (even if it matches a dynamic route)
+  if (path === NUXT_ERROR_SIGNATURE) { return }
```
BobbieGoede:
```suggestion
  // TODO: path should not have base url - check if base url is stripped earlier
  // skip nuxt error route - this path is hardcoded within nitro context code in nuxt
  if (path.endsWith('/__nuxt_error')) { return }
```

These functions could be used in both nuxt and nitro contexts, we can't import from `#app` and nuxt doesn't expose it through virtuals for nitro/server context.

src/runtime/shared/matching.ts
```diff
@@ -18,6 +19,9 @@ const getI18nPathToI18nPath = (path: string, locale: string) => {
 
 export function isExistingNuxtRoute(path: string) {
   if (path === '') { return }
+  // Skip if this is a Nuxt error page (even if it matches a dynamic route)
```
BobbieGoede:
```suggestion
```

BobbieGoede (approved):
Thank you for contributing!

PR #3900, opened 2025-12-25
fix: handle falsy `options.i18n` types

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/32949

### 📚 Description

as `false` is now a valid type, this performs a `&&` check rather than using optional chaining.

## Summary by CodeRabbit

* **Chores**
  * Improved stability by adding defensive null/undefined checks around internationalization configuration access to prevent potential runtime issues.

* **Tests**
  * Updated a bundle size snapshot expectation to reflect a minor size delta.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/test/bundle.test.ts
+++ b/test/bundle.test.ts
@@ -58,7 +58,7 @@ describe.skipIf(process.env.ECOSYSTEM_CI || isWindows)(
         {
           "module": "69.4k",
           "module (without vue-i18n)": "26.0k",
-          "vue-i18n": "43.4k",
+          "vue-i18n": "43.3k",
           "vue-i18n (without message compiler)": "26.8k",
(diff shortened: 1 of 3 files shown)


test/bundle.test.ts
```diff
@@ -58,7 +58,7 @@ describe.skipIf(process.env.ECOSYSTEM_CI || isWindows)(
         {
           "module": "69.4k",
           "module (without vue-i18n)": "26.0k",
-          "vue-i18n": "43.3k",
+          "vue-i18n": "43.4k",
```
danielroe: the test was passing before without changing the snapshot

PR #3911, opened 2026-01-22
fix: prevent redirection to catch-all for unlocalized routes

### LINKED ISSUE #3910: Catch all roue and ssr navigation
### Environment

|                      |                           |
| -------------------- | ------------------------- |
| **Operating system** | `macOS 25.2.0`            |
| **CPU**              | `Apple M3 Pro (11 cores)` |
| **Node.js version**  | `v22.21.0`                |
| **nuxt/cli version** | `3.32.0`                  |
| **Package manager**  | `yarn@4.0.2`              |
| **Nuxt version**     | `3.18.1`                  |
| **Nitro version**    | `2.11.13`                 |
| **Builder**          | `vite@7.0.6`              |
| **Config**           | `-`                       |
| **Modules**          | `-`                       |

### Reproduction

https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-pfrzjnzy?file=app%2Fapp.vue

### Describe the bug

Hi,

I am trying to integrate a headless CMS (Storyblok) into my Nuxt / nuxt‑i18n project.
I need to create a catch‑all route and a preview route to handle the business requirements.
The preview route has nuxt‑i18n disabled because I do not need localization on that route.
The problem is that, when I navigate to the preview route with an SSR request, the matched route is the catch‑all one, because the module prepends the locale to the path, creating a redirect before the correct route is matched.

This does not seem to be a Nuxt bug, and I think it may be related to issue #3871.

In the reproduction demo I provided, you can see the problem.
The expected behavior is that both SSR and non‑SSR navigation match the preview route.

Thanks for your support.

### Additional context

_No response_

### Logs

```shell

```

fedetorre:
### 🔗 Linked issue

resolves https://github.com/nuxt-modules/i18n/issues/3910

### 📚 Description

There is a bug in the latest version of nuxt-i18n. If you have a catch‑all route (e.g. [...,slug].vue) defined in the pages directory, any routes that have i18n: false in their meta cannot be reached because the catch‑all route matches them first and triggers a redirect.
The expected behavior is that routes with i18n: false should take precedence over localized routes.

Further details and a reproduction are available in the related issue: https://github.com/nuxt-modules/i18n/issues/3910
 

## Summary by CodeRabbit

* **Refactor**
  * Improved internationalization routing to properly track and match routes that are intentionally not localized alongside localized variants.

* **Bug Fixes**
  * Prevents undesired redirects for routes disabled for localization, preserving original paths on SSR and CSR.

* **Tests**
  * Added tests to verify non-localized routes remain accessible and behavior is preserved.

* **Chores**
  * Added example project fixtures, sample pages, and an API endpoint to exercise i18n behaviors.


BobbieGoede:
Thank you for your PR!

Could you add a test so we can prevent regressions in the future? 🙏

fedetorre:
> Thank you for your PR!
> 
> Could you add a test so we can prevent regressions in the future? 🙏

@BobbieGoede I create the e2e test for the issue #3910 tell me if is it ok for you!

PR #3920, opened 2026-02-03
fix: respect trailingSlash on root

divine:
### 📚 Description

Resolves an issue when visiting: `example.com` it redirects to `example.com/en` while it should redirect to `example.com/en/`

It doesn't respect the `trailingSlash` option which is breaking change from v9 to v10.

I don't know where to push the test file as `trailingSlash` option isn't a runtime config anymore.

## Summary by CodeRabbit

* **New Features**
  * Added configurable trailing slash formatting for localized paths. Localized URLs can now be automatically formatted with or without trailing slashes based on your application configuration settings.


divine:
Hello,

@BobbieGoede can you please take a look please?

Thanks!

BobbieGoede (approved):
LGTM, thanks!

PR #3938, opened 2026-02-16
fix: add warning for missing domains when multiDomainLocales is enabled

DotwoodMedia:
### 📚 Description
If you have multiDomainLocales enabled now, you’ll get a warning about the defaultLocale, but it’s not required when using multiDomainLocales.

### 🗒️ Note
Developed by: Social Deal (@socialdeal)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation that warns when multi-domain locale mode is enabled but no locale has domains configured, helping catch configuration issues early.
  * Improved warning behavior for default-strategy detection: clearer guidance when a default strategy is present but no default locale is set and multi-domain mode is not active.

### DIFF
--- a/src/prepare/options.ts
+++ b/src/prepare/options.ts
@@ -34,9 +34,9 @@ export function prepareOptions({ options }: I18nNuxtContext, nuxt: Nuxt) {
   }
 
   if (hasMultiDomainLocales) {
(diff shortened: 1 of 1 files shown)


BobbieGoede (approved):
LGTM, thanks!

PR #3958, opened 2026-04-01
fix: inline `@intlify/h3` to avoid dupes

danielroe:
### 📚 Description

otherwise we can end up with different versions of `@intlify/utils/h3` and `@intlify/h3`

https://github.com/nuxt/ecosystem-ci/actions/runs/23833781405/job/69473060451

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to include an additional dependency so it’s picked up by transpilation, optimization, and TypeScript hoisting.
  * Added a type-only import to server routing code; no runtime behavior, API surface, or handler logic changed.

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -76,6 +76,7 @@ export default defineNuxtModule<NuxtI18nOptions>({
       '@intlify/core-base',
       '@intlify/utils',
       '@intlify/utils/h3',
+      '@intlify/h3',
       '@intlify/message-compiler',
     ]
 
--- a/src/nitro.ts
+++ b/src/nitro.ts
@@ -41,11 +41,8 @@ export async function setupNitro(ctx: I18nNuxtContext, nuxt: Nuxt) {
     { name: 'defineI18nLocaleDetector', from: ctx.resolver.resolve('runtime/composables/server') },
   ])
 
-  const h3UtilsExports = await resolveModuleExportNames(resolveModule('@intlify/utils/h3'))
-  addServerImports([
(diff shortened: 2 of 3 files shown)


danielroe:
@BobbieGoede  do you know why we replace one set of h3 exports with another?

BobbieGoede:
> @BobbieGoede do you know why we replace one set of h3 exports with another?

I'm not sure which exports/replacements you mean exactly, `@intlify/h3` adds utilities but does not replace any if I recall correctly.

/cc @kazupon

kazupon:
@danielroe 
As @BobbieGoede mentioned,  `@intlify/h3` does not replace any h3 exports.

It re-exports all utilities from `@intlify/utils/h3` (e.g. `getHeaderLocale`, `getCookieLocale`, `getPathLocale`, etc.) via `export * from '@intlify/utils/h3'` and adds its own i18n middleware/translation utilities on top (`defineI18nMiddleware`, `useTranslation`, `detectLocaleFromAcceptLanguageHeader`).

The re-export from `@intlify/utils/h3` is just a convenience so that consumers can import everything from `@intlify/h3` without needing to depend on `@intlify/utils` directly.

I can understand this could cause the duplication issue you're seeing when both packages are resolved separately.

BobbieGoede:
@danielroe is this PR still relevant to fix the issue you ran into?

danielroe:
i'm not sure - ecosystem ci is blocked by the ts-expect-error that my other pr removes - i'll check again after that merges

danielroe:
nope! we are good now

Issue #3970, opened 2026-04-28
v10: serving lazy-loaded locale messages from a CDN / surviving blue–green deployments without sticky sessions


Hronom:
### Describe the feature

Hi 👋 — first off, thanks for v10, the migration was smooth.

Filing this as a feature/docs request after a deep dive into the v10.3.0 source and the published docs. The concern is real-world deployment topology: a Nuxt SSR app fronted by CloudFront (or any similar CDN), with a separate static-asset origin for `_nuxt/*` and a dynamic origin for the Nitro server. In our case we use blue/green deploys without sticky sessions.

### Background — what v10 does today

- `lazy` was removed; lazy-load is now the default. ([migration guide](https://i18n.nuxtjs.org/docs/guide/migrating))
- All locale messages are served from a Nitro dynamic route at `/_i18n/<hash>/<locale>/messages.json` ([`module.mjs:1477-1480`](https://github.com/nuxt-modules/i18n/blob/main/src/module.ts), [`runtime/server/routes/messages.js`](https://github.com/nuxt-modules/i18n/blob/main/src/runtime/server/routes/messages.ts)).
- The hash is a content-based SHA-256 of the concatenated locale files ([`module.mjs:221-238`](https://github.com/nuxt-modules/i18n/blob/main/src/module.ts)).
- The handler **does not validate the hash** — it reads only the `:locale` param and returns the current deployment's messages for that locale.

### The two pain points

**1. There is no documented way to serve `/_i18n/*` from a CDN / static origin.**

I checked every docs page and every option:

- `serverRoutePrefix` is described as a *path* prefix only (default `/_i18n`); no support for absolute URLs.
- `baseUrl` (and `runtimeConfig.public.i18n.baseUrl`) is scoped to hreflang/SEO per the [Options docs](https://i18n.nuxtjs.org/docs/api/options).
- The [Runtime config API](https://i18n.nuxtjs.org/docs/api/runtime-config) exposes only `baseUrl` and `domainLocales`.
- The [Lazy-load translations guide](https://i18n.nuxtjs.org/docs/guide/lazy-load-translations) does not mention CDN/prerender.
- No `nitro.prerender`/`routeRules` recipe is documented to emit `/_i18n/<hash>/<locale>/messages.json` into `.output/public/`.

[...]

PR #3976, opened 2026-04-29
feat: first-class CDN support for lazy-loaded messages

Hronom:
## Summary

Adds a top-level `cdnURL` option so lazy-loaded locale messages can be served from a CDN / static origin instead of the dynamic Nitro server.

When `cdnURL` is set:

- **Client-side** fetches target `<cdnURL><serverRoutePrefix>/<hash>/<locale>/messages.json`. SSR keeps using a relative URL so the Nitro server doesn't bounce through the CDN for content the local handler can serve.
- **`nuxt build`** auto-pushes the hashed messages routes into `nitro.prerender.routes`, so the JSON files are emitted into `.output/public/_i18n/<hash>/<locale>/messages.json` ready to be uploaded alongside `_nuxt/*` chunks.

This puts locale messages on the same correctness footing as `_nuxt/*` chunks for blue/green deployments without sticky sessions — each color's HTML references its own immutable, content-hashed messages URL, so a request landing on the wrong color still gets matching messages.

Closes / addresses item 4 of #3970.

### Why this design

Minimal, opt-in, no behavior change when unset:

- `cdnURL=''` (default) → `__I18N_CDN_URL__` is `''`, all URL construction collapses back to the existing relative path. The default rendering and prerender paths are untouched.
- The constant is injected the same way as `__I18N_SERVER_ROUTE__` / `__I18N_LOCALE_HASHES__`, so this fits the existing bundler/define pattern.
- Server-side fetches stay relative on purpose. This avoids: (a) SSR latency for an extra hop, and (b) network configurations where the Nitro container can't reach the CDN.

### Files changed

- `src/types.ts`, `src/constants.ts` — new top-level `cdnURL?: string` option (default `''`).
- `src/bundler.ts`, `src/env.d.ts` — new `__I18N_CDN_URL__` build-time constant.
- `src/runtime/context.ts` — client-side URL prefix only on `import.meta.client`.
- `src/module.ts` — when `cdnURL` is set, push messages routes into `nitro.prerender.routes` so a regular `nuxt build` produces CDN-ready static assets.
- `docs/content/docs/04.api/00.options.md` — new `## cdnURL` section with the blue/green deployment recipe.

### What this deliberately doesn't do

The original issue lists four possible fixes; this PR implements item 4 (first-class CDN support) since it's the cleanest long-term solution and it makes items 1–3 unnecessary as standalone fixes for the CDN scenario. Hash validation in the messages handler (item 3) could be added as a separate defensive change if maintainers want it.

## Test plan

- [x] `pnpm test:types` — clean
- [x] `pnpm test:unit` — 107 passed, 1 skipped (pre-existing)
- [x] `pnpm lint` — 0 errors (only pre-existing jsdoc warnings)
- [ ] Manually verify with a sample app:
  - [ ] `nuxt build` with `cdnURL` set produces `.output/public/_i18n/<hash>/<locale>/messages.json`
  - [ ] In-browser, the lazy-load request goes to `<cdnURL>/_i18n/...`
  - [ ] SSR request keeps using the relative URL
  - [ ] With `cdnURL` unset, behavior is unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for CDN-backed serving of lazy-loaded locale messages
  * Configure CDN URL to automatically prerender and serve locale messages from your CDN
  * Server-side rendering continues using local endpoints while client-side can fetch from CDN

### DIFF
--- a/docs/content/docs/04.api/00.options.md
+++ b/docs/content/docs/04.api/00.options.md
@@ -648,22 +648,17 @@ This feature relies on [Nuxt's Auto-imports](https://nuxt.com/docs/guide/concept
 
 Sets the prefix for the server route used for loading locale messages.
 
-## cdnURL
+## Serving messages from a CDN
 
-- type: `string`{lang="ts-type"}
-- default: `''`{lang="ts"}
+The module reuses Nuxt's built-in [`app.cdnURL`](https://nuxt.com/docs/4.x/api/nuxt-config#cdnurl) to serve lazy-loaded locale messages from a CDN / static origin instead of the dynamic Nitro server. There is no i18n-specific option — set Nuxt's `app.cdnURL` and the module wires the rest:
 
-Absolute URL prefix used by the **client** to fetch lazy-loaded locale messages from a CDN / static origin instead of the dynamic Nitro server.
-
-When set:
-
-- Client-side requests target `<cdnURL><serverRoutePrefix>/<hash>/<locale>/messages.json`.
-- The corresponding hashed messages JSON files are emitted into `.output/public/` at build time so they can be uploaded alongside `_nuxt/*` chunks.
-- Server-side rendering keeps fetching from the local Nitro handler — SSR does not round-trip through the CDN.
+- **Client-side** requests target `<app.cdnURL><serverRoutePrefix>/<hash>/<locale>/messages.json`.
+- The corresponding hashed `messages.json` files are emitted into `.output/public/` at build time so they can be uploaded alongside `_nuxt/*` chunks.
+- **Server-side rendering** keeps fetching from the local Nitro handler — SSR does not round-trip through the CDN.
 
 ```ts [nuxt.config.ts]
 export default defineNuxtConfig({
-  i18n: {
+  app: {
     cdnURL: 'https://cdn.example.com',
   },
 })
@@ -673,4 +668,4 @@ Recommended setup for blue/green deployments:
 
 1. Run `nuxt build` — hashed `messages.json` files are written into `.output/public/_i18n/<hash>/<locale>/messages.json` along with the rest of your static assets.
 2. Upload `.output/public/` to the same origin (e.g. S3) that serves your `_nuxt/*` chunks.
-3. Point your CDN (e.g. CloudFront) at that origin for the `/_i18n/*` path so each color of a deploy serves its own immutable, content-hashed messages — no sticky sessions required.
+3. Point your CDN (e.g. CloudFront) at that origin so each color of a deploy serves its own immutable, content-hashed messages — no sticky sessions required.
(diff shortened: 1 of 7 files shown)


Hronom:
I'm also confirming that I'm ok that maintainers modify PR as they want and if needed - close it and re-open own from scratch. My main interest here to get this thing done.

warflash:
I've been following the discussion you had on discord and just wanted to share me 2 cents in real quick, it feels a bit odd to me having a dedicated cdnUrl config flag for i18n when [nuxt itself already has one](https://nuxt.com/docs/4.x/api/nuxt-config#cdnurl). I do agree with the idea of wanting to store the output on a cdn, we do the same on our v9 apps, but defining the URL twice feels confusing.

Hronom:
> I've been following the discussion you had on discord and just wanted to share me 2 cents in real quick, it feels a bit odd to me having a dedicated cdnUrl config flag for i18n when [nuxt itself already has one](https://nuxt.com/docs/4.x/api/nuxt-config#cdnurl). I do agree with the idea of wanting to store the output on a cdn, we do the same on our v9 apps, but defining the URL twice feels confusing.

This is valid point. Asked Claude Code to re-use what Nuxt have, since this module specific for Nuxt and not supposed to be used in something else.

Hronom:
Verified — not applying. The duplicate concern doesn't manifest in practice:

- The route segment is a content-derived sha256 (`computeLocaleHashes` in `src/utils.ts`) that users can't reasonably know upfront. The whole point of this feature is that users *don't* compute the hash themselves to seed `nitro.prerender.routes`.
- The block already preserves user-configured routes — `config.prerender.routes ??= []` only initializes when missing, then `push(...)` appends. Nothing is overwritten.
- Even if a user duplicated the exact hashed path, Nitro deduplicates the prerender queue internally via its visited-URL set, so duplicate entries in the array are a no-op.

Adding a `Set`/filter dedupe would be defensive code for a scenario that can't happen, so I'd rather leave it as is.

BobbieGoede (approved):
Looks good, thank you for contributing!

Hronom:
@BobbieGoede , hey if possible please release=) Thank you a lot!

BobbieGoede:
@Hronom sorry for the delay, I'm planning on publishing a release in the next week.

I realized this change may cause unexpected behavior in some projects, since it will prerender message endpoints whenever `app.cdnURL` is set — regardless of whether the project uses static or dynamic message loaders. I'd like to confirm that's actually the case and address it before cutting a release.

rczdgbg:
@BobbieGoede 

When will it be released? Today is May 21st.

BobbieGoede:
Apologies again for the delay, some unexpected personal matters required my attention these past few weeks. I'll make this feature opt-in to address the concern mentioned in my previous comment (so that detection/improvements do not block the release any longer) and publish a release as soon as I can.

BobbieGoede:
I have published a [new release](https://github.com/nuxt-modules/i18n/releases/tag/v10.4.0) that includes these changes, note that I have modified it a bit in #3991 to prevent breaking projects that rely on dynamic/runtime message fetching by making it opt-in using `experimental.prerenderMessages` (this also allows projects that make no use of cdnURL to use this functionality)

Issue #3983, opened 2026-05-07
vueI18n config silently skipped in production builds; API docs don't state the path is resolved relative to i18nDir


rodion-m:
## Summary

Two related, narrow defects around the `i18n.vueI18n` option:

1. **Silent skip in production builds** — when the configured `vueI18n` file cannot be resolved, the warning that flags the misconfiguration is gated behind `import.meta.dev`. In `nuxt build` the file is silently ignored, no warning, no error, the production bundle just lacks `pluralRules` / overrides without any feedback.
2. **API option docs do not mention path resolution semantics** — the path is resolved relative to `i18nDir` (`restructureDir`), not to the project root or `nuxt.config.ts`, but the [`vueI18n` API option page](https://i18n.nuxtjs.org/docs/api/options#vuei18n) does not state this. New users naturally expect a path relative to `nuxt.config.ts`.

I'm not asking to change resolution semantics (that was discussed in #3399 / #3394 / #3395 and won't-fixed). I'm asking for **diagnostics + docs** that surface the existing semantics earlier.

## Reproduction

`@nuxtjs/i18n@10.2.3`, Nuxt `4.2.2`, Vue I18n `11.4.2`.

1. Place `i18n.config.ts` at the **project root** (next to `nuxt.config.ts`):
   ```ts
   export default defineI18nConfig(() => ({
     legacy: false,
     pluralRules: {
       ru: (n) => {
         if (n === 0) return 0
         const lt = n % 100, ld = n % 10, teen = lt >= 11 && lt <= 19
         if (!teen && ld === 1) return 1
         if (!teen && ld >= 2 && ld <= 4) return 2
         return 3
       }
     }
   }))
   ```
2. In `nuxt.config.ts`:
   ```ts
   i18n: {
     vueI18n: './i18n.config.ts',
     // ...
   }
   ```
3. Run `nuxt build`.

**Observed**:
- `nuxt build` exits 0 with no warnings, no errors.
- Inspecting `.nuxt/dist/server/client.manifest.mjs` and `.nuxt/dist/client/_nuxt/*.js`: the config file is **not** referenced anywhere; the `pluralRules` chunk is missing.
- Runtime translations silently fall back to default `pluralDefault` rule, breaking russian plurals in production while everything looked fine locally (because `nuxt dev` printed a warning that the developer may or may not have noticed).

[...]

PR #3992, opened 2026-05-21
fix: warn on missing vueI18n config in builds

tatakaisun:
## Summary
Warn whenever an explicitly configured `vueI18n` config file cannot be resolved, including production builds.

## Why
Issue #3983 reports that `nuxt build` silently skips a missing `vueI18n` config because the warning is gated behind `import.meta.dev`.

## Changes
- Remove the dev-only gate around the missing `vueI18n` config warning.
- Add a production-mode regression test for explicit missing config paths.

## Testing
- `vitest run test/layers.test.ts -c vitest.config.test.ts --mode production`
- `eslint --no-ignore test/layers.test.ts`
- `eslint src/layers.ts`
- `git diff --check`

## Known Issues
Docs for `vueI18n` path resolution are not changed in this PR and can be handled separately.

Refs #3983

## Summary by CodeRabbit

* **Bug Fixes**
  * Vue I18n configuration warnings now display consistently in both development and production environments.

* **Tests**
  * Added test coverage for Vue I18n configuration validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nuxt-modules/i18n/pull/3992?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


BobbieGoede (approved):
Looks good, the regression test feels a bit excessive for a simple warning so I removed it

PR #3993, opened 2026-05-21
fix: skip path config generation for unlocalized routes

### LINKED ISSUE #3989: `strategy: 'no_prefix'` in SPA: `<NuxtLink>` no-ops and direct URLs revert to `/` on 10.3.0 (regression vs 10.2.4)
### Environment
- `@nuxtjs/i18n`: **10.3.0** (works on 10.2.4)
- Nuxt: 4.4.6
- `ssr: false`
- `strategy: 'no_prefix'`, `defaultLocale: 'en'`

### Reproduction

Minimal config:
```ts
// nuxt.config.ts
  ssr: false,
  modules: ['@nuxtjs/i18n'],
  i18n: { defaultLocale: 'en', vueI18n: './config.ts' },
})

// i18n/config.ts
export default defineI18nConfig(() => ({
  legacy: false,
  locale: 'en',
  strategy: 'no_prefix',
}))
```

Add two pages, e.g. `pages/index.vue` and `pages/about.vue`, with a `<NuxtLink to=\"/about\">` between them.

### Actual

Clicking the `<NuxtLink>` does nothing. Typing `/about` directly in the address bar reverts to `/`. Identical symptom whether routes are reached via `<NuxtLink>` or programmatic `router.push`.

Pin back to `10.2.4` — works.

### Root cause

In `dist/module.mjs` 10.2.4, `pathToConfig` is populated as a side effect of the resolver, which is *only* called from inside `localizeRoutes`. For `strategy: 'no_prefix'` (without `differentDomains`), `shouldLocalizeRoutes` returns `false` and `localizeRoutes` exits early — so `ctx.pathToConfig` stays **empty**, and the generated `i18n-route-resources.mjs` virtual module's `pathToI18nConfig` is `{}`. The runtime treats every path as non-localized. ✅

In 10.3.0 (#3957) the resolver was split into a \"pure\" version, and population moved to a separate `buildPathToConfig(ctx, localeCodes, resolver, pages)` call in `setupPages`:

\`\`\`js
// dist/module.mjs 10.3.0 setupPages
const resolver = createPureOptionsResolver(ctx, options.defaultLocale, options.customRoutes)
const localizedPages = localizeRoutes(pages, { ...options, optionsResolver: resolver, ... })
buildPathToConfig(ctx, localeCodes, resolver, pages)  // <-- runs unconditionally, even if localizeRoutes was a no-op
\`\`\`

`buildPathToConfig` runs regardless of strategy. For `no_prefix`, it populates `ctx.pathToConfig[mappedPath][locale] = false` for every page × locale. The downstream invert loop and `pathToI18nConfig` template then ship a non-empty mapping that mis-routes navigation in the runtime — matching the symptom in #3987 (also a `pathToI18nConfig` lookup issue under different conditions).

### Suggested fix

Skip `buildPathToConfig` when `shouldLocalizeRoutes(options)` returns `false` — i.e. mirror the same gate that `localizeRoutes` uses internally:

\`\`\`js
const localizedPages = localizeRoutes(pages, { ...options, optionsResolver: resolver, ... })
if (shouldLocalizeRoutes({ ...options, locales: normalizedLocales })) {
  buildPathToConfig(ctx, localeCodes, resolver, pages)
}
\`\`\`

This restores 10.2.4 semantics for `no_prefix` while keeping the new split for all other strategies. Alternatively, gate inside `buildPathToConfig` itself.

### Additional context

Symptom is identical to #3971 in shape (silent regression in `pages.ts` rewrite from #3957 under specific strategies) and the runtime data structure misuse aligns with #3987's analysis. Issues likely share a fix.

tatakaisun:
## Summary
Fixes #3989.

## Why
For `strategy: 'no_prefix'` without domain-based localization, routes are not localized, but `buildPathToConfig` still populated route resource mappings. This caused `pathToI18nConfig` to be non-empty for unlocalized SPA routes.

## Changes
- Reuse the same `shouldLocalizeRoutes` gate before building path config.
- Add a regression test for `setupPages` route resources under `no_prefix`.

## Testing
- pnpm vitest run test/pages/setup_pages.test.ts test/pages/route_localization.test.ts -c vitest.config.test.ts
- git diff --check

## Known Issues
- `pnpm test:types` currently fails in `src/runtime/shared/domain.ts(32,89)`, outside this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed route localization to conditionally build routes based on configuration, preventing unnecessary route generation in certain scenarios.

* **Tests**
  * Added comprehensive test coverage for page setup and localization behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nuxt-modules/i18n/pull/3993?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/pages.ts
+++ b/src/pages.ts
@@ -5,7 +5,7 @@ import { parse as parseSFC } from '@vue/compiler-sfc'
 import { parseAndWalk } from 'oxc-walker'
 import { mkdir, readFile, writeFile } from 'node:fs/promises'
 import { parseSegment, toVueRouterSegment } from 'unrouting'
-import { localizeRoutes } from './routing'
+import { localizeRoutes, shouldLocalizeRoutes } from './routing'
 import { dirname, parse as parsePath, resolve } from 'pathe'
 import { createRoutesContext, resolveOptions } from 'vue-router/unplugin'
 import { transform } from './transform/resource'
@@ -104,16 +104,20 @@ export const disabledI18nPathToPath = ${JSON.stringify(routeResources.disabledI1
 
       const resolver = createPureOptionsResolver(ctx, options.defaultLocale, options.customRoutes)
 
-      const localizedPages = localizeRoutes(pages as NarrowedNuxtPage[], {
+      const localizationOptions = {
         ...options,
         includeUnprefixedFallback,
         locales: normalizedLocales,
         optionsResolver: resolver,
         compactRoutes: !!options.experimental?.compactRoutes,
-      })
+      }
+
+      const localizedPages = localizeRoutes(pages as NarrowedNuxtPage[], localizationOptions)
 
       // Build path config from original pages (not localized copies)
-      buildPathToConfig(ctx, localeCodes, resolver, pages as LocalizableRoute[])
+      if (shouldLocalizeRoutes(localizationOptions)) {
+        buildPathToConfig(ctx, localeCodes, resolver, pages as LocalizableRoute[])
+      }
 
       // keep root when using prefixed routing without prerendering
       const indexPage = pages.find(x => x.path === '/')
--- a/src/routing.ts
+++ b/src/routing.ts
@@ -19,7 +19,7 @@ function createShouldPrefix(opts: SetupLocalizeRoutesOptions, ctx: RouteContext)
   }
 }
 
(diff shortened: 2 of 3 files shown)


BobbieGoede (approved):
Looks good, thanks for contributing!

Issue #3999, opened 2026-06-01
SSR locale redirect computes a malformed path when app.baseURL is set (triggers "Matcher cannot resolve relative paths" warning)


imlautaro:
### Environment

- `@nuxtjs/i18n`: 10.4.0
- `nuxt`: 4.4.x (vue-router 5.0.7)
- i18n config: `strategy: 'prefix'`, plus `app.baseURL` set to a sub-path (e.g. `/club`)

### Reproduction

Minimal config:

```ts
export default defineNuxtConfig({
  app: { baseURL: '/club' },
  modules: ['@nuxtjs/i18n'],
  i18n: { strategy: 'prefix', defaultLocale: 'en', locales: ['en', 'es'] }
})
```

SSR-render any localized route, e.g. `/club/en/calendar`. The dev console logs:

```
[Vue Router warn]: The Matcher cannot resolve relative paths but received "ub/en/calendar". Unless you directly called matcher.resolve("ub/en/calendar"), this is probably a bug in vue-router.
```

(Everything still works — the warning is cosmetic — but it's noisy and misleadingly points at vue-router.)

### Root cause

In `src/runtime/server/plugin.ts`, the `render:before` hook:

```js
const localeSegment = detector.route(event.path)   // event.path is base-stripped -> "/en/calendar" -> "en"
const pathLocale   = (isSupportedLocale(localeSegment) && localeSegment) || undefined
const path = (pathLocale && url.pathname.slice(localeSegment.length + 1)) ?? url.pathname
```

The locale segment is detected from `event.path`, which Nitro has **already stripped** of `app.baseURL`. But the slice runs against `url.pathname`, which **still contains** the base. With `baseURL: '/club'`:

```js
'/club/en/calendar'.slice('en'.length + 1) // slice(3) -> "ub/en/calendar"
```

That relative string is passed to `isExistingNuxtRoute(path)` -> `matcher.resolve({ path })`, and vue-router warns because the matcher only accepts absolute paths. It then finds no match and the redirect bails — so it's harmless, but the path is wrong.

### Suggested fix

Strip the base from `url.pathname` (or derive the pathname from the same base-stripped source used for locale detection) before slicing, so `path` stays absolute and base-free. Happy to open a PR.

imlautaro:
Duplicate of #3887. Fix opened in #4000.

PR #4000, opened 2026-06-01
fix: strip base URL from path in SSR locale redirect

### LINKED ISSUE #3887: `[Vue Router warn]: The Matcher cannot resolve relative paths` when `baseURL` and strategy: "prefix" is set
### Environment

- Operating System: Darwin
- Node Version:     v22.21.1
- Nuxt Version:     4.2.1
- CLI Version:      3.30.0
- Nitro Version:    2.12.9
- Package Manager:  pnpm@9.15.9
- Builder:          -
- User Config:      compatibilityDate, devtools, modules, i18n, typescript, app, experimental
- Runtime Modules:  @nuxtjs/i18n@10.2.1
- Build Modules:    -

### Reproduction

[reproduction](https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-y8qrpwcd?file=nuxt.config.ts)

set nuxt.config.ts
```
app: {
  baseURL: '/12345'
}
```
and
```
i18n:{
  strategy: 'prefix'
}
```

1. Open stackblitz reproduction page
2. Open the page in new tab
3. Reload the page.

console shows ` WARN  [Vue Router warn]: The Matcher cannot resolve relative paths but received "345/en". Unless you directly called matcher.resolve("345/en"), this is probably a bug in vue-router. Please open an issue at https://github.com/vuejs/router/issues/new/choose.`

### Describe the bug

I believe the root cause of this bug is that app.baseURL (defined in nuxt.config.ts) is not taken into account when calculating the path.

https://github.com/nuxt-modules/i18n/blob/15bb7bbe95500d99f6ef1a742b0b1e300f0f7d8a/src/runtime/server/plugin.ts#L158

Proposed Fix: I suggest modifying the code as follows:

```ts
const path = (pathLocale && event.path.slice(pathLocale.length + 1)) ?? event.path
```

**sorry, I haven't checked for potential side effects.**

### Additional context

_No response_

### Logs

```shell

```

imlautaro:
### 🔗 Linked issue

Resolves #3887 (duplicate: #3999)

### ❓ Type of change

- [x] 🐞 Bug fix (non-breaking change which fixes an issue)

### 📚 Description

When `app.baseURL` is set together with `strategy: 'prefix'`, the SSR locale-redirect middleware logs:

```
[Vue Router warn]: The Matcher cannot resolve relative paths but received "ub/en/calendar". ...
```

**Root cause** — in `src/runtime/server/plugin.ts` (`render:before` hook), the route `path` was derived from `url.pathname`, which still contains `app.baseURL`, while the locale was detected from `event.path`, which Nitro has **already stripped** of the base:

```js
const localeSegment = detector.route(event.path)            // "/en/calendar" -> "en"
const path = (pathLocale && url.pathname.slice(pathLocale.length + 1)) ?? url.pathname
// '/club/en/calendar'.slice(3) -> "ub/en/calendar"   (relative!)
```

That relative string reaches `isExistingNuxtRoute(path)` -> `matcher.resolve({ path })`, which vue-router rejects (it only accepts absolute paths). The match then fails and the redirect silently bails — so an unprefixed path under a base is never localized.

**Fix** — derive the path from `event.path` (already base-stripped) and run it through `parsePath` to drop any query string, keeping `path` an absolute, base-free route path. This also resolves the existing TODO in `src/runtime/shared/matching.ts`:

```js
// TODO: path should not have base url - check if base url is stripped earlier
```

### 🧪 Test

Added `specs/issues/3887.spec.ts`: with `app.baseURL: '/base-path'` + `strategy: 'prefix'`, an unprefixed path now redirects to its localized route.

- **With the fix:** `GET /base-path/about` -> `302` (`location` contains `/about`) ✅
- **Without the fix:** the same request returns `404` (redirect bailed) — so the test fails on the unpatched code, confirming it's a real regression test.

> Note: I couldn't run the browser-based e2e specs locally (Playwright browser download was blocked in my environment), so I'm relying on CI for that matrix. The added spec is request-based and runs without a browser.

### Checklist

- [x] Lint passes on the changed file
- [x] Added a regression test

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server-side locale redirect handling to correctly account for `app.baseURL` path prefixes, so unprefixed and root requests redirect to the right locale-prefixed URLs while already-prefixed paths do not change.
  * Ensured redirect URLs include the base path exactly once when working with locale domains.

* **Tests**
  * Added end-to-end Vitest coverage for Nuxt SSR locale redirects with `app.baseURL` set to a base path, including root, non-prefixed, and already-prefixed request scenarios.

### DIFF
--- a/specs/different_domains/base_path_ssr_redirect.spec.ts
+++ b/specs/different_domains/base_path_ssr_redirect.spec.ts
@@ -1,0 +1,29 @@
+import { test, expect } from 'vitest'
+import { fileURLToPath } from 'node:url'
+import { setup, undiciRequest } from '../utils'
+
+await setup({
+  rootDir: fileURLToPath(new URL(`../fixtures/different_domains`, import.meta.url)),
+  // overrides
+  nuxtConfig: {
+    app: {
+      baseURL: '/base-path'
+    },
+    i18n: {
+      defaultLocale: 'en',
+      rootRedirect: 'about',
+      detectBrowserLanguage: false
+    }
+  }
+})
+
+// the SSR redirect location joins the locale domain and `app.baseURL`, the base
+// should appear exactly once (domain and `baseUrl` values are configured without it)
+test('(#3887) SSR root redirect includes `app.baseURL` once after the locale domain', async () => {
+  const res = await undiciRequest('/base-path/', {
+    headers: { 'X-Forwarded-Host': 'en.nuxt-app.localhost' }
+  })
+
+  expect(res.statusCode).toBe(302)
+  expect(res.headers.location).toEqual('http://en.nuxt-app.localhost/base-path/about')
+})
--- a/specs/issues/3887.spec.ts
+++ b/specs/issues/3887.spec.ts
@@ -28,6 +28,19 @@ describe('#3887', async () => {
     const res = await fetch('/base-path/about', { redirect: 'manual' })
 
     expect(res.status).toBe(302)
-    expect(res.headers.get('location')).toContain('/about')
+    expect(res.headers.get('location')).toEqual('/base-path/en/about')
+  })
+
+  test('redirects the root path to the localized root when `app.baseURL` is set', async () => {
+    const res = await fetch('/base-path/', { redirect: 'manual' })
+
+    expect(res.status).toBe(302)
+    expect(res.headers.get('location')).toEqual('/base-path/en')
+  })
+
+  test('does not redirect a prefixed path', async () => {
+    const res = await fetch('/base-path/en/about', { redirect: 'manual' })
+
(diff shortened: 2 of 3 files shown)


BobbieGoede:
Thanks for the PR 🙏 While testing this against a reproduction I found the redirect was still broken with a base set: the root checks in `resolveRedirectPath` compared `requestURL.pathname` (which contains the base), so root requests were never redirected with `strategy: 'prefix'`, and the redirect location was built without the base (`/en` instead of `/base-path/en`).

I updated your branch to cover these cases as well, rebased it on main and added tests.

Issue #3558, opened 2025-04-17
Type 'true' is not assignable to type 'false' when using `<NuxtLinkLocale :to="..." custom>`


mrleblanc101:
### Environment

------------------------------
- Operating System: Darwin
- Node Version:     v20.18.0
- Nuxt Version:     3.16.2
- CLI Version:      3.24.1
- Nitro Version:    2.11.9
- Package Manager:  npm@10.8.2
- Builder:          -
- User Config:      modules, ssr, devtools, app, css, colorMode, future, compatibilityDate, vite, hooks, eslint, i18n, icon, pinia, tailwindcss
- Runtime Modules:  floating-vue/nuxt@5.2.2, @nuxtjs/color-mode@3.5.2, @vueuse/nuxt@13.1.0, @pinia/nuxt@0.11.0, @nuxt/icon@1.12.0, @nuxt/fonts@0.11.1, @nuxtjs/tailwindcss@6.13.2, @nuxtjs/i18n@9.5.3, @regle/nuxt@1.0.9
- Build Modules:    -
------------------------------

### Reproduction

```js
<li v-for="(step, index) in steps" :key="step.label" class="relative -mx-1 flex grow">
    <NuxtLinkLocale #default="{ isActive, href, navigate }" :to="step.to" custom>
        <Component
            :is="!disabled ? 'a' : 'div'"
            :href="href"
            @click="!disabled && navigate"
        >
            ...
        </Component>
    </NuxtLinkLocale>
</li>
```

### Describe the bug

Not sure if this is a regression in Nuxt-i18n or Nuxt in general.
But when I use `<NuxtLinkLocale :to="..." custom>` in get this error:

> Type 'true' is not assignable to type 'false'.ts-plugin(2322)

<img width="883" alt="Image" src="https://github.com/user-attachments/assets/8c9cb99c-46fd-4c79-92e0-8500546db02b" />

<img width="892" alt="Image" src="https://github.com/user-attachments/assets/41e60611-0564-4c49-861d-e4979a4c4e22" />

Here are additional information for the TS tooltip in VS Code:

[...]

PR #4007, opened 2026-06-15
fix(NuxtLinkLocale): make component generic to support `custom` prop …

houtanrocky:
### 🔗 Linked issue
#3558 

### 📚 Description

Convert component definition from non‑generic `defineComponent<NuxtLinkLocaleProps<boolean>>` to a generic function pattern. This allows the `custom` prop to accept `true` and correctly infers slot props, resolving the TypeScript error `Type 'true' is not assignable to type 'false'`.

## Summary by CodeRabbit

* **Refactor**
  * Strengthened the locale link component’s handling of internal vs. external destinations, with improved TypeScript typing and safer prop preparation to reduce edge-case issues.
* **Bug Fixes**
  * Ensured locale-aware links render the expected localized URL and navigate correctly when using custom link props/slot content.
* **Tests**
  * Added coverage verifying the localized `href` output and that clicking the link routes to the correct localized page.

### DIFF
--- a/specs/basic_usage.spec.ts
+++ b/specs/basic_usage.spec.ts
@@ -783,5 +783,16 @@ describe('basic usage', async () => {
 
       expect(await page.locator('#per-component-hello').innerText()).toMatch('Hello!')
     })
+    test('NuxtLinkLocale with custom prop works', async () => {
+      const { page } = await renderPage('/en')
+      // Check that the custom link renders with the correct href
+      const href = await page.getAttribute('#custom-link a', 'href')
+      expect(href).toBe('/en/about')
+      
+      // Optional: click and verify navigation
+      await page.locator('#custom-link a').click()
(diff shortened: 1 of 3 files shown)


BobbieGoede:
Thank you for your PR! The `custom` prop and slot types should already be fixed on `main` by #3995, that was merged just after v10.4.0 was tagged so it hasn't been published in a release yet. The exported component is already typed as a generic constructor, so from what I can tell the rewrite doesn't change anything functionally, could you check if you're still seeing the error from #3558 with the edge release by running `npm i @nuxtjs/i18n@npm:@nuxtjs/i18n-edge`? 🙏

Note that the added test checks runtime behavior, which was never broken — #3558 is a type-level error, so guarding it against regressions would need a type-level test (#3994 shows one way to do this). Closing this for now, but will reopen if you can still reproduce the issue with the edge release.

PR #4012, opened 2026-06-22
fix: resolve `bundle.onlyLocales` from the first layer that specifies it

cernymatej:
### 📚 Description

previously, `filterLocales` only read `bundle.onlyLocales` from the running project (`_layers[0]`) and ignored it entirely when a downstream layer set it

this pr makes it walk layers starting from the running project and use the first one that specifies it, which IMO is the expected behaviour

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `onlyLocales` handling by deriving the effective locale filter from the first configuration layer that specifies it, with behavior unchanged when no filtering is required.
* **Tests**
  * Added new test coverage for locale selection across multiple layer configurations, including fallback and the “explicit empty” edge case.
* **Documentation**
  * Updated `onlyLocales` option docs to clarify how Nuxt Layers merging works (non-merged resolution from the first defining layer, with project config taking precedence).


BobbieGoede (approved):
Looks good, thanks!

PR #4014, opened 2026-06-28
fix(compactRoutes): key NuxtPage by interpolated pattern, not full path

### LINKED ISSUE #4001: `compactRoutes`: `<NuxtPage>` key is the full path, remounting nested pages on every child navigation (non-default locales only)
### Environment

- @nuxtjs/i18n: 10.4.0
- strategy: prefix_except_default
- experimental.compactRoutes: true
- Nuxt: 4.4.6
- Nitro: 2.13.4
- Vue: 3.5.x (via Nuxt 4.4.6)
- Node: v24.15.0
- Builder: vite 7.3.3
- Package manager: pnpm 11.1.1
- OS: macOS 25.5.0

### Reproduction

No StackBlitz — minimal textual reproduction (cause and fix are known, see below):

1. `pages/foo/index.vue` containing `<NuxtPage />`, with children `pages/foo/bar.vue` and `pages/foo/baz.vue`.
2. In `foo/index.vue`: `onBeforeUnmount(() => console.log('parent unmounted'))`.
3. Config: `strategy: 'prefix_except_default'`, `experimental: { compactRoutes: true }`.
4. Navigate `/de/foo/bar` → `/de/foo/baz` (any non-default locale).

Expected: `foo/index.vue` stays mounted across the child navigation; it should only remount when the locale param actually changes (`/de/foo/bar` → `/fr/foo/bar`).
Actual: `foo/index.vue` remounts on the child navigation (`onBeforeUnmount` fires, parent state is lost). On the default locale (`/foo/bar` → `/foo/baz`) it does NOT remount.

Happy to provide a StackBlitz if needed, but the repro is deterministic on any nested-pages + non-default-locale setup.

### Describe the bug

With `compactRoutes`, the i18n plugin sets `route.meta.key = (r) => r.path` on compact pages (`runtime/plugins/route-locale-detect.js`). Nuxt uses `meta.key` as the `<NuxtPage>` component key, so keying by the full resolved path makes the key change on every child navigation — Vue then treats the parent record as a different instance and remounts the whole subtree. Any state held in a parent component across child navigation is wiped, and lifecycle hooks (`onBeforeMount` / `onBeforeUnmount`) re-run unexpectedly.

The default-locale tree isn't compact (no `meta.key`), so Nuxt falls back to the record's own interpolated pattern, which is stable across child navigation — which is why this only reproduces on non-default locales.

The upstream intent (remount when the locale param changes, e.g. `/de/x` → `/fr/x`) is legitimate, but keying by the full path over-fires on every child navigation. Introduced in #3957; still present in 10.4.0.

### Additional context

Suggested fix — use the record's own interpolated pattern instead of the full path (matches Nuxt's default `<NuxtPage>` keying, while preserving the locale-switch remount):

```js
const pattern = route.path.replace(/(:\w+)\([^)]+\)/g, "$1").replace(/(:\w+)[?+*]/g, "$1");
route.meta.key = (r) => pattern.replace(/:\w+/g, (m) => r.params[m.slice(1)]?.toString() || "");
```

Verified locally on 10.4.0: child navigation no longer remounts the parent; locale switch (`/de/x` → `/fr/x`) still remounts as intended.

### Logs

```shell

```

Togetic:
### 🔗 Linked issue

Fixes #4001

### 📚 Description

With `experimental.compactRoutes`, the `route-locale-detect` plugin keys compact routes by the **full resolved path**:

```ts
route.meta.key = (r) => r.path
```

Nuxt uses `meta.key` as the `<NuxtPage>` component key, so the key changes on **every child navigation** (e.g. `/de/foo/bar` → `/de/foo/baz`). Vue then treats the parent record as a different instance and **remounts the whole nested subtree** — state held in a parent component across child navigation is wiped, and `onBeforeMount` / `onBeforeUnmount` re-run unexpectedly.

This only reproduces on **non-default locales**: the default-locale tree isn't compact (no `meta.key`), so Nuxt falls back to the record's own interpolated pattern, which is stable across child navigation.

The upstream intent — remount when the locale param actually changes, e.g. `/de/x` → `/fr/x` (introduced in #3957) — is legitimate, but keying by the full path over-fires on every child navigation.

**Fix:** key by the record's own interpolated pattern instead, matching Nuxt's default `<NuxtPage>` keying:

```ts
const pattern = route.path.replace(/(:\w+)\([^)]+\)/g, '$1').replace(/(:\w+)[?+*]/g, '$1')
route.meta.key = (r) => pattern.replace(/:\w+/g, m => r.params[m.slice(1)]?.toString() || '')
```

For a compact record `/:locale(en|de|fr)/foo`, the key resolves to `/de/foo` for any child of `foo` under `/de` — stable across `/de/foo/bar` ↔ `/de/foo/baz`, while still changing to `/fr/foo` on a locale switch, so the intended locale-change remount is preserved.

### ✅ Verification

- `pnpm test:types`, `pnpm lint`, and `pnpm test:unit` all pass.
- Verified manually on a nested-pages app: child navigation no longer remounts the parent; a locale switch (`/de/x` → `/fr/x`) still remounts as intended.

### 📝 Note on tests

`compactRoutes` is currently covered only at the route-table level (`test/pages/route_localization.test.ts`); there's no e2e fixture exercising compact-route runtime navigation, and `meta.key` is assigned inside the runtime plugin. I'm happy to add a dedicated fixture + spec (nested pages asserting the parent stays mounted across child navigation and remounts on a locale switch) if you'd like that folded into this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved locale-aware navigation so nested pages stay mounted when moving between child routes.
  * Routes now refresh only when the locale-related route value actually changes, reducing unnecessary remounts and page resets.

### DIFF
--- a/src/runtime/plugins/route-locale-detect.ts
+++ b/src/runtime/plugins/route-locale-detect.ts
@@ -16,7 +16,12 @@ export default defineNuxtPlugin({
       const router = useRouter()
       for (const route of router.getRoutes()) {
         if (route.meta?.__i18nCompact && route.meta.key == null) {
-          route.meta.key = (r: { path: string }) => r.path
+          // Key by the record's own interpolated pattern (mirrors Nuxt's default `<NuxtPage>`
+          // keying) instead of the full resolved path, so nested parent pages stay mounted
+          // across child navigation and only remount when the locale param actually changes.
+          const pattern = route.path.replace(/(:\w+)\([^)]+\)/g, '$1').replace(/(:\w+)[?+*]/g, '$1')
+          route.meta.key = (r: { params: Record<string, string | string[]> }) =>
+            pattern.replace(/:\w+/g, m => r.params[m.slice(1)]?.toString() || '')
         }
       }
     }


BobbieGoede:
Thanks for the PR and the detailed analysis 🙏 While looking into this I found we likely don't need to set `meta.key` at all, Nuxt's default keying already interpolates the matched record's own pattern, so it stays stable across child navigation and still changes when the locale param does. I opened #4025 which removes the assignment instead and adds tests for both behaviors, so I'll close this one in favor of that.

Issue #3019, opened 2024-07-08
Support for `defineRouteRules`


warflash:
### Describe the feature

Nuxt added support for [inline route rules](https://nuxt.com/docs/api/utils/define-route-rules), when using locale prefixed routes however - the rules don't seem to match/be triggered.
More broadly speaking a section or note on how to deal with nuxt `routeRules` and prefixed routes might be nice, maybe even a helper function to pass config to and get a transformed result back

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution) (The contribution guideline of nuxt-modules/i18n is compliant with Nuxt too).
- [X] Check existing [discussions](https://github.com/nuxt-modules/i18n/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

BobbieGoede:
Depending on the order of execution for the pages hooks this will probably involve reimplementing this functionality inside this module 😥 Hopefully this won't be as complicated as adding support for `definePageMeta` was, but it's implementation seems similar to it (before `experimental.scanPageMeta` was added).

Xenossolitarius:
This would be amazing man tbh. Unfortunately not a part where I can contribute

BobbieGoede:
This will be resolved upstream if this gets merged https://github.com/nuxt/nuxt/pull/32897

PR #4063, opened 2026-07-22
fix: expand inline route rules on compact routes per locale

BobbieGoede:
* Related https://github.com/nuxt/nuxt/pull/35455
* Related #3019

Proof of concept: this expands inline route rules on compact routes into separate per-locale rules (`/en/account`, `/nl/account`) instead of nuxt mapping these to `/**`. We consume all page rules before nuxt's collection and duplicate nuxt's `globRouteRulesFromPages`/`pathToNitroGlob` to do so, I suspect this approach might conflict with nuxt's inline route rules handling and needs a closer look.

## Summary by CodeRabbit

- **New Features**
  - Added support for inline route rules on localized and compact routes.
  - Route-specific settings are now automatically applied across expanded locale paths, including nested and dynamic routes.
  - Existing route configuration is preserved and merged with inline rules.

- **Bug Fixes**
  - Improved handling of localized route patterns so their associated settings apply consistently to all generated paths.

### DIFF
--- a/src/pages.ts
+++ b/src/pages.ts
@@ -1,4 +1,5 @@
 import { addTemplate, updateTemplates } from '@nuxt/kit'
+import { defu } from 'defu'
 import { readFileSync } from 'node:fs'
 import { isString } from '@intlify/shared'
 import { parse as parseSFC } from '@vue/compiler-sfc'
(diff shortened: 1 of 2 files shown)


BobbieGoede:
Closing, this has been resolved upstream in nuxt and unrouting with https://github.com/nuxt/nuxt/pull/35455

PR #4064, opened 2026-07-23
fix: strip protocol when matching domain to host

adamkasper:
### 🔗 Linked issue

--

### 📚 Description

Since #4042 (v10.5.0), `differentDomains` + `prefix_except_default` breaks when `locale.domain` is configured with a protocol (e.g. `https://www.example.com`) and `domainDefault: true`: `getDefaultLocaleForDomain()` compares `defaultForDomains` entries verbatim against the request host, so they never match, the wrong locale's routes get unprefixed and unprefixed paths redirect to the prefixed variant (`/jobs` → `/en/jobs`).

Fix: reuse the protocol-stripping already done in `isLocaleOnHost()` for all remaining configured-domain vs. host comparisons (`getDefaultLocaleForDomain`, `matchDomainLocale`, `domainFromLocale`, and the build-time duplicate-domain check in `shouldLocalizeRoutes`). `ufo`'s `withoutProtocol` is deliberately not used — it treats `host:port` as a protocol scheme.

Tests: unit coverage for the normalization and all touched functions, plus an e2e spec reproducing the reported setup (fails without this fix, passes with it). Existing host-only `domain` tests are unaffected.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved locale detection for domains configured with `http://` or `https://`.
  - Default locales are now selected correctly when domain protocols differ between configuration and incoming requests.
  - Domain-based routing now consistently resolves locales across multiple domains.
  - Prevented duplicate domain configurations from incorrectly enabling localized routes, including protocol-only variations.
  - Fixed domain-specific default routes and redirects for prefixed and unprefixed URLs.


BobbieGoede (approved):
Thank you for contributing!

PR #4065, opened 2026-07-23
fix(hmr): use virtual module ids in `import.meta.hot.accept()` for vite 8

sawa-ko:
### 📚 Description

Vite 8 [removed the `import.meta.hot.accept()` resolution fallback](https://github.com/vitejs/vite/pull/21382). Previously, `accept()` could resolve filesystem-relative paths to the correct module in the module graph. In Vite 8, `accept()` requires the **exact module ID** as it appears in the import graph.

The i18n module imports locale files and vue-i18n configs through virtual module IDs (`#nuxt-i18n/<hash>`) via the `ResourcePlugin`, but the HMR `accept()` calls in `genLocaleLoaderHMR` and `genVueI18nConfigHMR` were using filesystem-relative paths (e.g. `../i18n/locales/en-us.json`). This mismatch caused `accept()` to silently fail, making Vite fall back to sending `{ type: "full-reload" }` via WebSocket — which results in a navigation request that gets stuck in a pending state.

### Root Cause

In `src/gen.ts`, the `generateLoaderOptions` function creates loader objects with a `relative` property using `relative(nuxt.options.buildDir, meta.path)`, but the actual `import()` uses `asI18nVirtual(meta.hash)`:

```ts
// The dynamic import uses the virtual ID
load: genDynamicImport(asI18nVirtual(meta.hash), ...)

// But HMR accept was using the filesystem-relative path
relative: relative(nuxt.options.buildDir, meta.path)
```

Then in `src/template.ts`, the generated HMR code used the wrong identifier:

```ts
// ❌ Before — filesystem path that Vite 8 can't resolve
import.meta.hot.accept("${loader.relative}", async mod => { ... })

// ✅ After — virtual module ID matching the import graph
import.meta.hot.accept("${loader.virtualId}", async mod => { ... })
```

### Changes

- **`src/gen.ts`**: Added `virtualId` property to `LocaleLoaderData` type and populated it with `asI18nVirtual(hash)` in both locale loaders and vue-i18n config loaders.
- **`src/template.ts`**: Changed `import.meta.hot.accept()` calls to use `virtualId` instead of `relative` in both `genLocaleLoaderHMR` and `genVueI18nConfigHMR`.

### How to Reproduce

1. Use Nuxt 4.5+ (ships with Vite 8)
2. Configure `@nuxtjs/i18n` with JSON locale files
3. Start dev server and navigate to any page
4. Edit a translation JSON file
5. **Expected**: Messages update in-place via HMR
6. **Actual**: A new navigation request appears in the Network tab (e.g. `http://localhost:3000/current-page`) that stays in a pending state indefinitely

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hot module replacement behavior for locale files and Vue I18n configuration updates by using stable generated identifiers for the hot-update wiring.
  * Ensured generated loader and configuration metadata no longer relies on relative paths, improving consistency of dynamic import identifiers.
* **Tests**
  * Updated generator tests to simplify Nuxt options setup and align expectations with the new loader option generation behavior.


BobbieGoede:
Thanks you for your contribution! We will need to check if HMR will still work for webpack/rspack with this change.

sawa-ko:
Thanks @BobbieGoede!

I just pushed an update that fixes the failed unit tests by updating the snapshots in `test/gen.test.ts.snap`.

Regarding **webpack/rspack** support:
The dynamic imports generated for locale loaders (`genDynamicImport`) already use the virtual module path
  (`#nuxt-i18n/<hash>`) across all bundlers.

Since `ResourcePlugin` resolves both the dynamic `import()` and `import.meta.hot.accept()` requests, using the
  exact same virtual module ID (`virtualId`) ensures that the HMR module graph stays aligned regardless of whether
  Vite, Webpack, or Rspack is used.

Let me know if you'd like me to add an end-to-end fixture or additional test cases for bundler validation!

BobbieGoede (approved):
I cleaned up the now unused logic and tested with vite 7 and 8 (I misremembered, we don't support HMR for webpack/rspack), LGTM!

PR #3043, opened 2024-07-26
fix: encode `switchLocalePath` during SSR replacement

BobbieGoede:
### 🔗 Linked issue
* https://github.com/nuxt-modules/i18n/issues/3042

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description
This should resolve #3042, but if I'm overlooking something please let me know.

@kazupon 
I created a [`v8` branch](https://github.com/nuxt-modules/i18n/tree/v8) which diverts from `main` at 7930a2773311bef05bd98a3a98540ce17b6a7d6b, I have also cherry picked some changes from the main branch where possible.

If this PR gets merged it will need to be added to the `v8` branch as well, will the normal release flow work from that branch? 🤔

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have added tests (if possible).
- [ ] I have updated the documentation accordingly.

Issue #3042, opened 2024-07-26
Experimental switchLocalePathLinkSSR causing possible XSS vector


KilianSSL:
### Environment

- Operating System: Darwin
- Node Version:     v20.11.1
- Nuxt Version:     3.10.1
- CLI Version:      3.12.0
- Nitro Version:    2.9.6
- Package Manager:  npm@10.2.4
- Builder:          -
- User Config:      site, debug, app, ssr, telemetry, devtools, sourcemap, modules, vite, nitro, runtimeConfig, i18n, hooks, graphqlMiddleware, multiCache, sitemap, robots, svgSprite, alias, purgecss, htmlValidator
- Runtime Modules:  @nuxtjs/html-validator@1.8.2, @nuxtjs/i18n@8.3.1, @nuxtjs/sitemap@5.2.0, @nuxtjs/robots@3.0.0, @nuxtjs/svg-sprite@1.0.2, @pinia/nuxt@0.5.1, nuxt-graphql-middleware@4.0.0, nuxt-multi-cache@3.1.1, nuxt-purgecss@2.0.0, @nuxt/test-utils/module@3.13.1
- Build Modules:    -

### Reproduction

Minimal Repro: https://github.com/KilianSSL/i18n-ssr-repro

Disable javascript, to best review the SSR response.

1. Set experimental switchLocalePathLinkSSR flag to true
2. Create a wildcard page like /pages/[...slug].vue
3. Make sure you have a language switcher utilizing `<SwitchLocalePathLink>`
4. Navigate to a page like: `http://localhost:3000/fr/%22%3Exss/`
5. If not manually overridden/sanitized via 

  ```
 setI18nParams({
    fr: { slug: '' },
    de: { slug: '' },
  })  
```

  the `">xss/` portion of the URL will cause `<SwitchLocalePathLink>` to output invalid HTML that will render whatever comes after `">xss/`.

### Describe the bug

The bug seems to be caused by `<SwitchLocalePathLink>` defaulting to the current URL parameters, if not set otherwise via `setI18nParams`. The parameters are not being sanitized before they're used in the generated HTML.

![grafik](https://github.com/user-attachments/assets/5e2046c7-202f-4467-8887-971c31b32023)

BobbieGoede:
This is problematic, thanks for reporting! I'll see if I can get this fixed soon and publish a patch release with it.

PR #4079, opened 2026-07-27
fix: let vue-router encode `switchLocalePath` paths

### LINKED ISSUE #2909: URL with URL-encoded characters like %23 gets decoded when changing locale
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.11.2
- CLI Version:      3.11.1
- Nitro Version:    2.9.6
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      devtools, modules, i18n, ssr
- Runtime Modules:  @nuxtjs/i18n@8.3.0
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/i18n-url-encoding-issue?file=app.vue

### Describe the bug

If an URL contains an URL-encoded character like "`%23`" and we switch the language, the url gets decoded and converted into a "`#`" which breaks for example the `$route.params` usage.

1. Open Reproduction
2. Click "Go to my-blog-post%23number1-someid" link
3. Change locale
4. Observe $route.params

Before locale change the route params are: (which vue-router already decodes correctly)
`{
  "slug": "my-blog-post#number1-someid"
}`

after locale change:
`{
  "slug": "my-blog-post"
}`

and we get a route hash
`#number1-someid`

You can also open the preview in a new tab to see the decoding in the URL bar.

The only other issue I found which could be maybe related to this is: https://github.com/nuxt-modules/i18n/issues/2698 

### Additional context

_No response_

### Logs

_No response_

Vincentdevreede:
* Resolves #2909 

## Summary

 `SwitchLocalePathLink` and the SSR `switch-locale-path-ssr` plugin both wrapped `switchLocalePath()`'s output in a plain `encodeURI(...)` call before rendering it as an `href`. For a locale with a custom route path containing non-ASCII characters (for example a Japanese `pages` override like `ja: '/約'`), `switchLocalePath()` already returns the path pre-encoded, because Nuxt's route compiler encodes non-ASCII static path segments into the registered route record at build time. Wrapping that in encodeURI` again re-encoded the existing `%XX` escapes into `%25XX`, so the rendered link (`href="/ja/%25E7%25B4%2584"`) 404'd instead of resolving to `/ja/%E7%B4%84`.

Dynamic route params don't have this problem: vue-router encodes those itself at resolve time and can return raw, unencoded text, which still needs `encodeURI` to become link-safe. So the fix can't just drop the encoding step, it has to tell already-encoded segments apart from raw ones.

Added `encodePath()` in `src/runtime/shared/utils.ts`, which encodes with `encodeURI` and then collapses only the `%25XX` sequences that call introduced on top of an already-valid `%XX` escape. Genuinely unsafe or unencoded characters (including the existing XSS regression case from #3043) are still encoded exactly as before. Used in place of the bare `encodeURI(...)` in both `SwitchLocalePathLink.ts` and `switch-locale-path-ssr.ts`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented localized links from being double-encoded, including for non-ASCII custom route paths and dynamic parameters.
  * Improved handling of reserved characters (such as `#`) and existing encoded values during locale switching.
  * Ensured server-rendered localized `href` attributes are safely escaped for HTML output.

* **Tests**
  * Expanded coverage for SSR `href` encoding/escaping, Unicode and escape behaviors, reserved characters, and navigation correctness across locales.
  * Added a non-ASCII fixture page to validate routing and link generation.


BobbieGoede:
Thanks for digging into this 🙏 The 404 is real, but I think the cause is a bit different: `switchLocalePath` copies the current route into a plain object that carries `path`, and vue-router's `resolve()` skips `encodeParams()` whenever `path` is set, even though the matcher ignores it when there's a `name`. So dynamic params come back unencoded while static custom paths are already encoded into the route record at build time, and the `encodeURI` was only ever doing anything to the params.

https://github.com/nuxt-modules/i18n/blob/559078cf/src/runtime/routing/routing.ts#L102

I'd rather match vue-router and encode as little as possible ourselves. Dropping `path` from that copy gives the same output as `localePath` and as a plain `router.resolve({ name, params })`, and then there's nothing left to encode on top so both `encodeURI` calls can go. Anything we add after the fact runs into the same ambiguity as the `%25XX` collapse here, `/count/C%2523` has a param value of `C%23` and should stay `C%2523`, but it collapses to `C%23` which now means `C#`.

I updated your branch with that. The SSR plugin escapes html instead, since that's where the XSS in #3042 happened, and your Japanese case is covered by tests along with some reserved-character ones (#2909).

PR #4080, opened 2026-07-27
fix: restrict multiDomainLocales locales per domain and prevent route collisions

Vincentdevreede:
## Summary

While preparing to use `multiDomainLocales`, I found that several parts of the implementation were not working correctly. These three related fixes address the issues I encountered and make `multiDomainLocales` work as expected.

### **Route collision in `setupMultiDomainLocales`** (`a5cbdcc`)
Under `prefix_and_default`, a locale that happened to be both the *build time* default and this domain's actual default kept its unprefixed path unchanged, while a locale that was this domain's default but wasn't the build time default only got unprefixed, without checking whether that path was already taken. Two routes could end up registered at the same path. Fixed by always computing the locale-agnostic content path first, then re-prefixing it if the locale isn't this domain's default, instead of only handling the unprefix direction.

### **Locales restricted to specific domains** (`812b938`)
Added support for a locale's `domains` to list only some of the domains in a shared `multiDomainLocales` config, rather than requiring every locale to appear on every domain:
- `setupMultiDomainLocales` now removes routes for locales not served on the current host, so visiting one 404s instead of silently rendering under the wrong locale.
- The locale switcher (`composer.locales`) now hides locales that aren't available on the current host.
- An unrecognized host (one that doesn't match any configured domain) isn't restricted at all, every locale is still served as a fallback, so things like health checks or a not-yet-configured staging domain keep working.
- Added a build time warning when both `differentDomains` and `multiDomainLocales` are enabled together, since only one is meant to be active at a time and having both causes inconsistent behavior such as the switcher hiding locales that `differentDomains` would otherwise cross-link.
- Documented all of this in `10.multi-domain-locales.md`, including a new "Restricting locales to specific domains" section.

### **`detectBrowserLanguage` redirecting across domain restrictions** (`56d91c7`)
Once locales can be restricted per domain, a locale detected from the browser or a cookie can point at a locale that doesn't exist on the current host. Added `resolveHostLocale()`, used from the server side redirect resolver, which falls back to the domain's own `defaultLocale` whenever the detected locale is restricted away from the current host, so detection never redirects to a path that would 404. Also documented the fallback behavior for unrecognized hosts, and why `defaultLocale` is still required even when every locale has its own `defaultForDomains`.


BobbieGoede:
Thanks for your PR!

I took a quick look at the diff and noticed it covers a few separate behaviors at once, and that it adds back some conditionals/exclusions specifically for `multiDomainLocales` (vs `differentDomains`). The route removal in `setupMultiDomainLocales` isn't behind that flag though, so it applies to `differentDomains` as well, where it would drop the routes `switchLocalePath` needs to resolve links to the other domains.

The route collision is a real bug, we hadn't covered the case where the build time `defaultLocale` isn't the default on a given domain. I think that one is better resolved where the routes are generated, the build time `defaultLocale` shouldn't be treated as the unprefixed locale under domains since that differs per domain.

The `detectBrowserLanguage` change I'd like to settle separately. From what I can tell it narrows undefined behavior rather than changing documented behavior, but the guide currently says a first visit can redirect to a different domain, so I want to work out what we actually want there. It would also need to apply on the client, `detectLocale` still picks up the cookie there after hydration.

I'm going to work out if we can resolve these issues while still working towards a unified approach for domain handling, and reduce the config surface for v11. I'll report back here once I have something concrete.

Vincentdevreede:
No longer needed, @BobbieGoede addressed this with a better approach.

PR #4085, opened 2026-07-27
feat(domain): restrict locales to a subset of `domains`

BobbieGoede:
* Related #4080

A locale's `domains` had to list every domain in practice. Listing a subset still served that locale's paths everywhere, but `matchDomainLocale` falls back to the domain's own default, so `https://mydomain.com/es` rendered the Spanish route in English.

A locale path the current domain doesn't serve now redirects to the domain that does, shaped for that domain's own default locale. Route records are left in the router so `switchLocalePath` and the hreflang alternates keep resolving them, and a host matching no configured domain serves every locale as before. Under `differentDomains` every locale already declares a `domain`, so this applies there too, a prefixed path for another domain's locale now redirects rather than rendering in the host's locale.

This also fixes the redirect resolver never running for a locale-prefixed root like `/ja`, where the prefix-free path was empty and treated as a non-route.

## Summary by CodeRabbit

* **New Features**
  * Added support for serving locales only on their configured domains via `locales[].domains`.
  * When a locale is requested on an incorrect domain, it redirects to the domain where that locale is available.
  * Locale switcher links now send restricted locales to their designated domains while keeping other links relative to the current host.
  * Browser-language detection is limited to locales available on the current domain, and the redirect cookie is respected to avoid cross-domain redirects.
* **Documentation**
  * Added a guide section and option clarification for “restricting locales to specific domains,” including example outcomes and fallback behavior.
* **Bug Fixes**
  * Improved domain normalization to handle mixed-case hosts and uppercase protocol prefixes correctly.
* **Tests**
  * Expanded multi-domain routing and locale restriction coverage.

### DIFF
--- a/docs/content/docs/02.guide/10.multi-domain-locales.md
+++ b/docs/content/docs/02.guide/10.multi-domain-locales.md
@@ -170,3 +170,46 @@ The same requests when using the `'prefix_except_default'`{lang="ts-type"} strat
 - https://fr.mydomain.com/en -> https://fr.mydomain.com/en (en language)
 
 A locale is only unprefixed on the domains it is the default for, on every other domain it keeps its prefix. This includes the locale set as `defaultLocale`.
+
+## Restricting locales to specific domains
+
+A locale's `domains` doesn't have to list every domain, it only has to list the domains that locale should be served on. A locale configured without `domains` is served on all of them.
+
+```ts [nuxt.config.ts]
+export default defineNuxtConfig({
+  i18n: {
+    locales: [
+      {
+        code: 'en',
+        domains: ['mydomain.com'],
+        defaultForDomains: ['mydomain.com']
+      },
+      {
+        // shared between both domains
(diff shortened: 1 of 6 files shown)


BobbieGoede:
@Vincentdevreede #4082 and #4083 have been merged and this PR covers the rest, together they should resolve the issues you described in #4080. The approach differs from yours in a few places, mainly that routes are kept so `switchLocalePath` and the hreflang alternates can still resolve them, and a locale path the current domain doesn't serve redirects to the domain that does instead of returning a 404.

Could you try the preview release of this PR by running `npm i https://pkg.pr.new/@nuxtjs/i18n@b6bf2a8` and let me know if it resolves your issues, or if you run into anything that doesn't work the way you need? 🙏

BobbieGoede:
I'm working on fixing more broken behavior related to domain features 🫣 a lot of its functionality is not covered by tests (seo, detection client/server, etc)

PR #4088, opened 2026-07-27
fix(nitro): support Nuxt 5 runtime APIs

### LINKED ISSUE #4066: feat: support Nuxt 5 and Nitro 3 while retaining Nuxt 4 compatibility
Nuxt 5 moves the server runtime to Nitro 3 and H3 v2. The i18n server runtime still imports Nitro 2 APIs and writes locale redirects through the old Node response, so it cannot run unchanged on Nuxt 5.

I have a small pnpm patch running in a private application, including a deployed preview. It updates those server seams and the application has behaved correctly there.

Could we keep Nuxt 4 and Nuxt 5 support in the same release line? The change I have in mind is deliberately narrow:

- Keep locale handling, detection, and redirect logic shared.
- Generate the tiny Nitro/H3 compatibility layer when the module is set up, so each installed Nuxt version gets imports it can resolve.
- Cover one locale redirect on both version lines, including the status, Location header, and cookie.

Nuxt Image is already moving to the Nitro 3 API style in https://github.com/nuxt/image/pull/2281. That work is a useful reference, although this proposal would retain Nuxt 4 support too.

If this direction makes sense, I can open one focused PR. If you would rather take it on internally, I hope the boundary above makes the change easy to pick up.

onmax:
### 🔗 Linked issue

Closes #4066

### 📚 Description

Nuxt 5 moves the server runtime to Nitro 3 and H3 v2, while Nuxt I18n still imports Nitro 2 runtime modules and relies on globals that are no longer available.

This adds a small generated compatibility module selected from the installed Nuxt version. Nuxt 4 keeps its existing Nitro 2 imports and render response contract; Nuxt 5 uses the Nitro 3 runtime, storage, H3, and render hook APIs. The shared locale detection and redirect logic stays unchanged, and client message loading now imports `$fetch` explicitly.

Pinned production fixtures cover Nuxt 4/Nitro 2 and Nuxt 5/Nitro 3. They verify the first locale redirect's status, `Location`, and cookie, confirm server preload fetches otherwise-unused messages, and exercise a hydrated client locale switch with preload disabled.

## Summary by CodeRabbit

* **New Features**
  * Added runtime compatibility for Nuxt 4 and Nuxt 5 applications.
  * Added support for YAML and JSON translation files.
  * Improved localized rendering, language switching, browser locale detection, and locale-based redirects.
  * Added client-side lazy loading for translations when preloading is disabled.

* **Bug Fixes**
  * Improved server-side redirect handling and translation loading across supported runtime versions.
  * Preserved translation caching and request processing behavior across environments.
  * Improved compatibility with server-side internationalization integrations.


BobbieGoede:
Thank you for your PR!

It's unfortunate we would need to re-export h3/nitro utilities to have compatibility with both nuxt 4 and 5 🤔 It feels like this or some similar re-exports for compat may happen in other modules as well.. Is this something that can be (at least partially) done upstream in nuxt 4? /cc @danielroe

onmax:
Good point. Nuxt's [`@nuxt/nitro-server/h3` compatibility export](https://github.com/nuxt/nuxt/blob/abe3bbe7e4518c77dcd4b3f67e63c9b0f88068b0/packages/nitro-server/src/runtime/h3-compat.ts) seems the right thing

I found similar version switching in:

- [Nuxt Better Auth's compatibility resolver](https://github.com/nuxt-modules/better-auth/blob/0ccb67498e920f669fd9c7891e0d69558cc5dda3/src/module/compatibility.ts)
- [Nuxt Icon's cached-handler virtual import](https://github.com/nuxt/icon/blob/fe6d83c5f4fd8b43d1344cf3a60212e12b9dc006/src/module.ts)
- [Nuxt Scripts' Nitro compatibility layer](https://github.com/nuxt/scripts/blob/ebcf948255eea409752ddce06d0ec8e6d51ce0bc/packages/script/src/nitro-compatibility.ts)
- [`nuxtseo-shared`'s compatibility layer](https://github.com/harlan-zw/nuxt-seo/blob/1e1aeb7c204c22ed1110032afc042a581168d1a1/packages/shared/src/nitro-compatibility.ts), which is then [configured by Nuxt Sitemap](https://github.com/nuxt-modules/sitemap/blob/54a74c18dfeb871720d171ef03f821ef18b3ab35/src/module.ts)

The implementations aren't identical, but they overlap around H3, cached handlers, Nitro app access, runtime config, storage, and types. Would it make sense for Nuxt to expose more compatibility primitives for module authors: stable runtime entry points under `@nuxt/nitro-server`, plus small `@nuxt/kit` helpers where the setup wiring itself is repeated?

@danielroe, If you prefer I can open a Nux issue to follow up with these APIs

paulchrisluke:
I validated PR #4088 head `68405a921fb50f65d93b7add88189062551f199f` against the exact disposable Nuxt 5/Nitro 3 reproduction (`nuxt@5.0.0-29778575.262cc544`, Nitro `3.0.260610-beta`). Before the fix, it still failed with the `useSwitchLocalePath` missing export from the server `#i18n` entry.

The direct composable-entry fix then exposed a second Nitro 3-specific alias failure: the `@intlify/utils` root alias swallowed `@intlify/utils/h3` and produced `dist/index.mjs/h3`. I opened #4134 with both fixes and SSR `SwitchLocalePathLink` regression coverage. The exact target build now passes, and the built `/fr` response renders `Bonjour` with the English switch path `/`.

khangln-amv:
hi, can we add this to the edge channel? I would be helpful if we can use nuxt i18n on Nuxt 5 nightly branch, even if it's unstable

PR #4099, opened 2026-07-28
fix(routing): re-escape '#' and '?' left over from localePath's path normalization

### LINKED ISSUE #4098: String paths passed to localePath() decodes percent-encoded characters in dynamic path segments
> Fix proposed in #4099.

## Environment

```
- Operating System: Darwin (macOS)
- Node Version:     v24.18.0
- Nuxt Version:     4.5.0
- Nitro Version:    2.13.4
- Package Manager:  pnpm@11.17.0
- Runtime Modules:  @nuxtjs/i18n@10.5.0
- vue-router:       5.2.0
```

## Reproduction

https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-mwca4nqv?file=app%2Fpages%2Findex.vue

`nuxt.config.ts`
```ts
  modules: ['@nuxtjs/i18n'],
  i18n: {
    defaultLocale: 'en',
    locales: [
      { name: 'English', code: 'en', language: 'en-US' },
      { name: 'Deutsch', code: 'de', language: 'de-DE' },
    ],
    detectBrowserLanguage: false,
  },
})
```

`app/pages/blog/[...slug].vue`
```vue
<template><div>Blog post page</div></template>
```

`app/pages/index.vue`
```vue
<script lang="ts" setup>

const rawPath = '/blog/' + encodeURIComponent('my post #1')
const localized = localePath(rawPath)
</script>

<template>
  <div>
    <pre id="raw-path">{{ rawPath }}</pre>
    <pre id="localized-path">{{ localized }}</pre>
  </div>
</template>
```

Load `/`. Observe `#localized-path`.

## Describe the bug

`localePath()` decodes percent-encoded characters instead of preserving them when the input path resolves to a route by matching a route with the *default* locale (no prefix needed) and that route's name can be looked up (i.e. the path matches an existing route record).

With the reproduction above:

- `rawPath` = `/blog/my%20post%20%231`
- `localePath(rawPath)` = `/blog/my post #1` — **decoded**, not `/blog/my%20post%20%231`

`#` is a URL fragment delimiter. A caller that uses the `localePath()` result directly as a navigation target (`navigateTo(localized)`, `router.replace({ path: localized })`, or binding it to `<NuxtLink :to="...">`) ends up navigating to a truncated path with a bogus fragment attached, silently dropping everything after the `#`.

### Root cause

For the default (`prefix_except_default`) strategy, `localePath('/blog/my%20post%20%231')` goes through:

1. `routing/routing.js` → `resolveRoute()`:
   ```js
   function resolveRoute(ctx, route, locale) {
     const normalized = normalizeRawLocation(route);
     const resolved = ctx.router.resolve(ctx.resolveLocalizedRouteObject(normalized, locale));
     ...
   ```
2. `ctx.resolveLocalizedRouteObject` → `routing/context.js` → `resolveLocalizedRouteByPath()`:
   ```js
   function resolveLocalizedRouteByPath(input, locale) {
     const route = routeByPathResolver(input, locale);
     const baseName = getRouteBaseName(route);
     if (baseName) {
       ...
       route.name = localizedName;
       return route;   // returns the object from step 3, already resolved
   ```
3. `routeByPathResolver` → `routing/utils.js` → `createLocalizedRouteByPathResolver()`, for any strategy other than `'prefix'`:
   ```js
   return (route) => router.resolve(route);   // <-- FIRST router.resolve() call
   ```

So step 3 does the first `router.resolve()` (correctly encoded), step 2 just mutates `.name` on that *already-resolved* object and returns it unchanged otherwise, and step 1's own `ctx.router.resolve(...)` call resolves that already-resolved object a **second** time. Vue Router's `resolve()` decodes the path when given an already-resolved `RouteLocationNormalized`-like object as input, and does not re-encode on output. Reproducible with vue-router alone:

```ts
const router = createRouter({ history: createMemoryHistory(), routes: [{ path: '/blog/:slug(.*)*', name: 'blog', component: {} }] })
const first = router.resolve('/blog/my%20post%20%231')
console.log(first.fullPath)                  // /blog/my%20post%20%231  (correct)
console.log(router.resolve(first).fullPath)  // /blog/my post #1        (decoded!)
```

This looks to be the same double-resolve pattern identified in https://github.com/nuxt-modules/i18n/issues/2909#issuecomment-2051812893. #2698/#2909 were closed as fixed by #3283, but the module appears to have been restructured again since then.

#4079 (merged a few days ago) fixed the same underlying issue for `switchLocalePath()` specifically, by omitting `path` from the object passed into the second `router.resolve()` call so it takes the name+params branch instead (which re-encodes correctly via `encodeParam`). That fix doesn't cover this case: `localePath()` called directly with a path string always goes through the path-based branch (`resolveLocalizedRouteByPath`), so the second resolve still receives an object with `path` set.

## Additional context

Found while migrating a Vue Router `[...path]` catch-all route (S3-object-style paths, some containing `#` in filenames). Worked around by prefixing the locale manually and skipping `localePath()`/`router.resolve()` entirely for these paths.

## Logs

n/a

dargmuesli:
Resolves #4098

## Summary

`resolveRoute()` (used by `localePath`, `localeRoute`, and `switchLocalePath`) resolves an already-resolved route object a second time to attach the localized route name (`ctx.router.resolve(ctx.resolveLocalizedRouteObject(normalized, locale))`). For the `prefix_except_default`/`prefix_and_default` strategies, `resolveLocalizedRouteObject`'s path-based branch has already gone through one `router.resolve()` call (`createLocalizedRouteByPathResolver`), which decodes percent-encoded characters — kept intentionally, for readable URLs (`%20` → literal space, already covered by an existing test in `test/kit.test.ts`).

The second `router.resolve()` call re-parses that decoded `path` as a fresh location string. A literal `#` or `?` left over from decoding is read as an actual fragment/query delimiter rather than path text — confirmed directly against vue-router alone, no i18n module involved:

```ts
const router = createRouter({ history: createMemoryHistory(), routes: [{ path: '/blog/:slug(.*)*', name: 'blog', component: {} }] })
const first = router.resolve('/blog/my%20post%20%231')
console.log(first.fullPath)                  // /blog/my%20post%20%231  (correct)
console.log(router.resolve(first).fullPath)  // /blog/my post #1        (decoded, '#' now a literal fragment delimiter)
```

`query`/`hash` are populated separately during resolution and are unaffected by this — verified they stay structurally correct (`{"foo":"bar"}`, not swallowed into the corrupted path) even while `path`/`fullPath` are wrong. Any caller that uses the `localePath()`/`switchLocalePath()` output directly as a navigation target — `navigateTo(...)` (SSR redirects), `router.replace({ path })`, `<NuxtLink :to>` — ends up navigating to a silently truncated path with a bogus fragment attached.

This is the same underlying double-resolve pattern #4079 fixed for `switchLocalePath`'s `routeCopy` (by omitting `path` so the second resolve takes the name+params branch, which does correctly re-encode via `encodeParam`). That approach doesn't work here without regressing the "readable URL" behavior `test/kit.test.ts` already covers (`localePath('/path/as%20a%20test...')` decoding to `/path/as a test...`) — using name+params always fully re-encodes via `encodeURI`, including turning spaces into `%20`. So rather than avoiding the second resolve, this re-escapes only the two characters that are genuinely unsafe to leave literal in a path string (`#` → `%23`, `?` → `%3F`) on the already-resolved object, after both resolves, right before the caller gets it.

## Fix

Added `sanitizeResolvedPath()` in `src/runtime/routing/routing.ts`, applied at both return points of `resolveRoute()` (the shared function `localePath`/`localeRoute`/`switchLocalePath` all funnel through). No-ops when `path` doesn't contain `#`/`?` (the common case).

## Test plan

- [x] Added a regression case to the existing `localePath` test in `test/kit.test.ts`, covering all four strategies, alongside the existing "readable URL" assertion it must not regress
- [x] `pnpm lint` — clean
- [x] `tsc --noEmit` — clean
- [x] `pnpm test:unit` — 407 passed, 1 skipped (unchanged from main + the 1 new assertion)
- [x] `pnpm build` — clean
- [x] `pnpm test:e2e` — 201 passed (33 files)

`pnpm test:types`'s `specs/fixtures/typed_routes` sub-check fails in my local sandbox with unrelated `Cannot find module '#app'`/`__I18N_*__` errors — reproduces identically on a clean, unmodified `main` checkout (verified via `git stash`), so it's a pre-existing local environment issue, not something this change introduces.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed localized route resolution to keep percent-encoded reserved delimiters inside path segments (including `%23` and `%3F`) from being treated as fragment/query separators on subsequent resolutions.
  * Ensured `fullPath`/`href` are correctly recomputed to preserve the expected escaping behavior.
* **Tests**
  * Added regression tests verifying `%23` and `%3F` remain encoded for both path-based and route-based localization resolution.

### DIFF
--- a/src/runtime/routing/routing.ts
+++ b/src/runtime/routing/routing.ts
@@ -57,17 +57,37 @@ function normalizeRawLocation(route: RouteLocationRaw): RouteLike {
 }
 
 /**
+ * `resolveLocalizedRouteObject` may return an already-resolved route (its `path`
+ * decoded for readability, e.g. `%20` becomes a literal space — see `context.ts`
+ * `resolveLocalizedRouteByPath`). Resolving that object a second time below re-parses
+ * `path` as a fresh location string, so a literal '#' or '?' left over from decoding is
+ * read as an actual fragment/query delimiter instead of path text, silently truncating
+ * the path — `query`/`hash` are unaffected, populated separately and not re-parsed.
+ * Only `path` (and the `fullPath` built from it) needs the fix; re-escape both.
+ */
+function sanitizeResolvedPath<T extends { path: string, fullPath: string }>(resolved: T): T {
+  if (!/[#?]/.test(resolved.path)) {
+    return resolved
+  }
+
+  const suffix = resolved.fullPath.slice(resolved.path.length)
+  resolved.path = resolved.path.replace(/#/g, '%23').replace(/\?/g, '%3F')
+  resolved.fullPath = resolved.path + suffix
+  return resolved
+}
+
+/**
  * Try resolving route and throw on failure
  */
 function resolveRoute(ctx: RoutingContext, route: RouteLocationRaw, locale: Locale) {
   const normalized = normalizeRawLocation(route)
   const resolved = ctx.router.resolve(ctx.resolveLocalizedRouteObject(normalized, locale))
   if (resolved.name) {
-    return resolved
+    return sanitizeResolvedPath(resolved)
   }
 
   // if unable to resolve route try resolving route based on original input
-  return ctx.router.resolve(route)
+  return sanitizeResolvedPath(ctx.router.resolve(route))
 }
 
 /**
--- a/src/runtime/routing/routing.ts
+++ b/src/runtime/routing/routing.ts
@@ -63,9 +63,13 @@ function normalizeRawLocation(route: RouteLocationRaw): RouteLike {
  * `path` as a fresh location string, so a literal '#' or '?' left over from decoding is
  * read as an actual fragment/query delimiter instead of path text, silently truncating
  * the path — `query`/`hash` are unaffected, populated separately and not re-parsed.
- * Only `path` (and the `fullPath` built from it) needs the fix; re-escape both.
+ * Only `path` (and the `fullPath`/`href` strings built from it) need the fix; re-escape
+ * both characters and recompute `href` the same way vue-router derives it from `fullPath`.
  */
-function sanitizeResolvedPath<T extends { path: string, fullPath: string }>(resolved: T): T {
+function sanitizeResolvedPath<T extends { path: string, fullPath: string, href?: string }>(
+  ctx: RoutingContext,
+  resolved: T,
+): T {
   if (!/[#?]/.test(resolved.path)) {
     return resolved
   }
@@ -73,6 +77,9 @@ function sanitizeResolvedPath<T extends { path: string, fullPath: string }>(reso
   const suffix = resolved.fullPath.slice(resolved.path.length)
   resolved.path = resolved.path.replace(/#/g, '%23').replace(/\?/g, '%3F')
   resolved.fullPath = resolved.path + suffix
+  if (resolved.href !== undefined) {
+    resolved.href = ctx.router.options.history.createHref(resolved.fullPath)
+  }
   return resolved
 }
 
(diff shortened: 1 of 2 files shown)


BobbieGoede:
Thanks! I'm a bit cautious with changes like these, ideally we leave encoding/decoding up to vue-router as much as possible as we have ran into issues trying to fix this in the past. Going to explore some other approaches as well.

dargmuesli:
Yes, I'd expect stuff like this to be vue (extension) logic instead of nuxt (extension) logic too of course. I just stumbled upon this and want to point to the finding :)

Issue #4016, opened 2026-06-28
Runtime-configurable default locale: single build serving regions with different defaults (prefix_except_default)


Togetic:
### Describe the feature

**Summary**

Allow the **default locale to be chosen at runtime** under `strategy: 'prefix_except_default'`, so a single build artifact can serve multiple deployments that each have a different default locale.

Today `defaultLocale` is a build-time decision. The default-locale route tree is generated unprefixed and named per-locale (`<base>___en`), and the runtime name getter (`createLocaleRouteNameGetter`) binds the default locale to that build value. To serve a different default locale you must rebuild. For "one image" deployments — one Docker image, N regional pods, each pod's default locale set from an env var at boot — this forces either N builds or a fork patch.

**Use case**

- A single CI build artifact deployed to multiple regions.
- Each region's default (unprefixed `/`) locale differs and is chosen at runtime (e.g. `NUXT_PUBLIC_I18N_DEFAULT_LOCALE`).
- Non-default locales stay prefixed (`/de/…`), exactly as `prefix_except_default` works today.
- Pairs naturally with `experimental.compactRoutes` (non-default locales are already one regex route).

**Why it can't fully be a module today (the crux)**

The *build* side can be done from a module: a `pages:resolved` hook can rename the default-locale tree to a **locale-agnostic** name (`<base>___default` instead of `<base>___en`) and widen the compact `:locale(...)` param to include the build default. That part works in userland.

The *runtime* side cannot. `createLocaleRouteNameGetter` (`runtime/routing/utils.ts`) hardcodes, for `prefix_except_default`, `getLocalizedRouteName(name, locale, /* isDefault */ false)` → `<base>___<locale>`. For name-based resolution (`localePath`/`localeRoute`, and the hreflang/canonical link generation built on them) to target the locale-agnostic tree, the default locale must resolve to `<base>___default`. There is no hook or option to influence this runtime getter, so the only way to make it agree with the renamed tree is to **patch `dist/`** — which is fragile across releases.

**Proposed direction**

An opt-in experimental flag (working name `experimental.localeAgnosticDefaultRoutes`) that, under `prefix_except_default`, resolves the default locale to the locale-agnostic `___default` route name in `createLocaleRouteNameGetter` — turning the `dist/` patch into a supported option. A complete feature would additionally:

1. generate the default-locale tree with the `___default` name natively (so no companion module rename is needed), and
[...]

Issue #3748, opened 2025-07-22
Support complex multi-domain i18n configurations in a single build


Ereaey:
### Description

Hi there,
I'm working on a Nuxt project that needs to serve multiple countries, each on a separate domain. Each domain has its own i18n requirements:

* Some domains have **only one official language** (e.g., `site.fr` → French only, no prefix needed).
* Others have **multiple languages**, and the strategy varies:

  * Some use `prefix_except_default` (e.g., `site.ca` with `fr` and `en`)
  * Others use `prefix` (e.g., `site.be` with `fr` and `nl`)

### Constraints:

* I want to **use a single Nuxt build** to serve all domains.
* I need to **dynamically adapt the i18n config** (strategy, locales, defaultLocale) based on the current domain (host).
* I also need to **detect the current domain** at runtime (server & client) to configure things like API base URLs.
* Compatible with SSR & SEO

### Problem

Currently, I need to generate and deploy **one build per domain** to accommodate the differences in `strategy`, `defaultLocale`, and available `locales`. That means I must build the app **8 times**, which is time-consuming and hard to maintain.

### Question

Is there currently any **official or recommended way** to support multiple domains with different i18n strategies **using a single build**?

If not, would this be considered a valid feature request, or are there technical constraints that prevent this kind of setup?

Thanks a lot for your help!

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [x] Read the [contribution guide](https://nuxt.com/docs/community/contribution) (The contribution guideline of nuxt-modules/i18n is compliant with Nuxt too).
- [x] Check existing [discussions](https://github.com/nuxt-modules/i18n/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

OlePc:
+1 for this. I've used layers in Nuxt to differentiate between the different site configurations I need. The issue is that a layer has to be set at build time, so I need a build per configuration. I would love to be able to override strategy, locales, and defaultLocale using runtime environment variables. I've tried to override the default values using plugins in Nuxt, but the timing is tricky and you end up with a lot of side effects.

Quineone:
+1 for this.

LindseyLouv:
+1 for this aswell !

romain-banchetti:
+1

JoniJnm:
In my case I need to change the defaultLocale and also disable some locales dynamically 

[...]

Issue #2656, opened 2024-01-05
Wrong locale of pre-rendered site


signor-pedro:
### Environment

Nuxt: 3.8.0
Nuxti18n: 8.0.0

### Reproduction

https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-7efeov

### Describe the bug

When `differentDomains` is set
```js
// nuxt.config.ts

locales: [
      {
        name: 'English',
        code: 'en',
        iso: 'en-US',
        domain: 'sth.com',
      },
      {
        name: 'Nederlands',
        code: 'nl',
        iso: 'nl-NL',
        domain: 'sth.nl',
      },
    ],
    differentDomains: true,
```
and a page is translated
```js
// pages/sth.vue

defineI18nRoute({
  paths: {
    en: "/something",
    nl: "/else",
  },
});
```
The output of the prerendering is never translated, but always in the default locale.
```js
// nuxt.config.ts
nitro: {
    prerender: {
      routes: [
        '/something', // should be in English
        '/else' // should be in Dutch
      ]
    }
  }
```

<img width="809" alt="scr" src="https://github.com/nuxt-modules/i18n/assets/53906348/874ace11-65a1-4a00-b8a9-2af404827b95">
(should be in Dutch)

WooHoo-Wu:
Yes, I have this problem too.

jekuer:
Experiencing the same issue even without translated paths. Basically with 2 different domains, the prerender always picks up the default only.
Basically because of exactly this circumstance - there are no different paths - so, nitro sees no need to render another page (at least my interpretation).

varzy:
I am also facing the same problem

Issue #2602, opened 2023-12-06
Support different domains config at runtime


bcadu:
### Environment

- Operating System: Linux
- Node Version:     v18.14.2
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@9.5.0
- Builder:          -
- User Config:      devtools, modules, pages, runtimeConfig, i18n
- Runtime Modules:  @nuxtjs/i18n@8.0.0-rc.7
- Build Modules:    -

### Reproduction

The code
https://stackblitz.com/edit/nuxt-starter-lm7gng

And a minimal Apache configuration
```
<VirtualHost *:443>
    
    ServerName fr.localhost
    
    ServerAlias fr.localhost2
    ServerAlias en.localhost
    ServerAlias es.localhost
    
    SSLEngine on
    SSLCertificateFile /etc/ssl/certs/localhost.crt
    SSLCertificateKeyFile /etc/ssl/certs/localhost.key
    
    ProxyPass / http://localhost:3000/
    ProxyPassReverse / http://localhost:3000/
</VirtualHost>
```

### Describe the bug

I'm encountering an issue with the Nuxt i18n module where its behavior differs between the development and production environments, specifically after a single build process.

**Development Environment:**
- The module functions as expected, aligning with the values set in the `.env` file.

**Production Environment:**
- After building the project, I observed two different behaviors based on how the application is started:

  1. **First Start (`npm run start`):**
     - The application behaves as expected. The default domain value from the build process is utilized.
       ```javascript
       {
           code: "fr",
           domain: process.env.NUXT_PUBLIC_I18N_LOCALES_FR_DOMAIN || "fr.localhost",
           file: "fr-FR.json",
       }
       ```
       Accessing https://fr.localhost/, the following is displayed:
       - Index Page
       - "Bienvenue"
       - Locale: fr
       - Site name: My Site Name

  2. **Second Start (`npm run start:2`):**
     - When accessed via https://fr.localhost2/, the results are inconsistent:
       - Index Page
       - "Welcome"
       - Locale: en (unexpectedly switches to the default locale)
       - Site name: My Site Name 2

     In this case, while the `NUXT_PUBLIC_SITE_NAME` variable is correctly recognized at runtime, the `NUXT_PUBLIC_I18N_LOCALES_FR_DOMAIN` variable is not, and the application defaults to the default locale.

[...]

PR #4101, opened 2026-07-28
feat: add multiDomainLocales: { isolate: true } for regional brand domains

Vincentdevreede:
## Summary
Adds `multiDomainLocales: { isolate: true }` for the case where each domain sells the same product under its own regional brand rather than one brand serving every region. Each domain should
  behave like a fully self-contained site. It should never redirect to or link into another domain's locale.

  With `isolate: true`:
  - A locale-prefixed path not served on the current domain **404s** instead of being redirected to the domain that serves it.
  - The locale switcher and hreflang/canonical alternates **only list locales served on the current domain**.
  - Browser-language detection and a `detectBrowserLanguage.cookieDomain` spanning multiple domains **never cross domains**. A first visit detected as another domain's locale stays on the current
  domain in its own locale instead of relocating.
  - A host matching no configured domain (staging, health checks) is unaffected. It still serves every locale, per the existing non-isolated fallback.

@BobbieGoede, feel free to accept this or not, and make any changes you see fit.

## Summary by CodeRabbit

* **New Features**
  * Added isolated multi-domain locale support with `multiDomainLocales: { isolate: true }`.
  * Unsupported locale paths now return 404 instead of redirecting to another domain.
  * Locale switchers and SEO alternate links show only locales available on the current domain.
  * Browser-language detection and locale cookies remain within the current domain.

* **Documentation**
  * Updated configuration and guide documentation with isolation behavior and usage details.

* **Tests**
  * Added coverage for domain-specific routing, detection, redirects, and SEO locale links.

---

## ~~Summary~~
- ~~Add `multiDomainLocales: { isolate: true }` for hosting genuinely unrelated brands on one shared build. A locale not served on the current domain now 404s instead of redirecting cross-domain, and the locale switcher and hreflang alternates only list locales actually served on the current domain.~~
- ~~Revert part of upstream #4083, which made `detectBrowserLanguage` unconditionally stay within the current domain for every domain setup. That's the right call for `{ isolate: true }`, since unrelated brands must never reveal a connection, but it was wrong for plain `multiDomainLocales`/`differentDomains`. The entire point of `differentDomains` is one product spanning several regional domains, so a detected locale should be able to send a visitor to the domain that actually serves it, exactly like an explicit locale-prefixed path already does. #4083's actual bug was narrower than that: an adopted off-host locale built a broken path on the current domain instead of relocating anywhere. This restores the intended cross-domain behavior for `differentDomains` and plain `multiDomainLocales` without reintroducing that original bug, and keeps `{ isolate: true }` exactly as strict as before.~~ 

This is an older summary from before I force-pushed the changes. I’m just keeping it here for reference.

### DIFF
--- a/docs/content/docs/02.guide/10.multi-domain-locales.md
+++ b/docs/content/docs/02.guide/10.multi-domain-locales.md
@@ -223,6 +223,38 @@ Locales served on other domains stay available in `locales` and in the locale sw
 
 A request on a host that doesn't match any configured domain, such as a staging domain or a health check by IP, isn't restricted. Every locale is served there and `defaultLocale` is used as the unprefixed default, so it's worth setting even when every domain has its own default through `defaultForDomains`. Redirects on such a host stay relative, they never send a visitor to one of the configured domains.
 
+## Isolating domains with their own regional brand name
+
+The behavior above assumes every domain is the same brand shown in a different language, so a locale not served on the current domain still belongs to the same site and is worth redirecting to and worth linking to from the switcher. Some products are sold under a different brand name per market instead, on the same build, without sharing a single public identity. Each domain should feel like its own self-contained site for that market's brand name rather than a variant of another domain's. Set `multiDomainLocales` to `{ isolate: true }`{lang="ts"} for this case:
+
+- A locale-prefixed path not served on the current domain 404s instead of being redirected to the domain that serves it.
+- The locale switcher and hreflang/canonical alternates only list locales served on the current domain.
+- Browser-language detection never crosses domains either. A first visit whose detected language belongs to another domain stays on the current domain in its own locale, and a `detectBrowserLanguage.cookieDomain` spanning the domains no longer follows the cookie across them.
+
+```ts [nuxt.config.ts]
+export default defineNuxtConfig({
+  i18n: {
+    locales: [
+      {
+        code: 'en',
+        domains: ['brand-a.com'],
+        defaultForDomains: ['brand-a.com']
+      },
+      {
+        code: 'fr',
+        domains: ['brand-b.com'],
+        defaultForDomains: ['brand-b.com']
+      }
+    ],
+    defaultLocale: 'en',
+    strategy: 'prefix_except_default',
+    multiDomainLocales: { isolate: true }
+  }
+})
+```
+
+Given the above configuration, `https://brand-a.com/fr` 404s rather than redirecting to `brand-b.com`, and `brand-a.com`'s switcher and hreflang alternates only mention `en`. A host matching no configured domain is unaffected by `isolate`. It still serves every locale, per the staging/health-check exception described above.
+
 ## `defaultLocale` and `x-default`
 
 The domains are annotated as one cluster, each page links to its alternates on the other domains. A cluster has a single fallback for unmatched languages, so the `x-default` alternate is taken from `defaultLocale` rather than from the locale a domain happens to default to - otherwise every domain would name a different one.
--- a/docs/content/docs/04.api/00.options.md
+++ b/docs/content/docs/04.api/00.options.md
@@ -350,10 +350,12 @@ Set this to `true`{lang="ts"} when using different domains for each locale, with
 
 ## multiDomainLocales
 
-- type: `boolean`{lang="ts-type"}
+- type: `boolean | { isolate?: boolean }`{lang="ts-type"}
 - default: `false`{lang="ts"}
 
 Set this to `true` when using different domains with different locales. If enabled, you MUST configure locales as an array of objects, each containing a `domains` and `defaultForDomains` key. Refer to the [Multi Domain Locales](/docs/guide/multi-domain-locales) for more information.
+
+Pass `{ isolate: true }` instead of `true` when each domain has its own regional brand name for the same product: a locale not served on the current domain 404s instead of being cross-domain redirected there, and the locale switcher/hreflang alternates only include locales served on the current domain. See [Isolating domains with their own regional brand name](/docs/guide/multi-domain-locales#isolating-domains-with-their-own-regional-brand-name).
 
 ## compilation
 
--- a/specs/fixtures/multi_domains_locales/pages/index.vue
+++ b/specs/fixtures/multi_domains_locales/pages/index.vue
@@ -2,7 +2,7 @@
 import { useHead } from '#imports'
 import { useI18n, useLocalePath, useLocaleHead } from '#i18n'
 
-const { t, locale } = useI18n()
+const { t, locale, localeCodes, getBrowserLocale, getLocaleCookie } = useI18n()
 const localePath = useLocalePath()
 const i18nHead = useLocaleHead({ seo: { canonicalQueries: ['page'] } })
 
@@ -30,6 +30,13 @@ useHead(() => ({
     </section>
     <section>
       <code id="extend-message">{{ t('my-module-exemple.hello') }}</code>
+    </section>
+    <section>
+      <code id="locale-codes">{{ localeCodes }}</code>
+    </section>
+    <section>
+      <code id="browser-locale">{{ getBrowserLocale() }}</code>
+      <code id="cookie-locale">{{ getLocaleCookie() }}</code>
     </section>
     <NuxtLink id="link-about" exact :to="localePath('about')">{{ $t('about') }}</NuxtLink>
     <NuxtLink id="link-blog" :to="localePath('blog')">{{ $t('blog') }}</NuxtLink>
--- a/specs/multi_domains_locales/multi_domains_locales_isolate.spec.ts
+++ b/specs/multi_domains_locales/multi_domains_locales_isolate.spec.ts
@@ -1,0 +1,110 @@
+import { test, expect } from 'vitest'
+import { fileURLToPath } from 'node:url'
+import { setup, undiciRequest } from '../utils'
+import { getDom, getDataFromDom } from '../helper'
+
+// `brand-a` and `brand-b` are the same product under two different regional brand names -
+// `isolate: true` keeps them from ever redirecting or cross-linking into each other's locale
+await setup({
+  rootDir: fileURLToPath(new URL(`../fixtures/multi_domains_locales`, import.meta.url)),
+  nuxtConfig: {
+    i18n: {
+      baseUrl: 'http://localhost:3000',
+      defaultLocale: 'en',
+      locales: [
+        {
+          code: 'en',
+          language: 'en',
+          name: 'English',
+          domains: ['brand-a.nuxt-app.localhost'],
+          defaultForDomains: ['brand-a.nuxt-app.localhost']
+        },
+        {
+          code: 'fr',
+          language: 'fr',
+          name: 'Français',
+          domains: ['brand-b.nuxt-app.localhost'],
+          defaultForDomains: ['brand-b.nuxt-app.localhost']
+        }
+      ],
+      multiDomainLocales: { isolate: true },
+      strategy: 'prefix_except_default',
+      detectBrowserLanguage: {
+        useCookie: true,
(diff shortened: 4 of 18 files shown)


BobbieGoede:
Thanks for the PR!

Same as with #4080 this covers a few behaviors at once, could you split the fixes from the feature? As far as I can tell only one part of this fixes behavior that's expected today, the rest changes or adds behavior, and that's hard to prioritize in one PR.

From what I can tell the detection change isn't restoring behavior we broke. `domainFromLocale` in v10.4.1 and v10.5.0 ends at `lang?.domains?.find(v => v === url.host)`, so the origin is always the current host, no released version redirects to another domain based on detection. v10.5.0 sends an `es` cookie on `mydomain.com` to `mydomain.com/es` and renders `es` there, a locale's `domains` didn't restrict routing yet. So this looks more like a change in behavior than a fix to me, and I'd like to work out what we actually want there before we change it.

I think #4083 may have introduced a regression, `domainFromLocale` now falls back to `relocateHostForLocale` (https://github.com/nuxt-modules/i18n/blob/917c043f/src/runtime/shared/domain.ts#L74) so the redirect origin can resolve to a domain that isn't the current host. That's what your `multi_domains_locales_redirect_scope` case runs into, I'll fix it separately. I don't think the fix here fully covers it, your two tests show `brand-c/` going to `/en` and `/en` on `brand-c` relocating to `brand-a`, so it ends up on the same domain with an extra hop.

Could you open `isolate` as its own PR? Gating the route pruning behind the define resolves what I brought up on #4080, it's just easier to judge once the detection behavior is settled.

Could you also check this against the edge release (`@nuxtjs/i18n-edge`) or main? None of the domain changes are released yet so 10.5.0 behaves quite differently, it would help to know which behavior you're running into 🙏

Vincentdevreede:
@BobbieGoede I think the detection change is restoring behavior that is broken along the way. 
it was already mentioned in the creation of the doc file in 2022 ([8190810](https://github.com/nuxt-modules/i18n/blob/8190810dae0b9679d8b9a4a8ed37d3e80ce45bb0/docs/content/3.guide/10.different-domains.md#runtime-environment-variables)). If i than just skip a year because i couldn't find the implementation commit and in the v8.0.0 release the `navigate` function tells it was working someday, because when [differentDomains is true](https://github.com/nuxt-modules/i18n/blob/v8.0.0/src/runtime/utils.ts#L391), the code deliberately bypasses the normal in-app router navigation.
So I’m not sure exactly when it broke. I can look into it further if you really want me to, but I’m fairly sure it’s broken :smile: 

I can make a separate PR for the isolate feature because it was all in [1 commit](https://github.com/nuxt-modules/i18n/pull/4101/changes/cf080f70085c5c942be0fc5fa3d19aa8f4e4a71b), but i don't think it makes much sense if cross domain will be removed anyway? [Second commit](https://github.com/nuxt-modules/i18n/pull/4101/changes/c5efcc98eb22348e52ba0b4d20b59f494de6b308) is the cross domain in combination with the isolation feature and the other commits where just Coderabbit suggestions that it made.

Also, I wasn’t running a released version of i18n. I had checked out the `main` branch of yesterday evening and tested it in the playground.

Vincentdevreede:
@BobbieGoede Okay i did some research.

## History of crossdomain redirects

**2022-11-03 - #1606, "fix: different domain loop redirection"**
This is where `useRedirectState` was introduced, but it was already a bug fix for differentDomains:
```ts
if (process.client) {
  if (state.value !== redirectPath) {
    state.value = ''
    window.location.assign(redirectPath)
  }
} else if (process.server) {
  state.value = redirectPath
}
```
`state.value` starts as `''`, so on a client-initiated cross-domain switch the check was true immediately and `window.location.assign` fired. Server-side, `state.value` was set unconditionally, purely so a follow-up call could detect "we already redirected for this path" and avoid a loop.

**2023-08-15 - #2318, "fix: loop redirection on differentDomains"**
The client check picked up a `state.value && guard`:
```ts
if (process.client) {
  if (state.value && state.value !== redirectPath) {
```

Now the client branch also needs `state.value` to already be truthy before it'll fire. First-touch client redirects stop happening here. Server branch is untouched, still unconditional.

**2023-09-13 - #2404, "fix: erratic redirection on different domains on server mode"**
Both branches get folded under one shared gate:
```ts
if (state.value && state.value !== redirectPath) {
  if (process.client) {
    state.value = ''
    window.location.assign(redirectPath)
  } else if (process.server) {
    state.value = redirectPath
  }
}
```
This is the one that actually kills it. The server-side setter is now behind the same `state.value && check as everything else`, but `state.value` starts empty and `useRedirectState` has exactly one read/write site in the whole codebase. Nothing ever primes it. From this commit on, the condition can never become true, so neither branch including the `window.location.assign` can ever run.

**2023-12-25 - v8.0.0 ships**
The code is still sitting there, reads like it should work, and is what I originally linked to as proof cross-domain redirects worked. It didn't, it had already been dead for just over three months by release. 

**2025-05-20 - #3622, "refactor: remove dead code for different domains redirection"**
So the bypass looked alive in the source through v8 and v9, but was functionally inert the whole time. It got deleted by you and I think your conclusion in the PR is correct

## Conclusion
The intent was clearly that cross-origin redirects on detection should work for differentDomains, that's the only reason the useRedirectState / window.location.assign bypass exists at all, since Vue Router has no way to navigate across origins on its own. 
It did work for at least 9 months, when it was silently broken. :disappointed:

BobbieGoede:
@Vincentdevreede thanks for the deep dive! 🙏

You're right, it seems that at least redirection/navigation was broken for a long time for domain features.. (#4105 restores the cross-domain redirect) And our tests covered less than I thought they did. I'm also not sure how this went unreported for such a long time 🤔

I also found out that one of the things that broke language switching between domains is because NuxtLink (since nuxt v3) adds `noreferrer`, this triggered detection when switching domain (which in turn sends the user back to the original domain) because detection could not see where the user navigated from.

We can deal with that with a modified `<SwitchLocalePathLink>` (https://github.com/nuxt-modules/i18n/pull/4105/changes#diff-d23525f16313176c17032f17d1b4b42049ffe93b6fed332ef9d135c2ca6c7cd2R36-R56) and recommending its use in the docs.

A second issue is `detectBrowserLanguage.cookieDomain`, which defaults to the current host, which means it's not visible to other configured domains, this may be unexpected behavior to some. If locales are on different subdomains but share the same root domain, `cookieDomain` should be set to the root domain to be used across these subdomains.

I suspect the redirect state tracking was added to work around the symptoms of both the `noreferrer` and the `cookieDomain` behaviors.

I would like to get a release out this week (it includes a few fixes and some major perf improvements), but want to get `differentDomains` and `multiDomainLocales` to a working state for it. Domain isolation would likely come after that release in the interest of stability, and honestly I'm still thinking about whether domain isolation is within the scope of i18n at all. The domain features, linking to urls of the same website/brand in different languages makes sense. But isolating domains to share a build across unrelated websites/brands seems unrelated to i18n but more of an optimization to a specific use case (single server + build).

The fixes and hardening of the domain features are on #4105, feel free to try out the preview release and report feedback 🙏

Vincentdevreede:
@BobbieGoede I see the new release is live! 👍 

What about the isolation function, I think is a fair concern. The “different brands sharing a single build” example may have made the feature sound more unrelated to i18n than I intended. It was mainly the clearest example I could think of to explain the isolation behaviour in the documentation.

The intended use case is not necessarily a set of completely unrelated websites. It can also be the same company or product operating under different names in different countries. A well-known example would be Lay’s, which is known as Walkers, Smith’s or Sabritas in different markets.

Those websites may still share the same application, content structure and translations, but not every configured locale should be considered available on every domain. Even when the visitor’s language is technically available elsewhere in the build, you may not want to redirect them to another country’s domain or brand name.

For example, a visitor on the Walkers website should not necessarily be redirected to the Lay’s or Smith’s domain just because their preferred language is available there. For that domain, the locale should instead be treated as unavailable.

That is why I see this as part of the i18n domain functionality rather than only as a build optimization. multiDomainLocales already defines which locales belong to which domains. Isolation makes that relationship consistent by ensuring that locale detection, routing and navigation only use the locales that are actually available for the current domain.

The shared build is useful for this setup, but it is not the main purpose of the feature. The main purpose is controlling locale availability per domain and preventing unintended navigation between different regional identities of the same product or organisation.

Vincentdevreede:
@BobbieGoede I’ve updated my PR. I used my original PR as a reference and reworked the feature from scratch on top of v10.6.0. I no longer refer to different brands, but instead use the example of a single brand operating under different regional names.

Vincentdevreede:
@coderabbitai review

BobbieGoede:
Thank you for the further clarification 🙏 The intended use case is clear to me, and I think I managed to narrow my concerns about its implementation.

Today the module assumes a project has a single identity: every configured locale is an alternate representation of the same content, and the config describes where those alternates live. `multiDomainLocales` as it exists fits that model, it says *where* each locale is served, which is why an off-host locale path relocates instead of 404ing.

Looking at the diff makes explicit what isolation indirectly does:

- request on host `brand-a.com` => use config
  ```ts
  [
    { code: 'en', domain: 'brand-a.com', defaultForDomain: true },
    { code: 'nl', domain: 'brand-a.com' }
  ]
  ```
- request on host `brand-b.nl` => use config
  ```ts
  [
    { code: 'nl', domain: 'brand-b.nl', defaultForDomain: true },
    { code: 'en', domain: 'brand-b.nl' }
  ]
  ```

Within each selected sub-config the module's behavior is unchanged; everything new happens in the selection step. So what this feature really is, is configuration determined at runtime: which locales exist, which is the default, and how detection behaves would depend on the incoming request. (Not Nuxt's `runtimeConfig`, which is fixed at process startup, but configuration resolved per request.)

That puts it in a family with several open requests we've been collecting:

- #2602: different domains config at runtime
- #3748: per-domain strategy, locales and default locale from a single build
- #4016: runtime-configurable default locale
- #2289: limit available languages per request

We're not opposed to some form of this, but it conflicts with how much of what the module produces is fixed at build time, and any real version of it has to confront these:

- Routes are the most tractable of these, and `multiDomainLocales` already demonstrates the pattern: the build emits a route variant for every locale that is a domain default somewhere, and at plugin start the route table is rebuilt for the current host from those variants. But that only works because the build enumerated every possible shape in advance; runtime config can select among pre-built route shapes, not produce new ones. Every candidate default has to be declared up front so its variants exist (#4016), and the strategy has no variant mechanism at all (#3748 asks to vary it per host).
- Translations are similar: which locale files end up in the build and how they're loaded is decided at build time. On the server, compiled messages are additionally shared across requests on the assumption that the config is stable, and per-request configuration breaks that sharing at a significant performance cost.
- Prerendering renders each page under exactly one config, and no request host exists at generate time, so any host-dependent behavior gets frozen into the HTML (#2656 is a live example of this class of bug).
- A fully static build serves every host the same files, so a per-host decision like "404 instead of relocate" has nowhere to run at all.

To be fair, the proposed isolation feature avoids most of these problems, precisely because it is still configured at build time: every per-host sub-config is known up front, and they all share the same routes, messages and pages. That constraint is what makes it much easier to implement than general runtime configuration. But it gets there by leaking the identity semantic into places that shouldn't carry it: the locale config now encodes which identity a locale belongs to, and detection, redirects, the switcher and the head alternates each have to interpret that meaning, instead of the config selection happening in one explicit step. And the shortcut only holds while the sub-configs share everything; the moment the identities need different pages or translations, the built resources disagree with the selected config.

So rather than merging isolation as a flag that implicitly implements runtime config selection for one use case, I'd prefer to treat runtime-determined configuration as its own design question: decide which subset is honestly supportable given the constraints above, and state its boundaries up front. I'm keeping this use case on the table as part of that, but I don't want to land it in its current shape.

sivo1981:
Great feature, also waiting for this to be merged.

PR #4118, opened 2026-08-01
fix: reject a messages.json hash that doesn't match the build's own hash (CWE-770)

Vincentdevreede:
## Summary
- The `/_i18n/:hash/:locale/messages.json` endpoint keyed its cache entirely off the `hash` and `locale` router params, but never checked `hash` against the build's own content hash. Any string in that segment produced a fresh 200 response and a 24h cache entry, letting an unauthenticated caller mint unbounded cache entries by iterating the segment (CWE-770).
- Reject a `hash` that doesn't equal `__I18N_LOCALE_HASHES__[locale]` before the request reaches either caching layer, restoring the hash's intended role as a content fingerprint rather than an arbitrary accepted cache key. Combined with the existing locale check into a single 404, since both reject for the same externally observable reason.
- Fixes `specs/lazy_load/app_context_locale.spec.ts`, which had hardcoded a fake hash value and documented the old behavior as expected. It now reads the real hash out of the fixture's own build output, so it can't go stale.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for localized message requests.
  * Missing locales now return a clear bad-request response.
  * Invalid locales or outdated URL hashes now return a 404 instead of attempting to load messages.
  * Prevented crafted requests from accessing another locale’s cached messages.
  * Valid requests continue using build-generated locale hashes for reliable message retrieval.
  * Added safeguards to ensure cached messages remain isolated between similar locales.

### DIFF
--- a/specs/helper.ts
+++ b/specs/helper.ts
@@ -1,6 +1,8 @@
 // @ts-ignore
 import { getBrowser, startServer, TestContext, url, useTestContext } from './utils'
 import { onTestFinished } from 'vitest'
+import { readFileSync, readdirSync } from 'node:fs'
+import { join } from 'node:path'
 
 import type { BrowserContextOptions, Page } from 'playwright-core'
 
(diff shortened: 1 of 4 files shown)


Vincentdevreede:
@BobbieGoede PR got a little bigger. I’ve implemented CodeRabbit’s suggestion, but feel free to ignore that commit if you disagree with it.

BobbieGoede:
Thank you for the PR!

This is a good find and needs to be addressed. But I think this changes behavior for an edge case, where a user with a stale client could hit the endpoint with an old hash and get a 404 🤔 

I'll push some changes tomorrow, would like to include this in the next patch (which is probably next week).

PR #4119, opened 2026-08-02
[WIP] feat: runtime request config (proof of concept)

BobbieGoede:
### 🔗 Linked issue
* #4101

### 📚 Description
This is an exploration and proof of concept on another approach to changing the module behavior by changing the config passed at setup, this would allow projects to introduce other semantics without introducing them to the module code directly.

This is based on what I describe in https://github.com/nuxt-modules/i18n/pull/4101#issuecomment-5151695637, and could be a solution to some of the runtime config requests (a small subset of configs), which is a recurring feature request.

/cc @Vincentdevreede

## Summary by CodeRabbit

- **New Features**
  - Added support for request-specific locale configuration, allowing available languages and default locales to vary by request or domain.
  - Unavailable language routes are now automatically removed, preventing navigation to unsupported localized pages.
  - Improved locale detection using request-specific language settings, browser preferences, and persisted cookies.
  - Added Dutch (`nl-NL`) as an available locale in the playground.
- **Bug Fixes**
  - Unknown or unsupported locales are filtered from runtime configuration to improve locale resolution and fallback behavior.
- **Tests**
  - Added coverage for host-based locale selection, route handling, detection, and localized navigation.


Togetic:
A consumer datapoint for the "which subset is honestly supportable" question, since #4016 is mine and you listed it in this family.

**Our shape.** One image, seven regional deployments, 13 locales in the build, four distinct default locales across those deployments (`en`, `tr`, `ka`, `kk`). `prefix_except_default` + `compactRoutes`. What varies per deployment is exactly the two fields this hook exposes: which locale owns the unprefixed tree, and which locales that region advertises. Same pages, same messages, same strategy everywhere.

The part that matters for scoping: **for us the config is per-process, not per-request.** Each pod serves one region and is pinned at boot by env (`NUXT_PUBLIC_I18N_DEFAULT_LOCALE`; the deploy fails when it is unset). Two configs are never alive in one process.

That slice avoids three of the four constraints you listed on #4101: compiled messages stay shareable across requests, because the config is constant for the lifetime of the process; prerendering has exactly one config to see, and freezing it into the HTML is the intent rather than a bug; and a per-region static build is unaffected. Only routes are left — the one you called most tractable. So if the general per-request version turns out to be too costly to support, the per-process slice is still worth having on its own, and it is all #4016 actually needs.

**Where the hook stops short of #4016.** `defaultLocale` from the hook does reach `prefixable`, so the prefixing decision is already correct. Name-based resolution is not: it still goes through the build-bound `<base>___<buildDefault>`. On a build whose default was `en`, with the hook selecting `de`, `prefixable` says `de` is unprefixed while `localePath('index', 'de')` returns `/de`. (Verified against a pristine 10.6.0 downstream.)

You named that gap here as "every candidate default has to be declared up front so its variants exist (#4016)". A locale-agnostic `<base>___default` name for the unprefixed tree removes the enumeration entirely: one unprefixed shape, and whichever locale the process defaults to resolves through it. That is all #4017 does, and it composes with this hook rather than competing — the hook decides *which* locale, the neutral name gives it *something to resolve to*. Equally happy if the flag isn't the shape you want; a hook that can influence the route-name getter would let the whole thing live in userland.

**On the selection half**, `pruneUnservedLocaleRoutes` is the supported version of something we do by hand today: our per-region locale subset comes from an API at runtime and only narrows the language picker, so `/hu/...` stays reachable on a pod that never advertises `hu`. We would use it.

One cross-link for the perf constraint you raised: @ikz2018 posted a CPU profile on #4130 (Aug 20) where the request hook's `useRuntimeI18n(undefined, event)` → `useRuntimeConfig(event)` accounts for ~16% of non-idle samples on 10.6.0 as it ships (`getEnv` alone 8.9%). That was measured on static-asset requests served from their Node pods, so the share is specific to that traffic mix — but it is independent evidence for the cost you flagged, and per-request config would put that path on the critical path by design rather than by accident.

PR #4124, opened 2026-08-04
fix: fetch a region-tagged locale's base tag when it's configured

Vincentdevreede:
## Summary

vue-i18n automatically falls back a region-tagged locale to its base language tag (`en-US` tries `en`) before it even looks at `fallbackLocale`. This is documented, default behavior in vue-i18n itself. Nuxt i18n's lazy loading never fetched that base tag's messages unless it was listed explicitly in `fallbackLocale`, so this fallback silently did nothing for anyone with a region-tagged locale like `en-US` alongside `en`.

`getFallbackLocaleCodes` now adds a locale's base tag to its fallbacks when that base tag is itself a configured locale, walking multiple dash-separated levels where needed (`de-DE-bavarian` → `de-DE` → `de`). If the base tag isn't configured at all, nothing changes, there's no file to load for it and no reason to widen the lazy-loading footprint for a tag nobody defined.

Also fixes a duplicate entry this introduced: an explicit `fallbackLocale` entry can name the same locale the base-tag walk already adds on its own (writing `{'en-US': ['en']`} achieves exactly what `en-US`'s implicit `en` already does), so the result is deduped.

## ~~Summary~~
~~Vue I18n automatically falls back a region-tagged locale to its base language tag (fr-CA tries fr), but that only works for messages already loaded in memory. Since Nuxt i18n lazy-loads locale files, it only fetches what's explicitly listed in fallbackLocale, so that implicit chain silently does nothing unless the base tag is configured too.~~

~~I ran into this and initially thought it was a bug, but on reflection I think it's probably intentional: automatically fetching a base tag's file on top of whatever's explicitly configured would mean loading more locale files than the user asked for, which cuts against the whole point of lazy loading. So instead of changing behavior, this PR just documents it, with a callout showing how to get the same effect by listing the base tag yourself in fallbackLocale.~~

~~@BobbieGoede can you confirm that's the intended reasoning? If not, and this is actually considered a bug, let me know and I'll put together a code fix instead of the docs note.~~

This is an older summary from before I force-pushed the changes. I’m just keeping it here for reference.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved locale fallback handling for regional and multi-level locale tags.
  - Regional locales now correctly include configured base locales in fallback results.
  - Prevented duplicate fallback locales across disabled, array-based, and mapped fallback configurations.

- **Tests**
  - Added coverage for regional, plain, multi-level, missing-base, and duplicate locale scenarios.

### DIFF
--- a/docs/content/docs/02.guide/11.locale-fallback.md
+++ b/docs/content/docs/02.guide/11.locale-fallback.md
@@ -25,3 +25,7 @@ export default {
 ```
 
 More information in [Vue I18n documentation](https://vue-i18n.intlify.dev/guide/essentials/fallback.html)
+
+<callout icon="i-heroicons-exclamation-triangle" color="warning">
+**Vue I18n** automatically falls back from a region-specific locale to its base language (`fr-CA` tries `fr` before giving up), but this only works for messages that are already loaded in memory. Because **Nuxt i18n lazy-loads locale files**, it only loads the current locale’s file and any fallback locales explicitly configured in `fallbackLocale`. To make `fr-CA` fall back to `fr`, you therefore need to configure `fr` as a fallback yourself.
+</callout>
--- a/docs/content/docs/02.guide/11.locale-fallback.md
+++ b/docs/content/docs/02.guide/11.locale-fallback.md
@@ -25,7 +25,3 @@ export default {
 ```
 
 More information in [Vue I18n documentation](https://vue-i18n.intlify.dev/guide/essentials/fallback.html)
-
-<callout icon="i-heroicons-exclamation-triangle" color="warning">
-**Vue I18n** automatically falls back from a region-specific locale to its base language (`fr-CA` tries `fr` before giving up), but this only works for messages that are already loaded in memory. Because **Nuxt i18n lazy-loads locale files**, it only loads the current locale’s file and any fallback locales explicitly configured in `fallbackLocale`. To make `fr-CA` fall back to `fr`, you therefore need to configure `fr` as a fallback yourself.
-</callout>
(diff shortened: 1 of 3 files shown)


BobbieGoede:
This looks like a bug, at the same time I can imagine users are already defining their fallback locale files using the file arrays (e.g. `files: ['en.json', 'en-US.json']`). Not sure if we're already deduplicating file chains, otherwise we may need to take that into account when fixing this.

Explicitly configuring fallbacks does work (example below), so I guess we only need to parse the codes and add these as fallbackLocales if such locales exist.

```ts
// in nuxt.config.ts i18n config
  // ...
  i18n: {
    locales: [
      { code: 'en', language: 'en', file: 'en.json', name: 'English' },
      { code: 'en-US', language: 'en-US', file: 'en-US.json', name: 'English (US)' },
    ]
  }
})
```

```ts
// i18n.config.ts
export default defineI18nConfig(() => {
  return {
    fallbackLocale: {
      'en-US': ['en'], // en.json is loaded when locale is en-US
    },
  }
})
```

>I ran into this and initially thought it was a bug, but on reflection I think it's probably intentional: automatically fetching a base tag's file on top of whatever's explicitly configured would mean loading more locale files than the user asked for, which cuts against the whole point of lazy loading.

This is true for some projects (with dynamic locale files or files that need to be run from client-side/nuxt), but most projects (static files) will fetch through the messages endpoint where the server already has the locale messages loaded/merged.

Vincentdevreede:
@BobbieGoede Updated this to an actual fix rather than just a docs note. getFallbackLocaleCodes now adds a region-tagged locale's base tag as an implicit fallback when that base tag is itself configured (e.g. en alongside en-US), matching vue-i18n's own behavior.

Vincentdevreede:
@BobbieGoede Opened #4128 on top of this branch. 
Testing this fix against vue-i18n's real chain building turned up a few more cases it doesn't cover yet: a fallbackLocale map entry can redirect the walk before it ever reaches a base tag, redirecting abandons any other entries still left in that block, and there's a ! suffix that stops a redirect target from walking its own base tag.

PR #4127, opened 2026-08-10
fix(vite): load YAML messages in server environments

onmax:
### 🔗 Linked issue

Related to #4066. This is extracted from #4088 so the Vite resource fix can be reviewed independently.

### 📚 Description

With message bundling optimization enabled, static YAML and other raw locale resources should be handled by the server JSON-parse plugin. Vite's server environment currently reaches the resource plugin first, which resolves the virtual ID to the YAML file and leaves development SSR loading it through the wrong path.

The resource plugin now leaves tracked raw virtual IDs unresolved only for Vite server consumers, while preserving existing client and non-raw resolution. The focused unit test fails with the previous resolver and covers the server yield and client fallback.

Verified with:

- `pnpm test:unit` — 465 passed, 1 skipped
- `pnpm lint` — passes with six existing JSDoc warnings
- `pnpm exec tsc --noEmit`

`pnpm test:types` passes its root TypeScript phase, then reaches existing `typed_routes` fixture errors on current `main`.

### DIFF
--- a/src/transform/resource.ts
+++ b/src/transform/resource.ts
@@ -46,8 +46,14 @@ export const ResourcePlugin = (options: BundlerPluginOptions, ctx: ResolvedI18nC
           return id
         }
 
-        if (i18nFileHashSet.has(id)) {
-          return i18nFileHashSet.get(id)
+        const path = i18nFileHashSet.get(id)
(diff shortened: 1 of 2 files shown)


BobbieGoede (approved):
Thanks!

PR #4130, opened 2026-08-11
perf: cache createLocaleConfigs per fallbackLocale value

Vincentdevreede:
## Summary

`createLocaleConfigs` recomputes every configured locale's fallback chain from scratch on every request that doesn't already have a cached i18n context, `initializeI18nContext` runs on Nitro's `request` hook unconditionally, so this happens on every single request. Its only real input, `fallbackLocale`, is normally the same value every time.

This adds a small in-memory cache keyed by a serialization of `fallbackLocale` itself, rather than memoizing once and reusing forever. `fallbackLocale` comes from `defineI18nConfig`, which can be async, so in theory it could differ per request, keying by value keeps that case correct (a different value just gets its own cache entry) while still skipping the work entirely whenever it's what it almost always is: unchanged.

## Benchmark

Measured directly against the real function (forcing a genuine cache miss on every "uncached" call versus repeated calls with the same value):

| locales | uncached (per call) | cached (per call) | speedup |
|---|---|---|---|
| 5 | 5.3 µs | 0.9 µs | 5.6x |
| 20 | 11.8 µs | 0.8 µs | 14.9x |
| 100 | 39.8 µs | 0.7 µs | 54.9x |

The cached path stays flat regardless of locale count, it's just a map lookup. The uncached path scales with locale count, since it walks every configured locale's fallback chain. Small in absolute terms per request, but free to keep, and it scales better the more locales are configured.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved locale configuration handling by reusing previously generated configurations, reducing repeated processing.

### DIFF
--- a/src/runtime/shared/locales.ts
+++ b/src/runtime/shared/locales.ts
@@ -5,7 +5,19 @@ import type { FallbackLocale } from 'vue-i18n'
 import type { NormalizedLocaleObject } from '#internal-i18n-types'
 
 type LocaleConfig = { cacheable: boolean, fallbacks: string[] }
+
+// `fallbackLocale` is normally the same value on every request, computing every locale's fallback
+// chain from scratch each time is wasted work. It comes from `defineI18nConfig`, which can be
+// async, so it could in theory differ per request, keying the cache by the value itself (rather
+// than memoizing once and reusing forever) keeps that case correct while still skipping the work
+// whenever it's what it almost always is: unchanged
+const localeConfigsCache = new Map<string, Record<string, LocaleConfig>>()
+
 export function createLocaleConfigs(fallbackLocale: FallbackLocale): Record<string, LocaleConfig> {
+  const cacheKey = JSON.stringify(fallbackLocale)
+  const cached = localeConfigsCache.get(cacheKey)
+  if (cached) { return cached }
+
   const localeConfigs: Record<string, LocaleConfig> = {}
 
   for (const locale of localeCodes) {
@@ -14,6 +26,7 @@ export function createLocaleConfigs(fallbackLocale: FallbackLocale): Record<stri
     localeConfigs[locale] = { fallbacks, cacheable }
   }
 
+  localeConfigsCache.set(cacheKey, localeConfigs)
   return localeConfigs
 }
 


ikz2018:
We ran into the same "runs on every request" problem you describe here, and I wanted to share profiling data in case
  it's useful — in our case the dominant cost inside initializeI18nContext turned out to be a different line.

  Context: we upgraded a Nuxt 3.15 monorepo to Nuxt 4.5.2 / @nuxtjs/i18n 10.6.0 and saw response times at our ingress go
  from ~50 ms to 250–500 ms, with pods saturating on CPU. We serve static assets from the Node pods themselves
  (node-server preset, no CDN in front), so every /_nuxt/*.js chunk request goes through the Nitro request pipeline.

  Profiling the production build under load (node --cpu-prof, requests for a single 442-byte JS chunk, 20 concurrent
  keep-alive connections) showed this stack:

  getEnv
    ← applyEnv
    ← useRuntimeConfig
    ← useRuntimeI18n
    ← initializeI18nContext
    ← nitro hook "request"

  Sample distribution in that profile:

  ┌──────────────────┬─────────────────────────────────────┐
  │     function     │            % of samples             │
  ├──────────────────┼─────────────────────────────────────┤
  │ getEnv           │ 8.9 %                               │
  ├──────────────────┼─────────────────────────────────────┤
  │ kebabCase        │ 2.1 %                               │
  ├──────────────────┼─────────────────────────────────────┤
  │ snakeCase        │ 1.7 %                               │
  ├──────────────────┼─────────────────────────────────────┤
  │ splitByCase      │ 1.1 %                               │
  ├──────────────────┼─────────────────────────────────────┤
  │ deepCopy         │ 1.1 %                               │
  ├──────────────────┼─────────────────────────────────────┤
  │ applyEnv + klona │ 1.2 %                               │
  ├──────────────────┼─────────────────────────────────────┤
  │ total            │ 16.5 % (≈ half of all non-idle CPU) │
  └──────────────────┴─────────────────────────────────────┘

  That's useRuntimeI18n(undefined, event) → useRuntimeConfig(event), which clones the whole runtime config and walks
  every key through splitByCase/kebabCase/snakeCase looking for process.env overrides — once per request, including
  requests that will never touch i18n.

  Measured impact, same build, only difference being a guard on the request hook:

  ┌───────────────────────────────────────────────────────────────┬───────┐
  │                                                               │  rps  │
  ├───────────────────────────────────────────────────────────────┼───────┤
  │ 10.6.0 as-is                                                  │ ~2065 │
  ├───────────────────────────────────────────────────────────────┼───────┤
  │ 10.6.0 with the request hook skipping buildAssetsDir          │ ~8290 │
  ├───────────────────────────────────────────────────────────────┼───────┤
  │ (our previous stack, Nuxt 3.15.4 + i18n 9.5.6, for reference) │ ~9060 │
  └───────────────────────────────────────────────────────────────┴───────┘

  The guard we tested:

  const { buildAssetsDir } = useRuntimeConfig().app // resolved once at plugin setup

  nitro.hooks.hook('request', async (event) => {
    if (event.path.startsWith(buildAssetsDir)) return
    await initializeI18nContext(event)
  })

  Static asset 404s still return 404 (Nitro throws those before the renderer, so render:before never runs and
  useI18nContext is never reached), and SSR requests are untouched.

  Worth noting our results don't contradict yours — they're complementary. We run 2–3 locales, and createLocaleConfigs
  didn't appear in our profile at all, which lines up with your benchmark showing the win scaling with locale count.

  Happy to open a separate issue with the full reproduction and profiles, and a PR for the guard, if that's useful. Not
  suggesting any change to this PR.

@BobbieGoede  Could you take a look when you get a chance? I don't have much experience contributing, so I posted here since this looked like the same problem — happy to move it to a separate issue if that's a better place.

PR #4137, opened 2026-08-14
fix: add normalization for repeatable parameters

### LINKED ISSUE #4135: Bug: `trailingSlash` produces double slash for catch-all routes in `useLocaleHead` and `useSwitchLocalePath`
### Environment

|                      |                              |
| -------------------- | ---------------------------- |
| **Operating system** | `macOS 25.6.0`               |
| **CPU**              | `Apple M4 Pro (12 cores)`    |
| **Node.js version**  | `v24.19.0`                   |
| **nuxt/cli version** | `3.37.0`                     |
| **Package manager**  | `pnpm@11.21.0`               |
| **Nuxt version**     | `4.5.2`                      |
| **Nitro version**    | `2.13.4`                     |
| **Builder**          | `vite@8.2.1`                 |
| **Config**           | `i18n`, `modules`, `nitro`   |
| **Modules**          | `@nuxtjs/i18n@10.6.0`        |

### Reproduction

I created reproduction [here](https://github.com/rastuhacode/i18n-trailing-slash-reproduction)

To be short:

1) Create repo with `nuxt` and `nuxtjs/i18n`.
2) Modify nuxt config to be like:
```ts
  modules: ["@nuxtjs/i18n"],

  i18n: {
    baseUrl: "https://example.com",
    locales: [
      { code: "en", language: "en-US" },
      { code: "fr", language: "fr-FR" },
    ],
    defaultLocale: "en",
    strategy: "prefix_except_default",
    trailingSlash: true,
  },

  nitro: {
    prerender: {
      routes: ["/posts/"],
    },
  },
});
```
importnant to have several locales and `trailingSlash: true`.
3) Create catch-all page like `app/pages/[...slug].vue` and add basic locale switcher and logs.

### Describe the bug

When `trailingSlash: true` is used with a catch-all Nuxt route such as
`app/pages/[...slug].vue` from `nuxt/content`, URLs generated from the current route can contain
two trailing slashes.

The issue appears to affect URL generation based on `switchLocalePath()`,
which is also used by `SwitchLocalePathLink` and `useSwitchLocalePath()`.

This issue have two real problems:
1) Visiting `/test/` produces `/test//` in metadata generated by
`useLocaleHead()`.
2) Switching between locales with prefixes (`/` -> `/fr/`) spawns and stacks additional slash in each switch (`/fr/test/` -> n switches -> `/fr/test///...///`).

Result would look like this in SEO:
```html
<link rel="canonical" href="https://example.com/posts//">
<link rel="alternate" href="https://example.com/fr/posts//">
<meta property="og:url" content="https://example.com/posts//">
```
Expected the same but with one `/`

### Additional context

_No response_

### Logs

```shell

```

rastuhacode:
### 🔗 Linked issue

Resolves #4135

### 📚 Description

With `trailingSlash: true`, route localiztion creates:
```
/:slug(.*)*/
/fr/:slug(.*)*/
```

Vue Router parses it like:
```ts
// Url: .../test/
{ slug: ['test', ''] }
```
Then `switchLocalePath()` copies those params into the localized named route. Vue Router joins the array as `test/`, then the localized route pattern adds its configured slash, so basically happens:
```
/fr/ + test/ + / = /fr/test//
```

Navigating there produces another empty parameter element, so the process repeats indefinitely.
The malformed SEO URLs are a downstream symptom: locale head generation calls switchLocalePath() for canonical and alternate URLs.

### Solution

- Added normalization process for repeatable params (`normalizeRepeatableParams`), which removes only trailing empty elements from repeatable route params before named-route resolution.
- Added regression tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved localized route handling for repeatable and catch-all parameters.
  * Preserved interior empty path segments while removing unnecessary trailing segments.
  * Ensured locale switching and route resolution produce consistent URLs.
  * Corrected canonical localized URLs to use exactly one trailing slash where applicable.

### DIFF
--- a/src/runtime/routing/routing.ts
+++ b/src/runtime/routing/routing.ts
@@ -2,7 +2,7 @@ import { hasProtocol, parsePath, parseQuery } from 'ufo'
 import { assign, isString } from '@intlify/shared'
 
 import type { Locale } from 'vue-i18n'
-import type { RouteLocationNamedRaw, RouteLocationPathRaw, RouteLocationRaw } from 'vue-router'
+import type { RouteLocationNamedRaw, RouteLocationPathRaw, RouteLocationRaw, RouteParamsRawGeneric } from 'vue-router'
 import type { CompatRoute } from '../types'
 import type { RoutingContext } from './context'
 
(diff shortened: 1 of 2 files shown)


rastuhacode:
UPD: The PR which resolves the same issue #4136 

I kept this open for reference and just in case. Maintainers can close it if they find it unuseful.

PR #4139, opened 2026-08-15
fix(types): type the i18n config key with the augmentable ModuleOptions interface

### LINKED ISSUE #4138: NuxtConfig['i18n'] is typed via a type alias, so it cannot be augmented by other modules
`NuxtConfig['i18n']` is declared through a type alias, so other modules cannot augment it.

`dist/module.d.mts`:

```ts
type UserNuxtI18nOptions = Omit<NuxtI18nOptions, 'locales'> & { locales?: ... }

interface ModuleOptions extends UserNuxtI18nOptions {}   // augmentable

declare module '@nuxt/schema' {
  interface NuxtConfig {
    ['i18n']?: Partial<UserNuxtI18nOptions>   // but this uses the alias
  }
}
```

`ModuleOptions` is an interface and can be augmented, but the config key is typed via the alias, so an augmentation never reaches it:

```
error TS2353: Object literal may only specify known properties,
and 'glossary' does not exist in type 'Partial<UserNuxtI18nOptions>'.
```

(Nuxt 4.4.2, `@nuxtjs/i18n` 10.2.3, checked against Nuxt's generated `.nuxt/tsconfig.node.json`. Unknown keys survive normalisation at runtime — only the types reject them.)

**Suggested fix:** declare the key in terms of the interface.

```ts
['i18n']?: Partial<ModuleOptions>
```

That keeps the public type identical while letting companion modules add their own keys under `i18n` via standard declaration merging.

lazerg:
### 🔗 Linked issue

* Resolves #4138

### 📚 Description

`ModuleOptions` is an interface, so companion modules can extend it through declaration merging. The `i18n` key on `NuxtConfig` and `NuxtOptions` is typed with the `UserNuxtI18nOptions` alias instead, which an augmentation never reaches, so adding a custom key under `i18n` fails with TS2353 even though the value survives at runtime.

Both keys now point at `ModuleOptions`. It extends `UserNuxtI18nOptions` and adds no members, so the exposed type stays the same and only augmentability changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved TypeScript support for internationalization configuration in Nuxt projects.
  - Configuration options now provide more accurate and consistent type checking across Nuxt configuration contexts.
  - Helps surface invalid settings earlier and improves editor autocomplete when configuring internationalization.

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -274,10 +274,10 @@ declare module '#app' {
 
 declare module '@nuxt/schema' {
   interface NuxtConfig {
-    ['i18n']?: Partial<UserNuxtI18nOptions>
+    ['i18n']?: Partial<ModuleOptions>
   }
   interface NuxtOptions {
-    ['i18n']: UserNuxtI18nOptions
+    ['i18n']: ModuleOptions
   }
   interface NuxtHooks extends ModuleHooks {}
   interface PublicRuntimeConfig extends ModulePublicRuntimeConfig {}


lazerg:
Closing as no one reviewed and this has been open more than a month with no activity. Happy to reopen if there's interest — thanks for your time.

PR #4145, opened 2026-08-23
perf: skip i18n context setup for build asset requests

### LINKED ISSUE #4144: initializeI18nContext runs on every request and response times at our ingress go from ~50 ms to 250–500 ms
### Environment

│ │  Node.js version       v22.17.0                                │
│ │  nuxt/cli version      3.37.0                                  │
│ │  Package manager       pnpm 9.15.9                             │
│ │  Nuxt version          4.5.2                                   │
│ │  Nitro version         2.13.4                                  │
│ │  Builder               vite 8.2.1                              │
│ │  Config                -                                       │
│ │  Modules               - 

### Reproduction

every repo

### Describe the bug

Context: we upgraded a Nuxt 3.15 monorepo to Nuxt 4.5.2 / @nuxtjs/i18n 10.6.0 and saw response times at our ingress go
from ~50 ms to 250–500 ms, with pods saturating on CPU. We serve static assets from the Node pods themselves
(node-server preset, no CDN in front), so every /_nuxt/*.js chunk request goes through the Nitro request pipeline.

Profiling the production build under load (node --cpu-prof, requests for a single 442-byte JS chunk, 20 concurrent
keep-alive connections) showed this stack:

getEnv
← applyEnv
← useRuntimeConfig
← useRuntimeI18n
← initializeI18nContext
← nitro hook "request"

Sample distribution in that profile:

┌──────────────────┬─────────────────────────────────────┐
│ function │ % of samples │
├──────────────────┼─────────────────────────────────────┤
│ getEnv │ 8.9 % │
├──────────────────┼─────────────────────────────────────┤
│ kebabCase │ 2.1 % │
├──────────────────┼─────────────────────────────────────┤
│ snakeCase │ 1.7 % │
├──────────────────┼─────────────────────────────────────┤
│ splitByCase │ 1.1 % │
├──────────────────┼─────────────────────────────────────┤
│ deepCopy │ 1.1 % │
├──────────────────┼─────────────────────────────────────┤
│ applyEnv + klona │ 1.2 % │
├──────────────────┼─────────────────────────────────────┤
│ total │ 16.5 % (≈ half of all non-idle CPU) │
└──────────────────┴─────────────────────────────────────┘

That's useRuntimeI18n(undefined, event) → useRuntimeConfig(event), which clones the whole runtime config and walks
every key through splitByCase/kebabCase/snakeCase looking for process.env overrides — once per request, including
requests that will never touch i18n.

Measured impact, same build, only difference being a guard on the request hook:

┌───────────────────────────────────────────────────────────────┬───────┐
│ │ rps │
├───────────────────────────────────────────────────────────────┼───────┤
│ 10.6.0 as-is │ ~2065 │
├───────────────────────────────────────────────────────────────┼───────┤
│ 10.6.0 with the request hook skipping buildAssetsDir │ ~8290 │
├───────────────────────────────────────────────────────────────┼───────┤
│ (our previous stack, Nuxt 3.15.4 + i18n 9.5.6, for reference) │ ~9060 │
└───────────────────────────────────────────────────────────────┴───────┘

The guard we tested:

const { buildAssetsDir } = useRuntimeConfig().app // resolved once at plugin setup

nitro.hooks.hook('request', async (event) => {
if (event.path.startsWith(buildAssetsDir)) return
await initializeI18nContext(event)
})

Static asset 404s still return 404 (Nitro throws those before the renderer, so render:before never runs and
useI18nContext is never reached), and SSR requests are untouched.

Worth noting our results don't contradict yours — they're complementary. We run 2–3 locales, and createLocaleConfigs
didn't appear in our profile at all, which lines up with your benchmark showing the win scaling with locale count.

### Additional context

_No response_

### Logs

```shell

```

lazerg:
* Resolves #4144

The nitro `request` hook sets up the i18n context for every request, including the `/_nuxt/*` chunks a node-server deployment serves itself. That setup calls `useRuntimeConfig(event)`, which clones the whole runtime config and walks every key looking for env overrides, so a 442-byte JS chunk pays for a full config pass it never reads. The reporter profiled it at ~16.5% of non-idle CPU, with throughput dropping from ~9060 rps on Nuxt 3.15 + i18n 9.5.6 to ~2065 rps on 10.6.0.

The hook now returns early for paths under `app.buildAssetsDir`, resolved once during plugin setup. Nothing downstream reads the context on that path: nitro serves or 404s build assets before `render:before` runs, so `useI18nContext` is never reached for them. The reporter measured ~8290 rps with the same guard, and SSR requests are unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented i18n initialization for build asset requests, improving asset delivery and avoiding unnecessary request processing.
  * Preserved i18n initialization for regular page requests, including applications served from the root path.

* **Tests**
  * Added coverage for build assets with and without a configured base URL, as well as standard page requests.

### DIFF
--- a/src/runtime/server/plugin.ts
+++ b/src/runtime/server/plugin.ts
@@ -35,6 +35,9 @@ function createRedirectResponse(event: H3Event, dest: string, code: number) {
 
 export default defineNitroPlugin(async (nitro) => {
   const runtimeI18n = useRuntimeI18n()
+  const { baseURL, buildAssetsDir } = useRuntimeConfig().app
+  // unlike in `render:before`, `event.path` still carries `app.baseURL` in the `request` hook
+  const buildAssetsPath = buildAssetsDir !== '/' ? joinURL(baseURL, buildAssetsDir) : undefined
   const rootRedirect = resolveRootRedirect(runtimeI18n.rootRedirect)
   const _defaultLocale: string = runtimeI18n.defaultLocale || ''
 
@@ -99,6 +102,9 @@ export default defineNitroPlugin(async (nitro) => {
   }
 
   nitro.hooks.hook('request', async (event: H3Event) => {
+    // build assets are served without ever reading the context, and setting it up reprocesses
+    // the runtime config on every one of those requests (#4144)
+    if (buildAssetsPath && event.path.startsWith(buildAssetsPath)) { return }
     await initializeI18nContext(event)
   })
 
--- a/test/mocks/i18n.locale-detector.ts
+++ b/test/mocks/i18n.locale-detector.ts
@@ -1,0 +1,1 @@
+export const localeDetector = undefined
--- a/test/mocks/nuxt.internal-config.ts
+++ b/test/mocks/nuxt.internal-config.ts
@@ -1,0 +1,1 @@
+export const appId = 'test'
--- a/test/server-plugin.test.ts
+++ b/test/server-plugin.test.ts
@@ -1,0 +1,61 @@
+import { expect, test, vi } from 'vitest'
+import type { H3Event } from 'h3'
+
+const { initializeI18nContext, app } = vi.hoisted(() => ({
+  initializeI18nContext: vi.fn(),
+  app: { baseURL: '/', buildAssetsDir: '/_nuxt/' },
+}))
+vi.mock('../src/runtime/server/context', () => ({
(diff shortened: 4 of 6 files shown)


lazerg:
Closing this since it has been inactive for a while.

